PR #109, opened 2020-04-15
add coverage

frostming:
## Pull Request Check List


frostming:
Don't know why tests failed on macOS

danieleades:
can i help with this?

frostming:
@danieleades Thanks, there are some mystery errors with macOS. Some tomlkit internal data types have some problem sharing between processes of `pytest-xdist` and I cannot figure out why. It might be an issue of `tomlkit` or `pytest-xdist`.

Even so, I suspect test coverage is worth doing for such a CLI tool that heavily depends on the environment setup, among which there are different shells, different OS, different terminal apps and different Python installations to test. Unless we cover as many test paths as possible but that requires hard work of the design of testing infrastructure.

So far, Pipenv has 270 test cases and Poetry has 649 test cases, neither of them has test coverage.  For now, adding more test cases (226 by far) could be of more help and that calls for some contribution and I will appreciate that.

uranusjr:
LIkely it’s a `tomlkit` issue. Last time I checked, it inherits several built-in types without overriding `__reduce__` (or `__reduce_ext__`), which is used to serialise objects for IPC, so instances loose information when passed around.

PR #188, opened 2020-12-05
Add zsh completion script

Aloxaf:
## Pull Request Check List

Add a zsh completion script.

<s>I am not familiar with the python package and not sure where should I place this.
And should I turn `pdm completion zsh`'s behavior into printing this file?</s>
Done


pdm/cli/commands/completion.py
```diff
@@ -22,5 +23,11 @@ def handle(self, project: Project, options: argparse.Namespace) -> None:
         import shellingham
         from pycomplete import Completer
 
-        completer = Completer(project.core.parser)
-        stream.echo(completer.render(options.shell or shellingham.detect_shell()[0]))
+        if options.shell == "zsh":
```
frostming, line 26: `shell = options.shell or shellingham.detect_shell()[0]`

frostming (commented):
is it possible to make the script a template that is responsible to the commands and options list?

Aloxaf:
> is it possible to make the script a template that is responsible to the commands and options list?

Yes, it's possible. But in that way, I will reinvent pycomplete.

I think this should be done in pycomplete: pycomplete should be able to accept a template and generate something like this for options with args.

```zsh
case $state in; 
  update)
    if (( $+functions[_pdm_update] )); then
	  _pdm_update
      return
	fi
    opts=(
	  '--section+[Specify the sectionn the package belongs to]:_pdm_update_SECTION'
      '*:packages:_pdm_update_packages'
    )
    ;;
esac
```

Then we just need to define function `_pdm_update_packages` and `_pdm_update_SECTION` in template. And we can define `_pdm_update` to take over the completion of this subcommand if needed (optional).

frostming:
I am not a shell expert but that seems great. I am always thinking of how `pycomplete` can be extended. I am looking forward to your PR. BTW the completion script looks terrific.

Aloxaf:
> What about Fish?

Well, I am not familiar with Fish. The Fish part may need a Fish expert to finish it. XD

frostming:
@Aloxaf Sorry to disturb you, but I am planning to switch the completion from automated to partially-automated and to be built into the release process. I would like to know if you can relook this PR and refine the zsh completion script to be up-to-date? Thanks in advance.

Aloxaf:
Updated. In fact, I have been updating the script according to the release note all the time.

frostming:
Thanks for the contribution, I would start from `completion` branch to add a release script for completion generation, which might require your review if you have time.

PR #214, opened 2020-12-30
Don't add default build deps onto specified build deps when PEP517 bu…

### LINKED ISSUE #213: Error building a wheel for a dependency using setuptools and without a PEP517 build-backend specified

## Steps to reproduce
pdm use python3.9
pdm add "multidict==4.7.6"

This will reproduce for any dependency meeting the following criteria:
* Has no prebuilt wheels for the in-use python
* Has a pyproject.toml with a build-system table and a suitable build-requires per PEP518
* Does NOT have a build-backend specified per PEP517

## Actual behavior
pip will fail when setting up the isolated build environment. 

Build log:
```Preparing isolated env for PEP 517 build...
ERROR: Double requirement given: setuptools>=40.8.0 (from -r /tmp/pdm-build-reqs-9v3tpxwp.txt (line 4)) (already in setuptools>=40 (from -r /tmp/pdm-build-reqs-9v3tpxwp.txt (line 2)), name='setuptools')
Preparing isolated env for PEP 517 build...
ERROR: Double requirement given: setuptools>=40.8.0 (from -r /tmp/pdm-build-reqs-t530gg95.txt (line 4)) (already in setuptools>=40 (from -r /tmp/pdm-build-reqs-t530gg95.txt (line 2)), name='setuptools')
```

contents of the errant build-reqs.txt:
```
pip>=18
setuptools>=40
wheel
setuptools >= 40.8.0
wheel
``` 
This is the [build system requirements of multidict](https://github.com/aio-libs/multidict/blob/93bf4b6192ad7f1826a2fd66f34e9192200f5e1e/pyproject.toml#L2) prepended to the requirements of pdm's [DEFAULT_BACKEND](https://github.com/frostming/pdm/blob/005e6ed1549330479c12d9152fab14c5f04e22cd/pdm/models/builders.py#L131).

## Expected behavior
The wheel should build succesfully, as it does with both poetry and pip. 

According to PEP518:
> If the file exists but is lacking the [build-system] table then the default values as specified above should be used. If the table is specified but is missing required fields then the tool should consider it an error.

So, when build-system.requires *is* present, pdm should just accept it verbatim. Appending the default backend requirements to the specified backend requirements is not supported by either PEP517 or PEP518.

## Environment Information
Reproduces on both dev (1.0.0b0) stable (0.12.13)

orions-stardom:
…ild-backend is missing.

fixes #213 
## Pull Request Check List

- [*] A news fragment is added in `news/` describing what is new.

### DIFF
--- a/news/213.bugfix
+++ b/news/213.bugfix
@@ -1,0 +1,1 @@
+Correctly build wheels for dependencies with build-requirements but without a specified build-backend
--- a/pdm/models/builders.py
+++ b/pdm/models/builders.py
@@ -151,10 +151,6 @@ class EnvBuilder:
 
         if "build-backend" not in self._build_system:
             self._build_system["build-backend"] = self.DEFAULT_BACKEND["build-backend"]
-            self._build_system["requires"] = (
-                self._build_system.get("requires", [])
-                + self.DEFAULT_BACKEND["requires"]
-            )
 
         if "requires" not in self._build_system:
             raise BuildError("Missing 'build-system.requires' in pyproject.toml")


frostming:
Thanks 🎉

PR #220, opened 2021-01-15
Add ability to read venv location from file

uranusjr:
## Pull Request Check List

### DIFF
--- a/news/220.feature
+++ b/news/220.feature
@@ -1,0 +1,3 @@
+Allow detecting a virtual environment from a ``.venv``, ``venv``, or ``env``
+file. The file should contain a single line specifying the path to the virtual
+environment. If the path is relative, it is resolved against the project root.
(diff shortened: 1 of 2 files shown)


frostming:
Since the Python interpreter path is already written in the `.pdm.toml` under the project, I would prefer Single Source of Truth.

I am working on #219 so that one can achieve this by following:

```bash
$ pdm use -f <python_path>
```

Tell me if it is OK or you still insist on the current implementation.

uranusjr:
I would prefer if it’s possible to also do `pdm use <path-to-virtual-environment>`. Easier to tab-complete, especially when I switch between Windows and POSIX 🙂

frostming:
I got that and created #221 to track it.

frostming:
#221 and #219 are already implemented in the main branch, I would like to close this PR

PR #298, opened 2021-03-12
update type annotations

danieleades:
- replaced a few type comments with type annotations

wanted to check this was of interest before i got too deep into this

### DIFF
--- a/pdm/models/specifiers.py
+++ b/pdm/models/specifiers.py
@@ -82,9 +82,9 @@ class PySpecSet(SpecifierSet):
         if version_str == "*":
             version_str = ""
         super().__init__(version_str)
-        self._lower_bound = self.MIN_VERSION  # type: Tuple[int, int, int]
-        self._upper_bound = self.MAX_VERSION  # type: Tuple[int, int, int]
-        self._excludes = []  # type: List[Tuple[Union[int, str], ...]]
+        self._lower_bound = self.MIN_VERSION
+        self._upper_bound = self.MAX_VERSION
+        self._excludes: List[Tuple[Union[int, str], ...]] = []
         if version_str and analyze:
             self._analyze_specifiers()
 
@@ -92,7 +92,7 @@ class PySpecSet(SpecifierSet):
         # XXX: Prerelease or postrelease specifiers will fail here, but I guess we can
         # just ignore them for now.
         lower_bound, upper_bound = self.MIN_VERSION, self.MAX_VERSION
-        excludes = set()  # type: Set[Tuple[Union[int, str], ...]]
+        excludes: Set[Tuple[Union[int, str], ...]] = set()
         for spec in self:
             op, version = spec.operator, spec.version
             version = parse_version_tuple(version)
--- a/news/298.misc.md
+++ b/news/298.misc.md
@@ -1,0 +1,1 @@
+replace 'typing comments' with type annotations throughout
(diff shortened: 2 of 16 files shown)


frostming:
Glad to see this piece of work is taken and started. I will be even better if a step can be added to `Lint` CI that invokes `mypy` in order to see the effect of the improvement.

danieleades:
> Glad to see this piece of work is taken and started. It will be even better if a step can be added to `Lint` CI that invokes `mypy` in order to see the effect of the improvement.

Cool, I'll keep going.
I'd like to add mypy linting too, that's my eventual goal. but there's a way to go before that!

danieleades:
that's the rest of the type comments replaced with annotations.

mypy is currently returning >500 errors. Will probably start chipping away at those in separate pull requests, and keep this one focused on replacing the type comments only.

pdm/models/specifiers.py
```diff
@@ -82,17 +82,17 @@ def __init__(self, version_str: str = "", analyze: bool = True) -> None:
...
+        self._excludes: List[Tuple[Union[int, str], ...]] = []
         if version_str and analyze:
             self._analyze_specifiers()
 
     def _analyze_specifiers(self) -> None:
         # XXX: Prerelease or postrelease specifiers will fail here, but I guess we can
         # just ignore them for now.
         lower_bound, upper_bound = self.MIN_VERSION, self.MAX_VERSION
-        excludes = set()  # type: Set[Tuple[Union[int, str], ...]]
+        excludes: Set[Tuple[Union[int, str], ...]] = set()
```
danieleades, line 95: might be nice to tidy this up with a type alias? what exactly is represented by a `Tuple[Union[int, str], ...]` here?

frostming: Each item is a version tuple, which accepts the following forms:

- `(3, 8, 0)`
- `(3, 8, '*')`

frostming:
> I'd like to add mypy linting too, that's my eventual goal. but there's a way to go before that!

No hurry, just a reminder though.

> Will probably start chipping away at those in separate pull requests, and keep this one focused on replacing the type comments only.

That is okay, will start reviewing later

pdm/models/specifiers.py
danieleades: ah I see, so something along the lines of 

```python
VersionIncrement = Union[int, Literal['*']]
SemVer = Tuple[VersionIncrement, VersionIncrement, VersionIncrement]

...

excludes: Set[SemVer]
```

I'll have think about a tidy way to capture this.

almost worth a class.

danieleades: i feel like `pdm/models/specifiers.py` is crying out for some custom classes to encapsulate the version specifications. The code here looks challenging to debug/maintain

like this - https://docs.rs/semver/0.11.0/semver/struct.Version.html
and this - https://docs.rs/semver/0.11.0/semver/struct.VersionReq.html

I'm not planning to touch that in this pull request. will resolve this conversation for now

frostming: Looks good. But be noted Literal is not supported in Python 3.7

frostming (changes requested):
LGTM! Would you mind adding a news entry `298.misc.md` describing what you did in this PR?

danieleades:
> LGTM! Would you mind adding a news entry `news/298.misc.md` describing what you did in this PR?

sure. Is that what you were after?

(i've not used `towncrier` before)

frostming (approved):

frostming:
No more to do, that is it

PR #321, opened 2021-03-21
Typing/models.readers

danieleades:
## Pull Request Check List

refactor `models.readers` to improve typing support


danieleades:
`models.readers.SetupReader` has a really similar pattern to what's used in the `iostream` module - a class is used as a namespace for a bunch of stateless functions. Was tempted to turn these all into free functions, but i thought that might have been a bit drastic for this pull request.

danieleades:
here are the remaining type errors in this file (`mypy pdm/ | grep readers`)

```
pdm/models/readers.py:84: error: Incompatible types in assignment (expression has type "Optional[List[Any]]", variable has type "List[stmt]")
pdm/models/readers.py:171: error: Unsupported left operand type for + ("None")
pdm/models/readers.py:171: note: Left operand is of type "Optional[List[Any]]"
pdm/models/readers.py:185: error: "expr" has no attribute "id"
pdm/models/readers.py:205: error: Incompatible types in assignment (expression has type "Optional[Call]", variable has type "Tuple[Optional[Call], Optional[List[Any]]]")
pdm/models/readers.py:207: error: Unsupported operand types for + ("List[Any]" and "None")
pdm/models/readers.py:207: note: Right operand is of type "Optional[List[Any]]"
pdm/models/readers.py:209: error: Incompatible return value type (got "Tuple[Tuple[Optional[Call], Optional[List[Any]]], List[Any]]", expected "Tuple[Optional[Call], Optional[List[Any]]]")
pdm/models/readers.py:215: error: Need type annotation for 'install_requires' (hint: "install_requires: List[<type>] = ...")
pdm/models/readers.py:244: error: "expr" has no attribute "s"
pdm/models/readers.py:250: error: "expr" has no attribute "s"
pdm/models/readers.py:258: error: Need type annotation for 'extras_require' (hint: "extras_require: Dict[<type>, <type>] = ...")
pdm/models/readers.py:288: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "expr")
pdm/models/readers.py:291: error: Item "expr" of "Optional[expr]" has no attribute "s"
pdm/models/readers.py:291: error: Item "None" of "Optional[expr]" has no attribute "s"
pdm/models/readers.py:291: error: "expr" has no attribute "s"
pdm/models/readers.py:300: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "expr")
pdm/models/readers.py:303: error: Item "expr" of "Optional[expr]" has no attribute "s"
pdm/models/readers.py:303: error: Item "None" of "Optional[expr]" has no attribute "s"
pdm/models/readers.py:303: error: "expr" has no attribute "s"
```

pdm/models/readers.py
```diff
@@ -1,103 +1,105 @@
...
+            self.extras_require = other.extras_require
+        if other.python_requires:
+            self.python_requires = other.python_requires
+
+    def as_dict(self) -> Dict[str, Any]:
+        return asdict(self)
+
+    @classmethod
+    def from_directory(cls, dir: Path) -> "Setup":
+        return _SetupReader.read_from_directory(dir)
```
frostming, line 33: Why prefer exposing `Setup` instead of `SetupReader`? You know, it may involve changing the reference in other modules as well.

frostming (commented):
I think staticmethod or classmethod is good for this case, depending on the method sematic.

pdm/models/readers.py
danieleades: because i'm now exposing a strongly-typed 'Setup' class instead of a stringly-typed dictionary.

I've updated both external modules that reference this

danieleades: these changes bring the number of mypy errors down from 44 to 20. Still a few to go though! - https://github.com/frostming/pdm/pull/321#issuecomment-803589924

in the previous implementation the 'setup' is stored in a `Dict[str, Any]`, which means type inference doesn't work in any function that indexes any of the fields.

Now I could still have imported the `SetupReader` in other modules instead of `Setup`, but i think this is a neater abstraction. the `from_directory` static method on `Setup` is effectively a constructor, and the calling code doesn't need to know anything about the `SetupReader`, which is now effectively just a helper class

danieleades:
rebased on master

danieleades:
all the remaining type errors in this file are related to the `setup.py` parsing. I'm not exactly sure how to proceed, given the black magic here...

```
pdm/models/setup.py:88: error: Incompatible types in assignment (expression has type "Optional[List[Any]]", variable has type "List[stmt]")
pdm/models/setup.py:175: error: Unsupported left operand type for + ("None")
pdm/models/setup.py:175: note: Left operand is of type "Optional[List[Any]]"
pdm/models/setup.py:189: error: "expr" has no attribute "id"
pdm/models/setup.py:209: error: Incompatible types in assignment (expression has type "Optional[Call]", variable has type "Tuple[Optional[Call], Optional[List[Any]]]")
pdm/models/setup.py:211: error: Unsupported operand types for + ("List[Any]" and "None")
pdm/models/setup.py:211: note: Right operand is of type "Optional[List[Any]]"
pdm/models/setup.py:213: error: Incompatible return value type (got "Tuple[Tuple[Optional[Call], Optional[List[Any]]], List[Any]]", expected "Tuple[Optional[Call], Optional[List[Any]]]")
pdm/models/setup.py:248: error: "expr" has no attribute "s"
pdm/models/setup.py:254: error: "expr" has no attribute "s"
pdm/models/setup.py:292: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "expr")
pdm/models/setup.py:295: error: Item "expr" of "Optional[expr]" has no attribute "s"
pdm/models/setup.py:295: error: Item "None" of "Optional[expr]" has no attribute "s"
pdm/models/setup.py:295: error: "expr" has no attribute "s"
pdm/models/setup.py:304: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "expr")
pdm/models/setup.py:307: error: Item "expr" of "Optional[expr]" has no attribute "s"
pdm/models/setup.py:307: error: Item "None" of "Optional[expr]" has no attribute "s"
pdm/models/setup.py:307: error: "expr" has no attribute "s"
```

danieleades:
rebased on master

danieleades:
i've added decorators to ignore the typing errors in some of these functions. it's not an ideal solution, but it *will* suppress the errors and get you a step closer to using mypy in CI/pre-commit.

seems like parsing `setup.py` files should be a reasonably common problem. are there no third-party libraries that can be used?

frostming (approved):

PR #322, opened 2021-03-21
add return type annotation to spinner context manager

danieleades:
## Pull Request Check List


frostming:
Sorry to revert that when merging. would a protocol be possible to denote the return type of the context manager? You can refer to DummySpinner for the necessary interfaces to expose

danieleades:
> Sorry to revert that when merging. would a protocol be possible to denote the return type of the context manager? You can refer to DummySpinner for the necessary interfaces to expose

it would help *slightly*! a `Protocol` is basically Python's take on an interface/trait, right? it's python 3.8+ only, is that an issue for you?

frostming:
> > Sorry to revert that when merging. would a protocol be possible to denote the return type of the context manager? You can refer to DummySpinner for the necessary interfaces to expose
> 
> 
> 
> it would help *slightly*! a `Protocol` is basically Python's take on an interface/trait, right? it's python 3.8+ only, is that an issue for you?

It can be imported from typing_extensions. You can put it in a try except in _types.py

PR #326, opened 2021-03-22
Extract the logic of finding interpreters to a method

frostming:
## Pull Request Check List

As title

### DIFF
--- a/pdm/cli/actions.py
+++ b/pdm/cli/actions.py
@@ -417,7 +417,7 @@ def do_use(project: Project, python: str, first: bool = False) -> None:
     The python can be a version string or interpreter path.
     """
     python = python.strip()
-    found_interpreters = list(project.find_interpreters(python))
+    found_interpreters = list(dict.fromkeys(project.find_interpreters(python)))
     if not found_interpreters:
         raise NoPythonVersion(
             f"Python interpreter {python} is not found on the system."
--- a/pdm/cli/actions.py
+++ b/pdm/cli/actions.py
@@ -14,7 +14,6 @@ from pdm import termui
 from pdm.cli.utils import (
     check_project_file,
     find_importable_files,
-    find_python_in_path,
     format_lockfile,
     format_resolution_impossible,
     save_version_specifiers,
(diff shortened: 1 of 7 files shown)


pdm/project/core.py
```diff
@@ -515,3 +516,25 @@ def make_candidate_info_cache(self) -> CandidateInfoCache:
 
     def make_hash_cache(self) -> HashCache:
         return HashCache(directory=self.cache("hashes").as_posix())
+
+    def find_interpreters(self, python_spec: str) -> Iterable[str]:
```
danieleades, line 520: why return a `str` here instead of a `Path`?

pdm/utils.py
```diff
@@ -412,3 +412,27 @@ def is_venv_python(interpreter: os.PathLike) -> bool:
         else:
             return True
     return False
+
+
+def find_python_in_path(path: os.PathLike) -> Optional[str]:
+    """Find a python interpreter from the given path, the input argument could be:
+
+    - A valid path to the interpreter
+    - A Python root diretory that contains the interpreter
```
danieleades, line 421: typo - "A Python root **directory**..."

pdm/utils.py
```diff
@@ -412,3 +412,27 @@ def is_venv_python(interpreter: os.PathLike) -> bool:
         else:
             return True
     return False
+
+
+def find_python_in_path(path: os.PathLike) -> Optional[str]:
```
danieleades, line 417: why return a `str` instead of a `Path`?

pdm/project/core.py
```diff
@@ -515,3 +516,25 @@ def make_candidate_info_cache(self) -> CandidateInfoCache:
...
+                python_path = find_python_in_path(python_spec)
+                if python_path:
+                    yield python_path
+                return
+            args = [python_spec]
+        else:
+            args = [int(v) for v in python_spec.split(".") if v != ""]
+        finder = pythonfinder.Finder()
+        for entry in finder.find_all_python_versions(*args):
+            yield entry.path.as_posix()
```
danieleades, line 540: are you getting much benefit from using a generator here?

seems to me all the lists in the method are allocated to memory, so returning each item one by one as an iterator doesn't offer any performance benefits.

It *would* if the upstream method `finder.find_all_python_versions` returned an iterator, but it doesn't- it returns an allocated list.

frostming: Not much, just provides a more neat interface for subclass methods:

```python
def find_interpreters(self, python_spec: str) -> Iterable[str]:
    yield some
    yield another
    yield from super().find_interpreters(python_spec)
```
Instead of:
```python
def find_interpreters(self, python_spec: str) -> Iterable[str]:
    results = []
    results.append(some)
    results.append(another)
    results.extend(super().find_interpreters(python_spec))
    return results
```

pdm/cli/actions.py
```diff
@@ -417,7 +417,7 @@ def do_use(project: Project, python: str, first: bool = False) -> None:
     The python can be a version string or interpreter path.
     """
     python = python.strip()
-    found_interpreters = list(project.find_interpreters(python))
+    found_interpreters = list(dict.fromkeys(project.find_interpreters(python)))
```
danieleades, line 420: `found_interpreters = set(project.find_interpreters(python))`?

frostming: Nope, the difference is `dict.fromkeys()` preserves the order while `set` doesn't.

PR #333, opened 2021-03-24
Retrieve flit's version and description by parsing the ast

frostming:
## Pull Request Check List

Related issue https://github.com/pdm-project/pdm-pep517/issues/10

### DIFF
--- a/news/333.feature.md
+++ b/news/333.feature.md
@@ -1,0 +1,1 @@
+Retrieve the version and description of a flit project by parsing the AST of the main file.
--- a/pdm/formats/flit.py
+++ b/pdm/formats/flit.py
@@ -1,5 +1,7 @@
+import ast
 import re
 from pathlib import Path
+from typing import Optional, Tuple
 
 import tomlkit
 import tomlkit.exceptions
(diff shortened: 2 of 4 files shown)


ntninja:
Nit: Please emit a warning if no description could be retrieved (`self._data["description"] is None`). Somebody might have done `from ._core import __doc__` or something (which works in flit) and they should know if PDM failed to pick it up.

ntninja:
Your PR doesn't account for the common idiom of `from .version import __version__` in main. Many projects use (used?) this to be able to do `execfile(os.path.dirname(__file__) + "/name/version.py")` in their setuptools and sphinx configuration files for retrieving the just version number without importing everything. If your planning on keeping `version.from` then implementing this may be worthwhile since it allows to losslessly convert the version discovery logic from flit to pdm, otherwise emitting a warning when the version number is not found or found but now hard-coded into `pyproject.toml` would be helpful.

frostming:
Static is better than dynamic, if we are to support dynamic version and description, what about `__author__` and `__license__` which is also a commonly used pattern. They are common before because `setup.py` itself is a dynamic python script where you can play many magic their, such as conditional requirements, dynamically generated README. But when we come to PEP 621, static metadata is recommended, though dynamic fields are not obsoleted fully(they can appear in `dynamic`). So let's only keep the `from.version` which is from pdm 0.x versions.

I have submitted a PR to emit warnings when those fields can't be determined when importing, give it a try and feedbacks are welcome.

ntninja:
The warning LGTM, but in case where conversion succeeds a short notice about the values now being managed in `pyproject.toml` rather than the Python file would also be helpful to new users, I think.

Also it copying the version number from `__version__` into PEP-621 data, rather then creating a `from.version` reference, is done on purpose to encourage the use of static metadata?
If so, it would be nice if the notice also included a link to https://pdm.fming.dev/pyproject/#determine-the-package-version-dynamically for people preferring to keep the previous behaviour for that field at least.

(Saying this as somebody who thought the way it previously converted the version was a bug, particularly after I discovered how it's actually possible to keep the existing behaviour there.)

PR #335, opened 2021-03-24
ignore typing in functions that index distlib.metadata._data

danieleades:
closes #334

**this change wipes out 28 mypy errors.**

this is justified on the basis that the datastructure being indexed is `distlib.metadata._data`, and the structure is relatively stable.

PR #336, opened 2021-03-24
bump version of pdm-pep517

danieleades:
bump the version of `pdm-pep517` to resolve CI errors

as described in https://github.com/pdm-project/pdm/pull/335#issuecomment-805604042


danieleades:
@frostming, doesn't seem to have done the trick

frostming:
Oh, never mind, I will handle this

PR #341, opened 2021-03-24
tidy some import logic

danieleades:


pdm/_types.py
```diff
@@ -1,9 +1,10 @@
+import sys
 from typing import Dict, List, NamedTuple, Tuple, Union
 
-try:
+if sys.version_info < (3, 8):
```
danieleades, line 4: this pattern has two advantages

- mypy understands `if sys.version_info...` imports
- it becomes far easier to see when this import logic should be removed (ie when python < 3.8 support is dropped)

pdm/_vendor/halo/_utils.py
```diff
@@ -4,10 +4,7 @@
 import codecs
 import platform
 import six
-try:
```
danieleades, line 7: i *think* this is no longer required in python 3.7+?

i'll let the CI confirm that

pdm/models/pip_shims.py
```diff
@@ -83,13 +79,6 @@ def get_flag(var, fallback, expected=True, warn=True):
             "WITH_PYMALLOC", is_cpython, warn=False
         ):
             m = "m"
-        if python_version < (3, 3) and get_flag(
-            "Py_UNICODE_SIZE",
-            sys.maxunicode == 0x10FFFF,
-            expected=4,
-            warn=False,
-        ):
-            u = "u"
```
danieleades, line 92: I haven't got a single clue what this does, but it's unreachable

uranusjr: It’s for marking whether `u""` literals are supported (added in Python 3.3), so yeah irrelevant now.

pdm/models/pip_shims.py
```diff
@@ -30,17 +30,13 @@
...
 from pip._internal.utils.hashes import FAVORITE_HASH, STRONG_HASHES
 from pip._internal.utils.temp_dir import global_tempdir_manager
 from pip._internal.utils.urls import path_to_url, url_to_path
 from pip._internal.vcs.versioncontrol import VcsSupport
 
-try:
-    from pip._internal.utils.compatibility_tags import get_supported
-except ImportError:
-    from pip._internal.pep425tags import get_supported
-
```
uranusjr, line 43: The new import path was introduced in pip 20.0. Perhaps a `pip.__version__` check is needed?

frostming: I agree to just remove the import guard as the proposed change.

pdm/_vendor/halo/_utils.py
```diff
@@ -4,10 +4,7 @@
...
-    from shutil import get_terminal_size
-except ImportError:
-    from backports.shutil_get_terminal_size import get_terminal_size
+from shutil import get_terminal_size
```
frostming, line 7: We can just skip the files under `_vendor/` directory and keep this untouched. It will affect our vendoring workflow

frostming (changes requested):

Issue #347, opened 2021-03-28
MacOS failed 1 test in current master


ulwlu:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

**Make sure you run commands with `-v` flag before pasting the output.**

## Steps to reproduce

On MacOS, run `pdm run test`.

## Actual behavior

Failed 13 tests.

<img width="1258" alt="ss 4" src="https://user-images.githubusercontent.com/41639488/112740980-708f5f80-8fbc-11eb-8eb5-feb43847898e.png">

![ss 2](https://user-images.githubusercontent.com/41639488/112740983-7422e680-8fbc-11eb-8d6f-931c74c7c3fb.png)

## Expected behavior

Clear all tests.

## Environment Information

```bash
PDM version:        1.4.4
Python Interpreter: /usr/local/opt/python@3.9/bin/python3.9 (3.9.0)
Project Root:       /Users/kohei.murakami/.ghq/github.com/ulwlu/pdm
{
  "implementation_name": "cpython",
  "implementation_version": "3.9.0",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "19.6.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 19.6.0: Thu Oct 29 22:56:45 PDT 2020; root:xnu-6153.141.2.2~1/RELEASE_X86_64",
  "python_full_version": "3.9.0",
  "platform_python_implementaiton": "CPython",
  "python_version": "3.9",
  "sys_platform": "darwin"
}
```

## Remarks

I think this is related to https://github.com/pdm-project/pdm/pull/109#issuecomment-613786780 .

I'm now trying to fix tests, but would love to know the proposal is correct.

First of all, ` tests/test_utils.py::test_find_python_in_path` is failed because brew-installed-python's `sys.prefix` and `sys.executable` are different. So it always fails.

```
sys.executable: /usr/local/Cellar/python@3.9/3.9.0_1/bin/python3.9

sys.prefix: /usr/local/Cellar/python@3.9/3.9.0_1/Frameworks/Python.framework/Versions/3.9/bin/python3
```

Will it be fine if I change this test into like "find the sys.executable path up to version just like `/usr/local/Cellar/python@3.9`, and compare it to sys.prefix with `startwith`" if MacOS?

frostming:
That change makes sense to me, but what about other test cases? can you attach the entire test log?

ulwlu:
[...]

PR #348, opened 2021-03-28
Fix test_find_python_in_path: Check only up to version only when MacOS

ulwlu:
## Pull Request Check List

- Fixes https://github.com/pdm-project/pdm/issues/347.
- On MacOS, python's `sys.prefix` and `sys.executable` are different. So this tests must have failed.
- I changed this test into "find the sys.executable path up to version just like `/usr/local/Cellar/python@3.9`, and compare it to sys.prefix with `startwith`" only when MacOS.

ss 5

I checked this test with ubuntu on docker as well, and it runs correctly.


frostming (approved):

frostming:
Thanks for the contribution

PR #351, opened 2021-03-29
Feature: support groups in dev-dependencies

### LINKED ISSUE #345: Can the `dev-dependencies` field be in the PEP621 `[project]` section?
I wonder if it is okay to have the `dev-dependencies` field in the PEP621 `[project]` section of `pyproject.toml`. This came to my mind when reading [this Python packaging comment](https://discuss.python.org/t/pep-621-round-3/5472/124).

>  No tools may add fields to this table which are not defined by this PEP or subsequent PEPs.

-- https://www.python.org/dev/peps/pep-0621/#table-name

And indeed the `dev-dependencies` field name is not specified by PEP621 so far, which would mean that it is not be allowed.

frostming:
## Pull Request Check List

- [x] Move out `dev-dependencies`, `includes`, `excludes`, `package-dir` from `[project]` table to `[tool.pdm]`, close #345 
- [x] Update `pdm-pep517` to support both formats (transition solution)
- [x] Auto-migrate the `pyproject.toml` to the new format with a message, the change should not break anything.
- [x] Change the `dev-dependencies` to a group of dependencies
- [x] The dev-dependencies of the old format will be transferred to the new table under **dev** group by the migration. This may require possible relock as the lock file hash changes.
- [x] Adjust `add`, `sync`, `install`, `remove`, and `update` commands to support the new behavior, in a backward-compatible way.


docs/docs/pyproject.md
```diff
@@ -125,27 +87,7 @@ $ pdm install -s socks
 
 `-s` option can be given multiple times to include more than one groups.
```
pohlt, line 88: typo: "group."

docs/docs/usage/dependency.md
```diff
@@ -36,23 +35,44 @@ For details of the meaning of each field in `pyproject.toml`, please refer to [P
...
+
+```toml
+[tool.pdm.dev-dependencies]
+test = ["pytest"]
+```
+
+For backward-compatibility, if `-s/--section` is not given, dependencies will go to `dev` group under `[tool.pdm.dev-dependencies]` by default.
+
+!!! NOTE
+    The group names in `[tool.pdm.dev-dependencies]` MUST NOT conflict with those in `[project.optional-dependencies]`.
```
pohlt, line 74: What exactly is considered a "conflict"?
* A group name appears in both sections (independent of its content)?
* A group name appears in both sections with different content?

frostming: A group name appears in both sections (independent of its content)

This is required because currently the `sections` in `pdm.lock` can't distinguish whether from dev deps or optional deps

pohlt: Then I would explicitly state it like this:

> A group name MUST NOT appear in both `[tool.pdm.dev-dependencies]` and `[project.optional-dependencies]`.

... or something similar.

docs/docs/pyproject.md
```diff
@@ -123,29 +85,9 @@ To install a group of optional dependencies:
 $ pdm install -s socks
```
pohlt, line 85: Sorry, this is my first GitHub-base review, so I'm struggling a little bit with the process.

Anyway, above is still an example of `project.optional-dependencies` containing a `test` group which would now fit better into the `dev-dependencies`, right?

pohlt:
So if I understood the changes in the README correctly, there are now two flavors of grouped dependencies:
1. what is typically called "extra dependencies" (for extra functionality)
2. "dev dependencies" (only required for development)

The first is selected using `-s`, the second with `-d`. Is my summary correct?
I really like this concept, because it addresses all the raised concerns in #304.
Since this could be considered a breaking change, a mayor version bump might make sense (or you could say that everything below 1.x is a free-for-all).

frostming:
To make the behavior change more clear, let's do it in examples:

## New behavior
### install/sync

```bash
# Install all default dependencies:
$ pdm install
# Install default + web optional dependencies:
$ pdm install -s web
# Install default + all dev dependencies
$ pdm install -d
# Install web dependencies ONLY(without default dependencies):
$ pdm install --no-default -s web
```

### add

```bash
# Add to extra dependencies
$ pdm add -s web h11
# Add to test group dev dependencies
$ pdm add -ds test pytest
```

### update/remove
```bash
# Update all default
$ pdm update
# Update all dev
$ pdm update -d
# Update section web
$ pdm update -s web
# Update dev section test
$ pdm update -s test
# Plus, for update command, package name(s) can be supplied.
```

frostming:
## Old behavior

### install/sync
```bash
# Install all default dependencies:
$ pdm install
# Install default + web optional dependencies:
$ pdm install -s web
# Install default + all dev dependencies
$ pdm install -d
# Install web dependencies ONLY(without default dependencies):
$ pdm install --no-default -s web
```
As you can see, there is no difference in the behavior between new and old, because `-d` without specified sections in the new behavior means install ALL dev-dependencies.

### add

```bash
# Add to extra dependencies
$ pdm add -s web h11
# invaid combination,  -s test is IGNORED, add to dev
$ pdm add -ds test pytest
```
For the second case, since old behavior just ignores `-s test`, so the effect is the same, add it to dev dependencies.
### update/remove
```bash
# Update all default
$ pdm update
# Update all dev
$ pdm update -d
# Update section web
$ pdm update -s web
# Invalid combination, -d is ignored, update test section
$ pdm update -ds test
# Plus, for update command, package name(s) can be supplied.
```

pohlt:
Thanks for the examples. I think I now understand the concept.

For newcomers, it would be very helpful if the README states early and very clearly that
- there are two flavors of dependencies,
- their roles,
- their corresponding groups,
- a guidance when to use which and of course
- how to address them with PDM (`-s` and `-ds` flags).

pohlt:
Ah, and the README (and ideally the PDM code) should be consistent when addressing those two dependency flavors: "extra dependencies" and "dev dependencies"?

docs/docs/usage/dependency.md
```diff
@@ -77,6 +97,29 @@ To update the specified package(s):
...
+
+To update a given package in the specified section:
+
+```console
+$ pdm update -s security cryptography
+```
+
+If the section is not given, PDM will search for the requirement in the default dependencies set and raises an error if none is found.
+
+To update package in development dependencies:
```
pawamoy, line 114:
```suggestion
To update packages in development dependencies:
```

docs/docs/usage/dependency.md
```diff
@@ -77,6 +97,29 @@ To update the specified package(s):
...
+
+```console
+# Update all dev-dependencies
+$ pdm update -d
+# Update the specified group of dev-dependencies
+$ pdm update -ds test pytest
```
pawamoy, line 120: I think you meant `pdm update -ds test` (without `pytest`)?

pohlt:
So if `-d` installs all dev dependencies, is there something equivalent for all extra dependencies? If not, do we need it? I'm not sure.

docs/docs/usage/dependency.md
```diff
@@ -91,15 +134,31 @@ which is given by `--update-<strategy>` option:
...
 
-## Synchronize the project packages with lock file
+## Install the packages pinned in lock file
 
 There are two similar commands to do this job with a slight difference:
 
+```console
+# Install all default dependencies:
+$ pdm install
+# Install default + web optional dependencies:
```
pawamoy, line 152:
```suggestion
# Install default + web optional dependencies
```

(consistency)

docs/docs/usage/dependency.md
```diff
@@ -91,15 +134,31 @@ which is given by `--update-<strategy>` option:
...
+$ pdm remove -ds test pytest-cov
 ```
 
-## Synchronize the project packages with lock file
+## Install the packages pinned in lock file
 
 There are two similar commands to do this job with a slight difference:
 
+```console
+# Install all default dependencies:
```
pawamoy, line 150:
```suggestion
# Install all default dependencies
```

(consistency)

docs/docs/usage/dependency.md
```diff
@@ -91,15 +134,31 @@ which is given by `--update-<strategy>` option:
...
+$ pdm install
+# Install default + web optional dependencies:
+$ pdm install -s web
+# Install default + all dev dependencies
+$ pdm install -d
+# Install web dependencies ONLY(without default dependencies):
```
pawamoy, line 156:
```suggestion
# Install web dependencies ONLY(without default dependencies)
```

(consistency)

docs/docs/usage/dependency.md
```diff
@@ -91,15 +134,31 @@ which is given by `--update-<strategy>` option:
...
+## Install the packages pinned in lock file
 
 There are two similar commands to do this job with a slight difference:
 
+```console
+# Install all default dependencies:
+$ pdm install
+# Install default + web optional dependencies:
+$ pdm install -s web
+# Install default + all dev dependencies
```
pawamoy, line 154:
```suggestion
# Install default + all dev-dependencies
```

(consistency)

docs/docs/usage/dependency.md
```diff
@@ -91,15 +134,31 @@ which is given by `--update-<strategy>` option:
 To remove existing dependencies from project file and the library directory:
 
 ```console
+# Remove requests from the default dependencies
 $ pdm remove requests
+# Remove h11 from the 'web' optional group
+$ pdm remove -s web h11
+# Remove pytest-cov from the `test` dev dependencies group
```
pawamoy, line 141:
```suggestion
# Remove pytest-cov from the `test` dev-dependencies group
```

docs/docs/usage/dependency.md
frostming: No, you can give package name(s) to update command to tell PDM only to update those given packages under the given group(Yes, group should be given or it defaults to the default group, as the same package can appear in more than 1 group)

pawamoy: May I suggest something like this then:

```
# Update a package in the specified group of dev-dependencies
```

docs/docs/usage/dependency.md
```diff
@@ -77,6 +97,29 @@ To update the specified package(s):
...
+```
+
+If the section is not given, PDM will search for the requirement in the default dependencies set and raises an error if none is found.
+
+To update package in development dependencies:
+
+```console
+# Update all dev-dependencies
+$ pdm update -d
+# Update the specified group of dev-dependencies
```
frostming, line 119:
```suggestion
# Update a package in the specified group of dev-dependencies
```

frostming:
> The first is selected using -s, the second with -ds. Is my summary correct?

Correct

Since this could be considered a breaking change, a major version bump might make sense (or you could say that everything below 1.x is a free-for-all).

> I tend to use 1.5.0 for that release, and IMO sematic version doesn't fit well for such projects as PDM. `pip`, `virtualenv`, `pipenv` all switch to CalVer, and Poetry's sematic versioning also doesn't reflect the behavior change happenning.

docs/docs/pyproject.md
frostming: Yes, just to wipe them out from the distribution metadata

docs/docs/usage/dependency.md
frostming, line 74:
```suggestion
    The same group name MUST NOT appear in both `[tool.pdm.dev-dependencies]` and `[project.optional-dependencies]` .
```

pohlt:
Would `-ds` be equivalent to `-d -s`? Otherwise, it would violate the [POSIX way of command-line arguments](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).

pohlt:
We have three types of dependencies now:
1. mandatory dependencies (without grouping; wouldn't makes sense anyway)
2. extra dependencies (with grouping)
3. dev dependencies (with grouping)

I would prefer a more consistent design to work with all three.

An option would select the type of dependency:
* dev deps: `-d`
* extra deps: `-e`
* mandatory deps: neither `-d` nor `-e`

For both `-d` and `-e` an optional section can be added. If the section is `all`, all deps of this type are handled.
If no section is given, the deps are added to a default section (or without a section if possible).

What do you think?

frostming:
>Would `-ds` be equivalent to `-d -s`? Otherwise, it would violate the POSIX way of command-line arguments.
Yes of course

> For both `-d` and `-e` an optional section can be added. 

First `-s` is being used to specify group name and we are not gonna change it. 

> If the section is `all`, all deps of this type are handled. If no section is given, the deps are added to a default section (or without a section if possible).

Yes, that was something I had in mind, except I prefer a special value like `:all:` for that. And this value is allowed to follow `-s` for `install/sync` command only, namely, `pdm add -s :all: something` isn't a valid command. And to make the command shorter and be compatible with the old behavior, one can omit `-s :all:` in `pdm instal -ds :all:` so that it equals to `pdm install -d` which installs all dev dependencies.

pohlt:
Still, I think that having a dedicated option for specifying extra deps makes sense. Most developers have to deal more with dev deps than extra deps, so I don't see the point to make extra deps quicker to work with.

frostming:
> I'm wondering if the current implementation would allow installing both extras and specific dev-deps groups in one command:
> 
> 
> 
> ```bash
> 
> pdm install -s async -ds docs
> 
> ```

Unfortunately, no.

pawamoy:
> Unfortunately, no.

Well that's totally fine. I'm OK with typing one more command line 😄

frostming:
> Most developers have to deal more with dev deps than extra deps, so I don't see the point to make extra deps quicker to work with.

Yes, for dev dependencies only, not shortcut for all extra dependencies

frostming:
> Yes, for dev dependencies only, not shortcut for all extra dependencies

Added a special value `:all` to include all groups under this species.

```bash
# Install all optional-dependencies
$ pdm install -s:all
# Install all dev-dependencies
$ pdm install -ds:all
# Shortcut for the above command
$ pdm install -d
```

pohlt:
> > Most developers have to deal more with dev deps than extra deps, so I don't see the point to make extra deps quicker to work with.
> 
> Yes, for dev dependencies only, not shortcut for all extra dependencies

Hmm, I guess I didn't explain it well. Let me rephrase it:
IMHO, the interface (and its documentation) would be easier to understand if pdm had one option to signal that you are dealing with dev deps (`-d`) and another one to deal with extra deps (e.g. `-e`). If both are missing, mandatory deps are addressed.
The `-s` option would then be orthogonal to those two options and could be used with both of them.

With the current interface it's:
* mandatory deps: no `-s`, no `-d`
* extra deps: `-s`, no `-d`
* dev deps: `-d` and `-s`

which is less explicit. How would one add an extra dev without using a special section?

frostming:
> The -s option would then be orthogonal to those two options and could be used with both of them.

Do you mean to use `-e -s` to address extra deps? But to deal with extra deps, `-s` must be given since they are grouped dependencies, it will make `-e` a redundant functionless flag-only option. `-s` without `-d means to address extra deps, because mandatory deps don't support grouping.

> How would one add an extra dev without using a special section?
Sorry, I didn't quite get what you want by this

pohlt:
> > The -s option would then be orthogonal to those two options and could be used with both of them.
> 
> Do you mean to use `-e -s` to address extra deps? But to deal with extra deps, `-s` must be given since they are grouped dependencies, it will make `-e` a redundant functionless flag-only option. `-s` without `-d` means to address extra deps, because mandatory deps don't support grouping.
> 
> > How would one add an extra dev without using a special section?
> 
> Sorry, I didn't quite get what you want by this

I didn't know that extra deps _have to_ be grouped. Using the `-e` option would still be more explicit (though redundant because you would always have to use an `-s`). Never mind. 😄

frostming:
I think we are ready to merge this. A beta release will be out for testing in the next week. Thanks for the feedback and discussions.

Issue #261, opened 2021-02-13
Typing


danieleades:
I'm hoping to gradually improve the 'typing' in this library to the point where it can be linted using Mypy.

There's quite a lot to do first! This library uses a few loosely-typed data structures which lead to a lot of type errors. That will possibly involve some refactoring (or heavy use of `if not None` checks everywhere).

I'll use this is as a parent task for various efforts to improve the typing-

- [x] #298 
- [x] #301
- [x] #321 
- [x] #322
- [x] #335
- [ ] #427

frostming:
Thanks for pointing it out, I actually have no enough energy to enforce type annotations for the whole codebase. Contributions are welcomed of course.

danieleades:
for a sense of scale, here's the current output from mypy-
<details>
<summary>mypy output</summary>

```
pdm/formats/setup_py.py:13: error: Cannot find implementation or library stub for module named 'pdm.pep517.base'
pdm/project/metadata.py:3: error: Cannot find implementation or library stub for module named 'pdm.pep517.metadata'
pdm/models/readers.py:12: error: Need type annotation for 'DEFAULT'
pdm/models/readers.py:43: error: Incompatible return value type (got "Dict[str, Union[Dict[Any, Any], List[Any], None]]", expected "Dict[str, Union[List[Any], Dict[Any, Any]]]")
pdm/models/readers.py:54: error: Cannot find implementation or library stub for module named 'pdm.pep517.metadata'
pdm/models/readers.py:80: error: Incompatible types in assignment (expression has type "Optional[List[Any]]", variable has type "List[stmt]")
pdm/models/readers.py:82: error: Incompatible return value type (got "Dict[str, Union[Dict[Any, Any], List[Any], None]]", expected "Dict[str, Union[List[Any], Dict[Any, Any]]]")
pdm/models/readers.py:87: error: Incompatible types in assignment (expression has type "List[str]", target has type "Optional[str]")
pdm/models/readers.py:88: error: Incompatible types in assignment (expression has type "Dict[str, List[Any]]", target has type "Optional[str]")
pdm/models/readers.py:93: error: Incompatible return value type (got "Dict[str, Optional[str]]", expected "Dict[str, Union[List[Any], Dict[Any, Any]]]")
pdm/models/readers.py:111: error: Need type annotation for 'extras_require' (hint: "extras_require: Dict[<type>, <type>] = ...")
[...]

PR #363, opened 2021-03-31
refactor/typing

frostming:
## Pull Request Check List


frostming:
I would like to stop here, running `mypy pdm/` still outputs 379 errors. I will leave the tasks to other contributors #261

PR #367, opened 2021-04-01
Add automation script to generate completions on release

frostming:
## Pull Request Check List


frostming:
Hi, @Aloxaf , I updated the zsh script according to PEP 621 and latest CLI. But I have trouble extracting the pdm packages. It seems it is not easy to do it with pure pattern matching, unless we actually parse the TOML document. Concretely, the following formats are legal:

```toml
# All in one line
dependencies = ["requests", "click==7.0"]
# Each at its own line
dependencies = [
    "requests",
    "click==7.0"
]
# Or hybrid
dependencies = [
    "requests", "requests-toolbelt",
    "click", "click-completion"
]
```
How do I handle it?

Aloxaf:
Oh, I haven't noticed that the format has changed.

I think the best way is to call python:

```python
packages=($(command python -c 'import toml; d=toml.load("pyproject.toml"); print(*d["project"]["dependencies"])'))
```

Or with pure zsh?:

```zsh
setopt local_options re_match_pcre
[[ $(<pyproject.toml) =~ 'dependencies *= *\[((.|\n)*?)\]' ]]
packages=(${=${(Q)match//,/}})
```

pdm/cli/completions/pdm.zsh
```diff
@@ -0,0 +1,320 @@
...
+      ;;
+    export)
+      arguments+=(
+        {-g,--global}'[Use the global project, supply the project root with `-p` option]'
+        {-f+,--format+}"[Specify the export file format]:format:(pipfile poetry flit requirements setuppy)"
+        "--without-hashes[Don't include artifact hashes]"
+        {-o+,--output+}"[Write output to the given file, or print to stdout if not given]:output file:_files"
+        {-s+,--section+}'[(MULTIPLE) Specify section(s) of optional-dependencies or dev-dependencies(with -d)]:section:_pdm_sections'
+        {-d,--dev}"[Select dev dependencies]"
+        "--no-default[Don't include dependencies from default seciton]"
```
Lyonsclay, line 130: typo -- should be "...from default section]"

frostming:
> Oh, I haven't noticed that the format has changed.
> 
> I think the best way is to call python:
> 
> ```python
> packages=($(command python -c 'import toml; d=toml.load("pyproject.toml"); print(*d["project"]["dependencies"])'))
> ```
> 
> Or with pure zsh?:
> 
> ```shell
> setopt local_options re_match_pcre
> [[ $(<pyproject.toml) =~ 'dependencies *= *\[((.|\n)*?)\]' ]]
> packages=(${=${(Q)match//,/}})
> ```

Great, that helps. then how to catch all the matches:

```toml
dependencies = [...]

[project.optional-dependenceis]
extra1 = [...]
extra2 = [...]
extra3 = [...]
```

frostming:
It seems working like a charm! may need another look from @Aloxaf

pdm/cli/completions/pdm.zsh
```diff
@@ -1,6 +1,7 @@
 #compdef pdm
 
 PDM_PIP_INDEXES=($(command pdm config pypi.url))
+PYTHON="%{python_executable}"
```
Aloxaf, line 3: As a global variable, `PYTHON` is a too common name. Users may use this name for their own plugin.
I suggest something like `PDM_PYTHON` so that it won't be override/access by other script.

Aloxaf:
The rest LGTM.

PR #369, opened 2021-04-01
Python runs in isolated mode to avoid importing user packages

linw1995:
## Pull Request Check List

Python commands  in `pdm.models.in_process` which invoked via `subprocess` need to run in isolated mode [-I](https://docs.python.org/3/using/cmdline.html#id2), to avoid importing user packages. (like the `json.py` file in the current directory, etc)

### DIFF
--- a/news/369.bugfix.md
+++ b/news/369.bugfix.md
@@ -1,0 +1,1 @@
+Python runs in isolated mode via subprocess to avoid importing user packages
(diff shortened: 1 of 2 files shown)


frostming:
I think it doesn't make difference, those functions only import and use stdlib, which doesn't exist in user site. Correct me if I am wrong.

linw1995:
@frostming Yes, it affects nothing. Only avoids importing user codes. One day may need to import code from third parties. It's better to select importing via executing `python -Ic "import sys;sys.path.insert(0, "..."); import bar"`

frostming (approved):

frostming:
🎉

PR #379, opened 2021-04-04
Get the correct Python ABI tag of selected interpreter 

### LINKED ISSUE #378: PackageFinder.find_requirement always fallback to sdist because being given the wrong ABI tag

## Steps to reproduce

1. For example, use the Py3.9 to install PDM
2. use PDM to set up a project with Py3.9
3. execute `pdm add lxml` to install `lxml` via wheel installation. 
4. use PDM to switch Python executable to Py3.8. `pdm use -f 3.8`
5. execute `pdm sync -v` to install `lxml`

## Actual behavior

It installs lxml via compiling from source unlike the previous case although lxml has prebuilt wheels with Py3.8.

## Expected behavior

It installs lxml via wheel installation.

## Environment Information

```bash
PDM version:        1.5.0b1.dev12+gd48c6d3
Python Interpreter: /usr/bin/python3 (3.8.2)
Project Root:       /private/tmp/lxml
Project Packages:   /private/tmp/lxml/__pypackages__/3.8
{
  "implementation_version": "3.8.2",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "20.3.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64",
  "python_full_version": "3.8.2",
  "python_version": "3.8",
  "sys_platform": "darwin"
}
```

## Hotfix

I find out what causes this problem. The `get_abi_tag` function from `pdm.models.pip_shims` is executed by the Python interpreter of PDM instead of the interpreter which we have chosen.

https://github.com/linw1995/pdm/commit/f4fb1ea5ff112d79e63f4e1996bdf75d48fdc908 

This commit works well with most kinds of Python interpreters. But the better way is executing the whole `get_abi_tag` function by the interpreter we want. To do this, we may need to endure installing extras dependencies like the wheel package. But it is possible we are able to move the codes which we need from the wheel package into the PDM package.

linw1995:
## Pull Request Check List

resolve #378 

> `get_abbr_impl` is copied from https://github.com/pypa/packaging/blob/14ad3520dc866295963c677c6813646b8a59a6a6/packaging/tags.py#L801-L811


frostming:
We could start to collect the in-process scripts in their own folder. Can you help do it in this PR?

```
pdm/models/in_process
- __init__.py    # Exports all functions here
- pep508.py
- get_abi_tag.py
```

linw1995:
Very happy to help to do it.

frostming:
This way we can directly use the `__file__` variable to get rid of `importlib.import`

frostming (approved):

frostming:
Thank you for taking care of this

Issue #407, opened 2021-04-23
NameError with wrapped setup.py (pdm.installers._editable_install)


blueyed:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

When trying to add mkdocs in editable mode it fails with a NameError due to `re` not being known when `exec`ing the code in `pdm.installers._editable_install`: https://github.com/pdm-project/pdm/blob/6baf013c4a83685148c649c0ca485397c3393048/pdm/installers/_editable_install.py#L11-L26

```
% .venv/bin/pdm add -e ~/Vcs/mkdocs -s doc -d
Adding packages to doc dev-dependencies: mkdocs
✔ 🔒 Lock successful
Changes are written to pdm.lock.
Changes are written to pyproject.toml.
Synchronizing working set with lock file: 18 to add, 0 to update, 0 to remove

  ✖ Install mkdocs 1.2.dev1 failed
  See /tmp/pdm-install-2rcus8kw.log for detailed debug log.
  [BuildError]: Call command ['/usr/bin/python', '-u', '/tmp/pdm-test/.venv/lib/python3.9/site-packages/pdm/installers/_editable_install.py', '/home/daniel/Vcs/mkdocs/setup.py', '/tmp/pdm-test/__pypackages__/3.9', '/tmp/pdm-test/__pypackages__/3.9/lib', '/tmp/pdm-test/__pypackages__/3.9/bin'] return non-zero status.
  Add '-v' to see the detailed traceback
.venv/bin/pdm add -e ~/Vcs/mkdocs -s doc -d  16.46s user 1.49s system 41% cpu 43.665 total
```

Note: this uses the `1.1.2` tag of mkdocs, with pdb 1.5.1.

Log:
```
Preparing isolated env for PEP 517 build...
running egg_info
writing mkdocs.egg-info/PKG-INFO
writing dependency_links to mkdocs.egg-info/dependency_links.txt
writing entry points to mkdocs.egg-info/entry_points.txt
writing requirements to mkdocs.egg-info/requires.txt
writing top-level names to mkdocs.egg-info/top_level.txt
warning: the 'license_file' option is deprecated, use 'license_files' instead
adding license file 'LICENSE' (matched pattern 'LICENSE')
reading manifest file 'mkdocs.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching '*.mustache' under directory 'mkdocs'
warning: no previously-included files matching '__pycache__' found under directory '*'
[...]

PR #408, opened 2021-04-23
Fix globals with editable installs wrapping of `setup.py`

blueyed:
Fixes https://github.com/pdm-project/pdm/issues/407.

## Pull Request Check List

### DIFF
--- a/pdm/installers/_editable_install.py
+++ b/pdm/installers/_editable_install.py
@@ -23,7 +23,7 @@ def install(setup_py, prefix, lib_dir, bin_dir):
         "--site-dirs={0}".format(lib_dir),
     ]
     sys.path.append(lib_dir)
-    exec(compile(code, __file__, "exec"), globals())
+    exec(compile(code, __file__, "exec"), {})
 
 
 if __name__ == "__main__":
--- a/pdm/installers/_editable_install.py
+++ b/pdm/installers/_editable_install.py
@@ -23,7 +23,7 @@ def install(setup_py, prefix, lib_dir, bin_dir):
         "--site-dirs={0}".format(lib_dir),
     ]
     sys.path.append(lib_dir)
-    exec(compile(code, __file__, "exec"))
+    exec(compile(code, __file__, "exec"), globals())
 
 
 if __name__ == "__main__":


frostming:
Would you please add a news fragment inside `news/` folder? Thanks

blueyed:
> Would you please add a news fragment inside `news/` folder? Thanks

Sure.
More importantly it should have a test though - both are there as TODO.. ;)

frostming:
Sorry for the late review. This doesn't look correct to me. The original issue is failing in "build" time not "installation" time. And the editable installation script is adjusted from `SETUPTOOLS_SHIM` which can be found in pip. The `globals()` parameter doesn't make sense. The script shouldn't inherit the caller's global variables, and without the parameter the script should just work fine.

pdm/installers/_editable_install.py
```diff
@@ -23,7 +23,7 @@ def install(setup_py, prefix, lib_dir, bin_dir):
         "--site-dirs={0}".format(lib_dir),
     ]
     sys.path.append(lib_dir)
-    exec(compile(code, __file__, "exec"))
+    exec(compile(code, __file__, "exec"), globals())
```
frostming, line 26:
```suggestion
    exec(compile(code, __file__, "exec"), {})
```
I don't want globals() to contaminate the execution globals. An empty dict just does the trick from my testing result.

blueyed: It does not fix the referenced issue, where it still fails with a `NameError` for the `import re` from the toplevel of mkdoc's setup.py.

I think it would be good to have a test for this already, which would show the issue / verify the fix, but I have not looked into the existing tests, and how it fits best yet.

frostming: I am curious in what case the script requires `globals()`, since I tried:

```python
text = 'import re\ndef foo(): print(re)\nfoo()\n'
exec(compile(text, 'main', 'exec'))
```
And it works smoothly, also, as I mentioned earlier, this snippet is actually being used by `pip` to execute all `setup.py` commands

frostming: Got it, I should move the code out of the `install()` function

blueyed:
Closing in favor of #419.

PR #411, opened 2021-04-24
Check before setting the PYTHONPATH environment variable for PEP582

### LINKED ISSUE #410: Script of enabling PEP582 globally contaminates the sys.path

The `pdm --pep582` produces `PYTHONPATH=${PEP582_SCRIPT}:${PYTHONPATH}`. When the old `PYTHONPATH` is undefined which means it is an empty string, this environment variable will be causing the side effect is every Python script run with `sys.path` containing the current working directory.

## Steps to reproduce

This problem can be reproduced without a project existence.

1. run `eval $(pdm --pep582)` to enable PEP582 globally
2. run `mkdir -p /tmp/test/bin` to create a temporary directory for test
3. run `cd /tmp/test` and run `echo "import sys; print(sys.path)" > bin/script.py`
4. run `python bin/script.py` and it will output the value of `sys.path`

You can run `unset PYTHONPATH` and run `python bin/script.py` to see the different.

## Actual behavior

```plain
['/private/tmp/test/bin', '/private/tmp/test', ...]
```
## Expected behavior

```plain
['/private/tmp/test/bin', ...]
```

## Environment Information

I create a dummy project to get below information.

```bash
PDM version:        1.5.1
Python Interpreter: /Users/linw1995/.pyenv/versions/3.9.4/bin/python3 (3.9)
Project Root:       /private/tmp/test
Project Packages:   /private/tmp/test/__pypackages__/3.9
{
  "implementation_version": "3.9.4",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "20.3.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64",
  "python_full_version": "3.9.4",
  "python_version": "3.9",
  "sys_platform": "darwin"
}
```

linw1995:
## Pull Request Check List

resolve #410

### DIFF
--- a/pdm/cli/actions.py
+++ b/pdm/cli/actions.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+import textwrap
 from argparse import Namespace
 from pathlib import Path
 from typing import Dict, Iterable, List, Optional, Sequence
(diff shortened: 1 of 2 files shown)


frostming:
I am Ok with the change. Thanks

frostming (approved):

pdm/cli/actions.py
```diff
@@ -557,11 +558,39 @@ def print_pep582_command(ui: termui.UI, shell: str = "AUTO"):
...
-        result = f"export PYTHONPATH='{lib_path}':$PYTHONPATH"
+        result = textwrap.dedent(
+            f"""
+            if [ -n "$PYTHONPATH" ]; then
+                export PYTHONPATH='{lib_path}':$PYTHONPATH
+            else
+                export PYTHONPATH='{lib_path}'
+            fi
+            """
+        ).strip()
```
blueyed, line 569: I think this would be better, with the downside of not being that readable maybe though!?
`result = f"export PYTHONPATH='{lib_path}'${{PYTHONPATH:+:${{PYTHONPATH}}}}"`

(ref: https://unix.stackexchange.com/questions/26047/how-to-correctly-add-a-path-to-path)

blueyed (commented):
Apart from improving it for bash/zsh I think it could get improved in general by looking at `os.environ` to have the `if` / branches inside of Python (`pdm --pep582`) instead of in the shell code (which is meant to be evaluated right away, and should get an exported `$PYTHONPATH` then).

pdm/cli/actions.py
linw1995: > Apart from improving it for bash/zsh I think it could get improved in general by looking at `os.environ` to have the `if` / branches inside of Python (`pdm --pep582`) instead of in the shell code (which is meant to be evaluated right away, and should get an exported `$PYTHONPATH` then).

Using the shell code instead of executing `pdm --pep582` takes less overhead of shell startup.

------

Using shell parameter expansion in bash and zsh will get the shell code like below.

`export PYTHONPATH=/.../pdm/pep582${PYTHONPATH:+:${PYTHONPATH}} `

I think the readability of this line of code is ok to me. But it should be simpler and more readable for users due to that the output of `pdm --pep582` is an example to show how to enable the PEP582 automatically. Users always can alter the result depending on what they want.

frostming: > I think it could get improved in general by looking at os.environ to have the if / branches inside of Python

The determine of which branch to take shouldn't happen when running `pdm --pep582`, as the output is often stored in rc files and other env var manipulation may take place before that.

PR #421, opened 2021-04-27
Feat add candidate output

pohlt:
## Pull Request Check List

I'm currently struggling with installing packages from multiple repos and had the impression that the log output could be improved.
The PR adds some more log output (log level: DETAIL) about matching candidates and their origin. If no matching candidates could be found, the non-matching candidates are printed to give the user some idea what might have gone wrong.

### DIFF
--- a/pdm/models/candidates.py
+++ b/pdm/models/candidates.py
@@ -215,7 +215,8 @@ class Candidate:
         return self.metadata
 
     def __repr__(self) -> str:
-        return f"<Candidate {self.name} {self.version}>"
+        source = getattr(self.link, "comes_from", "unknown")
+        return f"<Candidate {self.name} {self.version} from {source}>"
 
     @classmethod
     def from_installation_candidate(
--- a/pdm/models/repositories.py
+++ b/pdm/models/repositories.py
@@ -112,6 +112,9 @@ class BaseRepository:
             reverse=True,
         )
 
+        if not sorted_cans:
+            termui.logger.debug("\tCould not find any matching candidates.")
+
         if not sorted_cans and allow_prereleases is None:
             # No non-pre-releases is found, force pre-releases now
             sorted_cans = sorted(
@@ -124,6 +127,35 @@ class BaseRepository:
                 key=lambda c: c.version,
                 reverse=True,
             )
+
+            if not sorted_cans:
+                termui.logger.debug(
+                    "\tCould not find any matching candidates even when considering "
+                    "pre-releases.",
+                )
+
+        def print_candidates(
+            title: str, candidates: List[Candidate], max_lines: int = 10
+        ) -> None:
+            termui.logger.debug("\t" + title)
+            logged_lines = set()
+            for can in candidates:
+                new_line = f"\t  {can}"
+                if new_line not in logged_lines:
+                    logged_lines.add(new_line)
+                    if len(logged_lines) > max_lines:
+                        termui.logger.debug(
+                            f"\t  ... [{len(candidates)-max_lines} more candidate(s)]"
+                        )
+                        break
+                    else:
+                        termui.logger.debug(new_line)
+
+        if sorted_cans:
+            print_candidates("Found matching candidates:", sorted_cans)
+        elif cans:
+            print_candidates("Found but non-matching candidates:", cans)
+
         return sorted_cans
 
     def _get_dependencies_from_cache(self, candidate: Candidate) -> CandidateInfo:
--- a/news/421.misc.md
+++ b/news/421.misc.md
@@ -1,0 +1,1 @@
+Add log output about found candidates and their origin.


pdm/models/candidates.py
```diff
@@ -215,7 +215,8 @@ def get_metadata(
         return self.metadata
 
     def __repr__(self) -> str:
-        return f"<Candidate {self.name} {self.version}>"
+        source = getattr(self.link, "comes_from", "unknown")
```
frostming, line 218: What is the value if the candidate is from a file link(including VCS, remote file and local file)?

pdm/models/repositories.py
```diff
@@ -124,6 +131,30 @@ def find_candidates(
...
+            with self.environment.project.core.ui.indent("  "):
+                for can in candidates:
+                    self.environment.project.core.ui.echo(
+                        can, verbosity=termui.DETAIL, fg="blue"
+                    )
+
+        if sorted_cans:
+            print_candidates("Found matching candidates:", sorted_cans)
+        else:
+            print_candidates("Found but non-matching candidates:", cans)
```
frostming, line 156: This list may be super long for a popular package that has existed for years.

pdm/models/repositories.py
```diff
@@ -112,6 +112,13 @@ def find_candidates(
             reverse=True,
         )
 
+        if not sorted_cans:
+            self.environment.project.core.ui.echo(
```
frostming, line 116: Should these messages be written into the log file when verbose is off?

frostming:
The UI needs more careful design, considering this:

1. What message should be written into the log file when verbose is off.
2. How many lines will be output to the screen and whether it is easy for users to see what is happening.
3. Any color formatting to improve the readability?
4. Should the output go to stderr or stdout?

pdm/models/repositories.py
pohlt: How about limiting it to the top 10 (which might be arbitrary for `cans`)?

pdm/models/candidates.py
pohlt: Good question. I haven't tried and don't know. At least it doesn't break if `comes_from` is not set.

pdm/models/repositories.py
pohlt: Good idea. What would you prefer? I'm fine with both.

pohlt:
> The UI needs more careful design, consider this:
> 
> 1. What message should be written into the log file when verbose is off.
> 2. How many lines will be output to the screen and whether it is easy for users to see what is happening.
> 3. Any color formatting to improve the readability?
> 4. Should the output go to stderr or stdout?
> 
> And in fact, there is a redundant output level `DEBUG` that is never used in PDM. It can be activated with `-vv` but currently no message falls into this level.

If you give me your preferences on those topics, I'm happy to adapt the code. I also thought about using the `DEBUG` level. Might be the better choice.

pdm/models/candidates.py
frostming: Just to make sure it displays a meaningful value not confusing

pdm/models/repositories.py
frostming: `logger` is preferred for this case, it can be imported from `termui` or got with the exact same name.

pdm/models/repositories.py
```diff
@@ -124,6 +131,30 @@ def find_candidates(
                 key=lambda c: c.version,
                 reverse=True,
             )
+
+            if not sorted_cans:
+                self.environment.project.core.ui.echo(
+                    "Could not find any matching candidates even when considering "
```
frostming, line 137: logger as well

pdm/models/repositories.py
```diff
@@ -124,6 +131,30 @@ def find_candidates(
...
+                    "pre-releases",
+                    verbosity=termui.DETAIL,
+                    fg="yellow",
+                )
+
+        def print_candidates(title: str, candidates: List[Candidate]) -> None:
+            self.environment.project.core.ui.echo(
+                title, verbosity=termui.DETAIL, fg="blue"
+            )
+            with self.environment.project.core.ui.indent("  "):
```
frostming, line 147: I prefer this message to all go to log file. But the length should be limited, because a project could have ~50 candidates and the log content grows quickly.

frostming:
Here is a output with the change of this PR: 
[log.txt](https://github.com/pdm-project/pdm/files/6390469/log.txt)

As you can see there are lots of duplicate lines that are not informative. I recommend only log failing cases here in `find_matches` and log the **pinned** candidate that is chosen by the resolver. You can do this by improving the output of `pdm/resolver/reporters.py`. The current output is poorly designed and lacks color emphasis.

frostming:
You can add the unfiltered candidate list (length limited to top 10) when `find_matches` failed to find any.

pohlt:
> Here is a output with the change of this PR:
> [log.txt](https://github.com/pdm-project/pdm/files/6390469/log.txt)
> 
> As you can see there are lots of duplicate lines that are not informative. I recommend only log failing cases here in `find_matches` and log the **pinned** candidate that is chosen by the resolver. You can do this by improving the output of `pdm/resolver/reporters.py`. The current output is poorly designed and lacks color emphasis.

I don't understand the "lacks color emphasis" in context with logging the output to a file: We cannot use color there, right?
W.r.t. formatting I tried to be more compliant.

frostming:
> > Here is a output with the change of this PR:
> > [log.txt](https://github.com/pdm-project/pdm/files/6390469/log.txt)
> > As you can see there are lots of duplicate lines that are not informative. I recommend only log failing cases here in `find_matches` and log the **pinned** candidate that is chosen by the resolver. You can do this by improving the output of `pdm/resolver/reporters.py`. The current output is poorly designed and lacks color emphasis.
> 
> I don't understand the "lacks color emphasis" in context with logging the output to a file: We cannot use color there, right?
> W.r.t. formatting I tried to be more compliant.

You are right, I turned verbose on so the output goes to stdout, improving the format is also good

pohlt:
I'm going to filter out the duplicate lines. Please don't accept yet.

pdm/models/repositories.py
```diff
@@ -124,6 +127,35 @@ def find_candidates(
...
+                new_line = f"\t  {can}"
+                if new_line not in logged_lines:
+                    logged_lines.add(new_line)
+                    if len(logged_lines) > max_lines:
+                        termui.logger.debug(
+                            f"\t  ... [{len(candidates)-max_lines} more candidate(s)]"
+                        )
+                        break
+                    else:
+                        termui.logger.debug(new_line)
```
linw1995, line 152: Lack of considering the `candidates` parameter is an empty list. Please inform that in log file.

pohlt: Good catch. Fixed.
Only print list of `cans` if non-empty. If it is empty, the user gets a `pdm.exceptions.CandidateNotFound` anyway.

linw1995 (approved):
LGTM 💯

pdm/models/candidates.py
pohlt: For the cases you mentioned, pdm doesn't seem to go through the repository search. At least I wasn't able to get the candidate list output.

frostming (approved):

frostming:
Thanks for contribution

PR #422, opened 2021-04-28
Seek for other sitecustomize.py to import

linw1995:
## Pull Request Check List

Try to use nix to build an isolating environment for PDM development. But pdm can not co-exist with Python packages in nix due to the import mechanism that Python only imports module once by name.
The Python in nix using sitecustomize.py to import packages and PDM also using it to enable PEP582 globally. 

Hence, we need to **make sure that another sitecustomize.py gets imported if it exists**. So PDM can make PEP582 enabling globally, and  Python in nix also can import its packages.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -224,6 +224,13 @@ def pip_global_tempdir_manager():
         yield
 
 
+def remove_pep582_path_from_pythonpath(pythonpath):
+    """Remove all pep582 paths of PDM from PYTHONPATH"""
+    paths = pythonpath.split(os.pathsep)
+    paths = [path for path in paths if "pdm/pep582" not in path]
+    return os.pathsep.join(paths)
+
+
 @pytest.fixture()
 def project_no_init(tmp_path, mocker):
     p = main.create_project(tmp_path)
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -235,8 +242,11 @@ def project_no_init(tmp_path, mocker):
     do_use(p, getattr(sys, "_base_executable", sys.executable))
     with temp_environ():
         os.environ.pop("VIRTUAL_ENV", None)
-        os.environ.pop("PYTHONPATH", None)
         os.environ.pop("PEP582_PACKAGES", None)
+        pythonpath = os.environ.pop("PYTHONPATH", "")
+        pythonpath = remove_pep582_path_from_pythonpath(pythonpath)
```
frostming, line 247: We could make this simpler for testing, just clean the `PYTHONPATH` here. No need to keep the old value. Test case can provide its value when running

tests/test_integration.py
```diff
@@ -33,3 +37,44 @@ def test_basic_integration(python_version, project_no_init, strict_invoke):
     assert not any(
         line.strip().lower().startswith("django") for line in result.output.splitlines()
     )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("python_version", ["2.7", "3.6", "3.7", "3.8", "3.9"])
+def test_import_another_sitecustomize(python_version, project_no_init, strict_invoke):
```
frostming, line 44: A smaller case around `pdm run` should be enough, isn't it? You can find some examples in `test_run.py`

tests/conftest.py
linw1995: We could not assume the development environment of others not taking PYTHONPATH in usage. Some may use nix to do development. And nix is using PYTHONPATH to import packages that may essential to pdm testings. For example, the pip package of python3.9 for downloading wheel files for testing.

frostming: I didn't know the approach used by nix.

PR #427, opened 2021-05-01
add mypy pre-commit hook

danieleades:
add mypy precommit hook with incremental 'whitelisting'


danieleades:
in addition to the mypy config file, i've also added a mypy pre-commit hook, configured with a *whitelist*.

This is to support incremental adoption. As files are completely removed of type errors, they can be added to the whitelist, which will ensure that type errors are not reintroduced to those files

PR #429, opened 2021-05-01
improve typing in pdm.cli.utils

danieleades:
improve typing in `pdm.cli.utils`

i have a few questions about the types in this file, which i'll put inline

this is stacked on #427

### DIFF
--- a/pdm/cli/utils.py
+++ b/pdm/cli/utils.py
@@ -403,7 +403,7 @@ def find_importable_files(project: Project) -> Iterable[Tuple[str, Path]]:
         if not project_file.exists():
             continue
         for key, module in FORMATS.items():
-            if module.check_fingerprint(project, project_file.as_posix()):
+            if module.check_fingerprint(project, project_file):
                 yield key, project_file
 
 
--- a/pdm/cli/utils.py
+++ b/pdm/cli/utils.py
@@ -38,19 +38,21 @@ class PdmFormatter(argparse.HelpFormatter):
 
         # no help; start on same line and add a final newline
         if not action.help:
-            tup = self._current_indent, "", action_header
-            action_header = "%*s%s\n" % tup
+            action_header = "%*s%s\n" % (self._current_indent, "", action_header)
 
         # short action name; start on the same line and pad two spaces
         elif len(action_header) <= action_width:
-            tup = self._current_indent, "", action_width, action_header
-            action_header = "%*s%-*s  " % tup
+            action_header = "%*s%-*s  " % (
+                self._current_indent,
+                "",
+                action_width,
+                action_header,
+            )
             indent_first = 0
 
         # long action name; start on the next line
         else:
-            tup = self._current_indent, "", action_header
-            action_header = "%*s%s\n" % tup
+            action_header = "%*s%s\n" % (self._current_indent, "", action_header)
             indent_first = help_position
 
         # collect the pieces of the action help
--- a/pdm/cli/utils.py
+++ b/pdm/cli/utils.py
@@ -140,9 +140,7 @@ class Package:
         return f"<Package {self.name}=={self.version}>"
 
     def __eq__(self, value: object) -> bool:
-        if not isinstance(value, Package):
-            return False
-        return self.name == value.name
+        return isinstance(value, Package) and self.name == value.name
 
 
 def build_dependency_graph(working_set: WorkingSet) -> DirectedGraph:
--- a/pdm/cli/utils.py
+++ b/pdm/cli/utils.py
@@ -5,7 +5,7 @@ import os
 from argparse import Action
 from collections import ChainMap
 from pathlib import Path
-from typing import TYPE_CHECKING, Sequence, Set
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Set, Tuple
 
 import cfonts
 import tomlkit
@@ -23,8 +23,6 @@ from pdm.models.specifiers import get_specifier
 from pdm.project import Project
 
 if TYPE_CHECKING:
-    from typing import Dict, Iterable, List, Optional, Tuple
-
     from resolvelib.resolvers import RequirementInformation, ResolutionImpossible
 
     from pdm.models.candidates import Candidate
@@ -136,10 +134,12 @@ class Package:
     def __hash__(self) -> int:
         return hash(self.name)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<Package {self.name}=={self.version}>"
 
-    def __eq__(self, value: Package) -> bool:
+    def __eq__(self, value: object) -> bool:
+        if not isinstance(value, Package):
+            return False
         return self.name == value.name
 
 
@@ -152,11 +152,11 @@ def build_dependency_graph(working_set: WorkingSet) -> DirectedGraph:
     def add_package(key: str, dist: Distribution) -> Package:
         name, extras = strip_extras(key)
         extras = extras or ()
-        reqs = {}
+        reqs: Dict[str, Requirement] = {}
         if dist:
-            requirements = [
+            requirements = (
                 Requirement.from_pkg_requirement(r) for r in dist.requires(extras)
-            ]
+            )
             for req in requirements:
                 reqs[req.identify()] = req
             version = dist.version
@@ -252,7 +252,7 @@ def format_reverse_package(
     requires: str = "",
     prefix: str = "",
     visited: Optional[Set[str]] = None,
-):
+) -> str:
     """Format one package for output reverse dependency graph."""
     if visited is None:
         visited = set()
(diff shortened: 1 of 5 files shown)


danieleades:
there's something very fishy going on in `PdmFormatter._format_action`.

here's the mypy error-
```
action_header: str
Incompatible types in assignment (expression has type "Tuple[int, str, int, str]", variable has type "Tuple[int, str, str]")
```

it looks like either a 3-tuple or a 4-tuple are used? is this intentional?

danieleades:
error in this-


```python
class PdmParser(argparse.ArgumentParser):
    def format_help(self) -> str:
        formatter = self._get_formatter()

        if getattr(self, "is_root", False):
            banner = (
                cfonts.render(
                    "PDM",
                    font="slick",
                    gradient=["bright_red", "bright_green"],
                    space=False,
                )
                + "\n"
            )
            formatter._add_item(lambda x: x, [banner])
            self._positionals.title = "Commands"
        self._optionals.title = "Options"
        # description
        formatter.add_text(self.description)

        # usage
        formatter.add_usage(
            self.usage,
            self._actions,
            self._mutually_exclusive_groups,
            prefix=termui.yellow("Usage", bold=True) + ": ",
        )
...
```

`self.usage` might be `None`, but `formatter.add_usage` requires it to be not `None`

danieleades:
type error here-

```python
def build_dependency_graph(working_set: WorkingSet) -> DirectedGraph:
    """Build a dependency graph from locked result."""
    graph = DirectedGraph()
    graph.add(None)  # sentinel parent of top nodes.
    node_with_extras = set()

    def add_package(key: str, dist: Distribution) -> Package:
        name, extras = strip_extras(key)
        extras = extras or ()
        reqs: Dict[str, Requirement] = {}
        if dist:
            requirements = (
                Requirement.from_pkg_requirement(r) for r in dist.requires(extras)
            )
            for req in requirements:
                reqs[req.identify()] = req
            version = dist.version
        else:
            version = None
```

this is the problematic line-
`reqs[req.identify()] = req`

`Requirement.identify` returns `Optional[str]`, but you need a `str` to index the dictionary

linw1995:
> pdm/cli/utils.py:46: error: Incompatible types in assignment (expression has type "Tuple[int, str, int, str]", variable has type "Tuple[int, str, str]")

The variable of `tup` is not used outside the block of control flow. Maybe we can change them into format like this `action_header = "%*s%s\n" % (self._current_indent, "", action_header)`.

> pdm/cli/utils.py:103: error: Argument 1 to "add_usage" of "HelpFormatter" has incompatible type "Optional[str]"; expected "str"

The internal codes of `HelpFormatter` actually accepts `usage=None`. There may be some mistakes in the type stub of `argparser`. This need further investigation.

> pdm/cli/utils.py:163: error: Invalid index type "Optional[str]" for "Dict[str, Requirement]"; expected type "str"

Original `Requirement` class comes from `pkg_resources` in `pip/_vendor`. It uses regex to parse requirement string and the regex format ensures the `req.name` is not None.

In `pdm.models.Requirement.from_pkg_requirement`, it seems the `req.project_name` is not passed by default and `req.name` is not None for sure. So the `key` of pdm's `Requirement` is not None and `identify()` result must be a string.

@frostming maybe we can change the function signature of `identify()` ?

danieleades:
> > pdm/cli/utils.py:103: error: Argument 1 to "add_usage" of "HelpFormatter" has incompatible type "Optional[str]"; expected "str"
> 
> The internal codes of `HelpFormatter` actually accepts `usage=None`. There may be some mistakes in the type stub of `argparser`. This need further investigation.

the typings in typeshed definitely specify `Text` rather than `Optional[Text]`. perhaps that's a mistake?

danieleades:
there's also quite a lot of type errors as a result of the attributes of `pdm.models.requirements.Requirement` being set at runtime. It would be better to set the attributes of that class statically. That should probably be in it's own pull request (and should probably be done by someone with more intimate knowledge of how this class works, though I'd be happy to review!)

pdm/cli/utils.py
```diff
@@ -40,19 +38,21 @@ def _format_action(self, action: Action) -> str:
...
         # short action name; start on the same line and pad two spaces
         elif len(action_header) <= action_width:
-            tup = self._current_indent, "", action_width, action_header
-            action_header = "%*s%-*s  " % tup
+            action_header = "%*s%-*s  " % (
+                self._current_indent,
+                "",
+                action_width,
+                action_header,
+            )
```
danieleades, line 50: not entirely sure why mypy had a problem with re-aliasing the 'tup' variable, given it's scoped to the `if` blocks...

in any case, i'd appreciate it if someone could walk me through what these c-string style formatting strings are doing. How would people feel about updating these to f-strings?

pdm/models/requirements.py
```diff
@@ -118,9 +118,7 @@ def marker(self, value: Optional[Marker]) -> None:
         except InvalidMarker as e:
             raise RequirementError("Invalid marker: %s" % str(e)) from None
 
-    def identify(self) -> Optional[str]:
-        if not self.key:
-            return None
+    def identify(self) -> str:
```
danieleades, line 121: @linw1995

changing this signature results in a failing test

```
  File "/home/runner/work/pdm/pdm/pdm/termui.py", line 192, in logging
    yield logger
  File "/home/runner/work/pdm/pdm/tests/resolver/test_resolve.py", line 46, in resolve_requirements
    mapping, *_ = resolve(resolver, requirements, requires_python)
  File "/home/runner/work/pdm/pdm/pdm/resolver/core.py", line 150, in resolve
    result = resolver.resolve(requirements, max_rounds)
  File "/home/runner/work/pdm/pdm/__pypackages__/3.7/lib/resolvelib/resolvers.py", line 473, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
  File "/home/runner/work/pdm/pdm/__pypackages__/3.7/lib/resolvelib/resolvers.py", line 341, in resolve
    name, crit = self._merge_into_criterion(r, parent=None)
  File "/home/runner/work/pdm/pdm/pdm/resolver/core.py", line 24, in _merge_into_criterion
    identifier, crit = _old_merge_into_criterion(self, requirement, parent)
  File "/home/runner/work/pdm/pdm/__pypackages__/3.7/lib/resolvelib/resolvers.py", line 140, in _merge_into_criterion
    identifier = self._p.identify(requirement_or_candidate=requirement)
  File "/home/runner/work/pdm/pdm/pdm/resolver/providers.py", line 30, in identify
    return requirement_or_candidate.identify()
  File "/home/runner/work/pdm/pdm/pdm/models/requirements.py", line 123, in identify
    return self.key + extras
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

linw1995: https://github.com/danieleades/pdm/blob/refactor/pdm.cli.utils/pdm/models/requirements.py#L309-L312

There has a missing procedure to read the name and version from the source distribution. So the name of the requirement is not set.

danieleades: What's the way forward here?

frostming: @linw1995 We can expand the logic to extract the name and version from egg info too: `<name>-<version>.tar.gz`

frostming: `None` is a possible value for `identify()`, when the package is a local directory and its name can't be known until it is built.

danieleades: > `None` is a possible value for `identify()`, when the package is a local directory and its name can't be known until it is built.

if `identify` can return `None`, then i don't think it should be used as a unique key. is there an alternative unique identify that can be used? Would you not get a collision if you have multiple keys which are `None`?

frostming: > if identify can return None, then i don't think it should be used as a unique key. is there an alternative unique identify that can be used? Would you not get a collision if you have multiple keys which are None?

Good point, we didn't consider that case, a generated random name should be better and we can replace it later.

linw1995: > > if identify can return None, then i don't think it should be used as a unique key. is there an alternative unique identify that can be used? Would you not get a collision if you have multiple keys which are None?
> 
> Good point, we didn't consider that case, a generated random name should be better and we can replace it later.

Maybe use the path of the local directory which is relative to project root as the result of `identify()`.

danieleades: i've updated this to use the path if the key is not set. this is failing one test-

```
_ test_resolve_local_artifacts[/home/runner/work/pdm/pdm/tests/fixtures/artifacts/demo-0.0.1.tar.gz] _
[gw0] linux -- Python 3.7.10 /opt/hostedtoolcache/Python/3.7.10/x64/bin/python

project = <Project '/tmp/pytest-of-runner/pytest-0/popen-gw0/test_resolve_local_artifacts__0'>
repository = <tests.conftest.TestRepository object at 0x7f1d3f6d3d50>
requirement_line = '/home/runner/work/pdm/pdm/tests/fixtures/artifacts/demo-0.0.1.tar.gz'

    @pytest.mark.parametrize(
        "requirement_line",
        [
            f"{(FIXTURES / 'artifacts/demo-0.0.1.tar.gz').as_posix()}",
            f"{(FIXTURES / 'artifacts/demo-0.0.1-py2.py3-none-any.whl').as_posix()}",
        ],
    )
    def test_resolve_local_artifacts(project, repository, requirement_line):
>       result = resolve_requirements(repository, [requirement_line])

tests/resolver/test_resolve.py:109: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/resolver/test_resolve.py:46: in resolve_requirements
    mapping, *_ = resolve(resolver, requirements, requires_python)
pdm/resolver/core.py:153: in resolve
    all_metasets = extract_metadata(result)
pdm/resolver/core.py:107: in extract_metadata
    new_metasets[k] = _build_metaset(crit, all_metasets, keep_unresolved)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

criterion = Criterion((<NamedRequirement chardet; os_name == "nt">, via=<Candidate demo 0.0.1 from None>))
all_metasets = {}, keep_unresolved = set()

    def _build_metaset(
        criterion: Criterion,
        all_metasets: Dict[str, Metaset],
        keep_unresolved: Set[Optional[str]],
    ) -> Metaset:
    
        metaset = None
    
        for r, parent in criterion.information:
            if parent and _identify_parent(parent) in keep_unresolved:
                continue
            this_metaset = Metaset(r.marker)
            # Use 'and' to connect markers inherited from parent.
            if not parent:
                parent_metaset = Metaset()
            else:
>               parent_metaset = all_metasets[_identify_parent(parent)]
E               KeyError: 'demo'

pdm/resolver/core.py:56: KeyError
```

danieleades:
ideally i'd like to remove *all* of the errors in this file so it can be added to the whitelist in #427

frostming:
Close in favor of #471

PR #430, opened 2021-05-01
add mypy pre-commit hook

danieleades:
this pull request adds a mypy pre-commit hook.

This is kind of the nuclear option, as this will throw a *lot* of errors for any file that gets touched in a commit. That could be very noisy.

It would also be a strong incentive to refactor and fix type errors ;) ...

I'll just leave this here @frostming and you can decide what you want to do with this. Maybe the best thing is to keep it on the back burner until more of the errors are squashed?

this is stacked on #427, so it can be rebased if that is merged

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -5,4 +5,3 @@ disallow_untyped_defs = true
 
 [mypy-pdm._vendor.*]
 ignore_errors = True
-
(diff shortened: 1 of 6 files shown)


danieleades:
for this to work, the pre-commit CI would have to be configured to run mypy on only changed files, or not at all

danieleades:
closed in favour of #427

PR #437, opened 2021-05-03
apply and enforce 'pyupgrade'

danieleades:
add 'pyupgrade' pre-commit hook.

**also deprecates python 2.7 support**


frostming:
> also deprecates python 2.7 support

We don't plan to do this. We apply a similar support strategy to `virtualenv`.

PR #438, opened 2021-05-03
add bugbear linting

danieleades:
not sure if this is of interest. Bugbear adds a few slightly opinionated lints

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,10 +4,13 @@ repos:
     hooks:
       - id: black
         exclude: ^pdm/_vendor
+
   - repo: https://github.com/PyCQA/flake8
     rev: 3.9.1
     hooks:
       - id: flake8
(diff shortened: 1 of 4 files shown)


frostming (commented):
We are not going to add more hooks, but OK with applying the changes.

danieleades:
> We are not going to add more hooks

Are you sure? this plugin only suggested 3 changes, and at least one of these was a bug. It's not exactly 'noisy'.

I'm more than happy to strip the `.pre-commit-config.yaml` change from this pull request if that's what you want. Just hoping to understand *why* you're against adding new hooks (given i've added them in other pull requests too, such as #427).

That said, this isn't actually a new hook, just a flake8 plugin to add more lints.

frostming:
> given i've added them in other pull requests too, such as #427

Mypy is an exception. This one(bugbear plugin) is just OK and I am against adding pyupgrade.

PR #451, opened 2021-05-08
Fix wrong mypy configuration

linw1995:
## Pull Request Check List

Fix the wrong Mypy configuration. It would be better that change to execute `mypy pdm` instead of executing `mypy` with `files=pdm/**/*.py` because `files=...` ignores `exclude=...`.

`follow_imports = silent` is for avoiding that it reports errors which came from `pdm/_vendor`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,21 +22,22 @@ repos:
     hooks:
       - id: mypy
         args:
-          - --follow-imports=skip
-        files: | # incremental 'fully-typed' list
-          (?x)^(
-              pdm\/__init__\.py|
-              pdm\/__main__\.py|
-              pdm\/exceptions\.py|
-              pdm\/formats\/.*\.py|
-              pdm\/installers\/.*\.py|
-              pdm\/models\/auth\.py|
-              pdm\/models\/caches\.py|
-              pdm\/models\/python\.py|
-              pdm\/models\/setup\.py|
-              pdm\/models\/specifiers\.py|
-              pdm\/models\/versions\.py|
-              pdm\/project\/config\.py|
-              pdm\/termui\.py|
-              pdm\/utils\.py|
-          )$
+          - pdm
+          - |
+            --exclude=(?x)^(
+              (?!pdm\/__init__\.py)
+              (?!pdm\/__main__\.py)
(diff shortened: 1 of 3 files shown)


linw1995:
I changed the configuration of mypy in `.pre-commit-config.yaml` to ensure that the behavior of executing mypy by pre-commit is the same as mypy pdm except that the first one has a different `--exclude=....`.

.pre-commit-config.yaml
```diff
@@ -22,21 +22,22 @@ repos:
...
+              (?!pdm\/models\/caches\.py)
+              (?!pdm\/models\/python\.py)
+              (?!pdm\/models\/setup\.py)
+              (?!pdm\/models\/specifiers\.py)
+              (?!pdm\/models\/versions\.py)
+              (?!pdm\/project\/config\.py)
+              (?!pdm\/termui\.py)
+              (?!pdm\/utils\.py)
+            ).*\.py$ # incremental 'fully-typed' list
+        pass_filenames: false
```
danieleades, line 43: what's the benefit of inverting the logic in this way?

setup.cfg
```diff
@@ -41,10 +41,10 @@ exclude_lines =
 ignore_errors = true
 
 [mypy]
+follow_imports = silent
```
danieleades, line 44: i think the idiomatic way to suppress errors in `_vendor/` would be with this-

```ini
[mypy-pdm._vendor.*]
ignore_errors = True
```

.pre-commit-config.yaml
linw1995: https://github.com/pdm-project/pdm/pull/452/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9 the latest change is here.

We want to executing `mypy` via pre-commit is same as executing `mypy` via cli directly (except the first one with different `--exclude=...`. In pre-commit we need to exclude no-fully typed files because we want to run mypy on fully typed files only. And using `--exclude=...` is easier rather than configuring them by using `[mypy-*]` in the setup.cfg file to exclude all the no-fully typed files.

> The document of `--exclude` optional argument : https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-exclude

Issue #443, opened 2021-05-05
Non-editable installation of the project itself


pawamoy:
## Is your feature request related to a problem? Please describe.

I'm currently experimenting with PDM in Docker multi-stage builds.

- In the "builder" phase, I install PDM, copy the necessary files, and run `pdm install --prod --no-lock`.
- In the "run" phase, I simply copy the `__pypackages__` lib directory from the builder phase, and add it to the PYTHONPATH.

```dockerfile
# build stage
FROM baseimage AS builder

# install PDM
RUN pip install -U pip setuptools wheel
RUN pip install pdm

# copy files
RUN mkdir -p /api/src
COPY pyproject.toml pdm.lock README.md /api/
COPY src/namespace /api/src/namespace

# install deps and project
WORKDIR /api
RUN pdm install --prod --no-lock



# run stage
FROM runimage

# retrieve packages from build stage
RUN mkdir /api
COPY --from=builder /api/__pypackages__/3.8/lib /api/pkgs
ENV PYTHONPATH=/api/pkgs

# set command/entrypoint
COPY run.py /api/run.py
CMD ["python3", "/api/run.py"]
```

The problem is that PDM installed the current project in editable mode, meaning all I have in my final `/api/pkgs` is the dependencies and the egg-link of my project. Upon running `run.py`, it fails with `No module named 'namespace'`.

## Describe the solution you'd like

Since I'm installing the dependencies and project for *production* (emphasized by the `--prod` flag), I would like my current project to be installed in non-editable mode in the `__pypackages__` lib directory. If you want this to stay flexible, I would gladly have a second `--no-editable` flag to explicitly ask for this behavior.

## Alternatives you've considered

I can run `pip install . --no-deps --use-feature=in-tree-build -t __pypackages__/3.8/lib` in the build stage, after `pdm install --prod --no-lock`.

I could also build a wheel in the build stage, and copy it in the run stage to install it with `pip install my-wheel.whl -t /api/pkgs --no-deps`. But it would require to have `pip` available in the run stage, which might also be problematic.

In the two cases above, I feel like I would need an additional flag in build stage to tell PDM *not* to install the project at all (only the prod deps), something like Poetry's `--no-root`. Otherwise there's an extraneous egg-link in my packages that I should probably remove.

I also considered using a `venv` as an artifact instead of the `__pypackages__` lib directory, but why keep using venv when we can get rid of it 🙂 

## Final note

[...]

PR #459, opened 2021-05-13
docs: Add multi-stage Dockerfile example

pawamoy:
## Pull Request Check List

- [ ] ~~Test cases added for changed code.~~

Added a Dockerfile example. Related to #443


frostming:
Thanks for that!

PR #479, opened 2021-05-25
Update setup_dev.py

dbeattie71:
Fix potential typo.

## Pull Request Check List


setup_dev.py
```diff
@@ -1,6 +1,6 @@
 """
 This script is used to setup development environment of pdm. It follows the naming
-convention of setuptools but it is not basically a setuptools setup script.
+convention of setuptools but is basically a setuptools setup script.
```
frostming, line 3:
```suggestion
convention of setuptools but isn't a real setuptools setup script.
```

Issue #487, opened 2021-05-31
"pdm add" generates invalid toml


blueyed:
`pdm add sentry-sdk` generated this diff for `pyproject.toml`:
```diff
 dependencies = [
     "django~=3.1",
-]
+, "sentry-sdk~=1.1"]
 requires-python = ">=3.9"
```

This then fails at the end, or with `pdm lock -v`:
```
…
🔒 Lock successful
Changes are written to pdm.lock.
Changes are written to pyproject.toml.
Traceback (most recent call last):
  File "…/project/.venv/bin/pdm", line 8, in <module>
    sys.exit(main())
  File "…/project/.venv/lib/python3.9/site-packages/pdm/core.py", line 184, in main
    return Core().main(args)
  File "…/project/.venv/lib/python3.9/site-packages/pdm/core.py", line 145, in main
    raise cast(Exception, err).with_traceback(traceback)
  File "…/project/.venv/lib/python3.9/site-packages/pdm/core.py", line 140, in main
    f(options.project, options)
  File "…/project/.venv/lib/python3.9/site-packages/pdm/cli/commands/add.py", line 44, in handle
    actions.do_add(
  File "…/project/.venv/lib/python3.9/site-packages/pdm/cli/actions.py", line 208, in do_add
    project.write_lockfile(lockfile, False)
  File "…/project/.venv/lib/python3.9/site-packages/pdm/project/core.py", line 395, in write_lockfile
    toml_data["metadata"].update(self.get_lock_metadata())
  File "…/project/.venv/lib/python3.9/site-packages/pdm/project/core.py", line 388, in get_lock_metadata
    content_hash = atoml.string("sha256:" + self.get_content_hash("sha256"))
  File "…/project/.venv/lib/python3.9/site-packages/pdm/project/core.py", line 417, in get_content_hash
    "sources": self.tool_settings.get("source", []),
  File "…/project/.venv/lib/python3.9/site-packages/pdm/project/core.py", line 103, in tool_settings
    data = self.pyproject
  File "…/project/.venv/lib/python3.9/site-packages/pdm/project/core.py", line 93, in pyproject
    data = atoml.parse(self.pyproject_file.read_text("utf-8"))
  File "…/project/.venv/lib/python3.9/site-packages/atoml/api.py", line 61, in parse
    return Parser(string).parse()
  File "…/project/.venv/lib/python3.9/site-packages/atoml/parser.py", line 152, in parse
    key, value = self._parse_table()
[...]

PR #491, opened 2021-06-01
docs: fix typo in news/487.bugfix.md

blueyed:


news/487.bugfix.md
```diff
@@ -1 +1 @@
-Fix the format of dependency arrays when a new valud is appended.
+Fix the format of dependency arrays when a new value is appended.
```
blueyed, line 1: Given that https://github.com/pdm-project/pdm/issues/487 was an upstream issue with atoml, it should be removed/get rephrased maybe?

(Note that the dep was not bumped to enforce the fix)

frostming: This commit alone also fixes the issue

blueyed: Ok, let's fix the typo then only?  (i.e. merge this)

PR #535, opened 2021-07-10
Update `run` to support `pyenv` shims on macOS

### LINKED ISSUE #494: Error executing Jupyter command 'lab': [Errno 8] Exec format error

```
$ pdm run -v jupyter lab

  File "/usr/local/bin/pdm", line 8, in <module>
    sys.exit(main())
  File "/usr/local/Cellar/pdm/1.6.2/libexec/lib/python3.9/site-packages/pdm/core.py", line 184, in main
    return Core().main(args)
  File "/usr/local/Cellar/pdm/1.6.2/libexec/lib/python3.9/site-packages/pdm/core.py", line 145, in main
  File "/usr/local/Cellar/pdm/1.6.2/libexec/lib/python3.9/site-packages/pdm/core.py", line 140, in main
    f(options.project, options)
  File "/usr/local/Cellar/pdm/1.6.2/libexec/lib/python3.9/site-packages/pdm/cli/commands/run.py", line 199, in handle
    self._run_command(
  File "/usr/local/Cellar/pdm/1.6.2/libexec/lib/python3.9/site-packages/pdm/cli/commands/run.py", line 99, in _run_command
    os.execv(expanded_command, expanded_args)
OSError: [Errno 8] Exec format error
```

Note that I get the same error when running:

```bash
$ ./__pypackages__/3.9/bin/jupyter lab
Error executing Jupyter command 'lab': [Errno 8] Exec format error
```

## Steps to reproduce

```bash
pdm init
pdm add --dev jupyterlab
```

## Actual behavior

Unable to launch Jupyter Lab

## Expected behavior

Expected to be able to launch Jupyter Lab

## Environment Information

```bash

pdm info && pdm info --env
PDM version:        1.6.2
Python Interpreter: /Users/steve/.pyenv/shims/python (3.9)
Project Root:       /Users/steve/code/storeleads-data
Project Packages:   /Users/steve/code/storeleads-data/__pypackages__/3.9
{
  "implementation_version": "3.9.4",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "20.4.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 20.4.0: Thu Apr 22 21:46:47 PDT 2021; root:xnu-7195.101.2~1/RELEASE_X86_64",
  "python_full_version": "3.9.4",
  "python_version": "3.9",
  "sys_platform": "darwin"
}
```

jeffreyc:
## Pull Request Check List

This is an attempt to fix #494, where PDM fails with "OSError: [Errno 8] Exec format error" when using pyenv shims on macOS. This error appears to be specific to macOS; I was unable to recreate it in a Linux Docker image.

Per https://stackoverflow.com/a/49131318, I believe this error happens when the shebang points to a text file instead of a binary. Manual testing confirms this error happens when the shebang points to `~/.pyenv/shims/python` and not when the shebang points to `~/.pyenv/versions/3.9.6/bin/python3.9` (the executable instead of the shim) or when invoked by passing the Python executable/shim as the first argument to `execv`.

Passing `project.python_executable` seems more straightforward than rewriting the shebang, since it's non-trivial to detect the shim's target, but feel free to suggest another approach.

I spent a little time looking at testing this only to realize it's not really testable, since the modified code is in a code path that's intentionally not executed when running tests. I did try modifying https://github.com/pdm-project/pdm/blob/main/pdm/cli/commands/run.py#L94 but it breaks a bunch of existing tests without adding clear value because it's not testing the modified code. Besides, this code is exercised when running `pdm run test`, so if it doesn't work, the test suite will fail.

Fixes #494.

### DIFF
--- a/news/535.bugfix.md
+++ b/news/535.bugfix.md
@@ -1,0 +1,1 @@
+Update `run` to support `pyenv` shims on macOS.
--- a/pdm/cli/commands/run.py
+++ b/pdm/cli/commands/run.py
@@ -95,7 +95,9 @@ class Command(BaseCommand):
         else:
             if chdir:
                 os.chdir(project.root)
-            os.execv(expanded_command, expanded_args)
+            os.execv(
+                project.python_executable, [project.python_executable] + expanded_args
+            )
 
     def _normalize_script(
         self, script: Any


frostming (changes requested):
Must test if the command is a python script.

jeffreyc:
> Must test if the command is a python script.

Is there a risk `python_executable` won't be a valid Python interpreter? The code around setting that appears pretty robust. I added a check, but it's not clear the `else` branch is reachable. If I misunderstood and there's another concern, please let me know.

frostming:
You might get me wrong, i mean the first arg following python_executable must be a valid Python script. Just think of `pdm run ls -l`. One possible way is to look at the shebang line or if there's any better approach.

jeffreyc:
Apologies, I misunderstood and didn't realize `pdm run ls -l` was a valid use case. I'll close this for now,

frostming:
It's OK, is it valid for pyenv shims Python as the shebang line? If not, we should avoid this in the first place.

jeffreyc:
> is it valid for pyenv shims Python as the shebang line? If not, we should avoid this in the first place.

It depends. I think using a script as the interpreter is only supported on Linux/Minix, _but_ it also appears to be supported by zsh on macOS. So a script in the local package directory using shims should work when called directly if your shell is zsh (the current default shell on macOS), but not when invoked via `os.execv` et al, which I believe use `/bin/sh`

PR #583, opened 2021-08-05
Feature/#531/more powerful list command

ElijahSink:
## Pull Request Check List

Added a new mode `--json` to the `list` command which outputs the dependency graph as a JSON document.

### DIFF
--- a/pdm/cli/utils.py
+++ b/pdm/cli/utils.py
@@ -181,7 +181,7 @@ def format_package(
     package: Package,
     required: str = "",
     prefix: str = "",
-    visited: set[str] | None = None,
+    visited: frozenset[str] = frozenset(),
 ) -> str:
     """Format one package.
 
@@ -191,8 +191,6 @@ def format_package(
     :param prefix: prefix text for children
     :param visited: the visited package collection
     """
-    if visited is None:
-        visited = set()
     result = []
     version = (
         termui.red("[ not installed ]")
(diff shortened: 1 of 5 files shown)


pdm/cli/utils.py
```diff
@@ -291,23 +301,107 @@ def _format_forward_dependency_graph(project: Project, graph: DirectedGraph) ->
...
+        top_level_packages = filter(
+            lambda n: not list(graph.iter_children(n)), graph
+        )  # leaf nodes
+        build_dependency_json_subtree: Callable = build_reverse_dependency_json_subtree
+    else:
+        top_level_packages = graph.iter_children(None)  # root nodes
+        build_dependency_json_subtree = build_forward_dependency_json_subtree
+    return [
+        build_dependency_json_subtree(p, project, graph)
+        for p in iter(sorted(top_level_packages, key=lambda p: p.name if p else ""))
```
frostming, line 388: `iter()` is redundant?

frostming (commented):
Be careful about circular dependencies, which may cause an infinite loop. You may also need to include such in the test case

pdm/cli/utils.py
ElijahSink: Yep, thanks! Removed.

ElijahSink:
The existing test case for circular dependencies results in no output for both the normal-reversed and json-reversed list actions. This is because the simple test case (direct circular dependency between `foo` and `foo-bar`) has no leaf nodes, which are our starting points for listing in reverse.

This being the case, I've modified the test cases for reverse listing a circular graph. Perhaps there's a better way to find "pseudo-leaf nodes" in a purely-circular dependency graph? Perhaps it doesn't matter.

frostming:
Good enough, let's see how to improve the rest in another go. Thank you for taking care of this!

PR #584, opened 2021-08-05
Fix tomli deprecation warning

hukkin:
Text file objects are deprecated in Tomli. This PR moves to using binary file objects to avoid deprecation warning

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -500,7 +500,7 @@ summary = "Python Library for Tom's Obvious, Minimal Language"
 
 [[package]]
 name = "tomli"
-version = "1.0.4"
+version = "1.2.0"
 requires_python = ">=3.6"
 summary = "A lil' TOML parser"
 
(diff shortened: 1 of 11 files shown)


frostming:
Hi, I would like to know why text file object is not allowed and how binary file solves the problem:

> Binary file objects should be used instead to avoid opening a TOML file with universal newlines or with an encoding other than UTF-8.

As you simply do

https://github.com/hukkin/tomli/blob/3eb26137faa786cdb5f187a0df52aaca903c23f1/tomli/_parser.py#L69

IIRC `.decode()` just chooses the default file encoding rather than detecting what is used to encode the file.

hukkin:
> IIRC .decode() just chooses the default file encoding rather than detecting what is used to encode the file

What you describe is how `open("file.txt")` works.

`bytes.decode()` [defaults to UTF-8](https://docs.python.org/3/library/stdtypes.html#bytes.decode) and that is platform indepependent.

frostming:
Fine, convinced.

PR #589, opened 2021-08-07
Feature: Opt in to cache wheel installation

frostming:
## Pull Request Check List

A new config item `feature.install_cache` is added. When it is turned on, PDM will install wheels into a *centralized package repo* and create `.pth` files under the project packages directory. 

FWIW, `.pth` works like a *symlink* whose content will be appended to `sys.path` for searching modules. `.pth` files are also supported by most auto-complete engines for module discovery.

In this way, we can save a lot of disk usage especially when the data science stack is used in many projects on the same platform -- We only keep one copy of a particular package and refer to that package in different projects.


pawamoy:
By one copy of each package, you mean one copy of each version of each library? That's great, it will save up a _lot_ of disk space, even without heavy data science libs :)

frostming:
yes, in fact the cache key is the wheel's filename. simple but works.

PR #613, opened 2021-08-20
Use test.pypi.org instead of testpypi.org

frafra:
## Pull Request Check List

`testpypi.org` URL seems wrong; `test.pypi.org` should be used instead.


frostming:
it's just a placeholder telling "I'm another index url". but i don't mind merging. thanks.

frafra:
> it's just a placeholder telling "I'm another index url". but i don't mind merging. thanks.

Sure. I did this PR because I was a bit confused by the documentation, mostly.

Thanks again for your quick replies.

PR #629, opened 2021-08-24
Split arguments to interpreter executable.

CharString:
Jython 2.7 doesn't handle the concatenated form.

## Pull Request Check List

This fixes the issue. But my clean checkout of pdm fails 3 tests:
- [ ] test_integration.py test_basic_integration[3.6] 
      ModuleNotFoundError: No module named 'pdm'
- [ ] test_integration.py test_basic_integration[3.7]
      ModuleNotFoundError: No module named 'pdm'
- [ ] cli/test_cli.py test_show_self_package (but only when all tests are run [!?]; when run in isolation, it passes)

      E       AssertionError: assert 'test_project\n0.0.0\n' == '    test_pro...\n    0.0.0\n'
      E         -     test_project
      E         ? ----
      E         + test_project
      E         -     0.0.0
      E         ? ----
      E         + 0.0.0

Looks like the first two tests find python interpreters that I have installed on my machine, but that don't have pdm installed. I might have to run pdm install with both of them. Why the third test fails when I run the whole suite, but not when I run the single test, I don't know.

### DIFF
--- a/news/628.bugfix.md
+++ b/news/628.bugfix.md
@@ -1,0 +1,1 @@
+Fix crash when Jython 2.7 is found.


frostming:
I can't see what this PR is trying to solve. Although Jython is not declared to be supported nor covered by CI, if you could make it work without touching many codes I won't refuse to accept. I am going to close this now, you can submit a new PR when you are ready with the change. Thanks.

CharString:
@frostming The problem is not that I want to use Jython. The problem is that it is installed on my system, `pdm init` finds it and invokes it in a way that it doesn't support, and then pdm crashes...

Writing this I think the real solution is finding the code in pdm that finds python interpreters and make it more resilient to things that look like a python, slither like a python, but don't quack like a python...

PR #665, opened 2021-10-02
Improve documentation

laike9m:


pawamoy:
I don't think users are supposed to modify `.pdm.toml` directly, so I'm not sure this change has value.
Instead of breaking the previous anchor by changing the title, maybe you could add a paragraph explaining how to get and set these configuration items, and that they are written in the `.pdm.toml` file?

pawamoy:
Actually, mayve a simple relative link to https://pdm.fming.dev/usage/project/#configure-the-project would suffice.

frostming:
Agree with @pawamoy , you can add a line below saying "The configurations can be retrieved and modified by `pdm config` command`"

Issue #653, opened 2021-09-13
Could not install ortools


lgblkb:
- [ ] I have searched the issue tracker and believe that this is not a duplicate.

**Make sure you run commands with `-v` flag before pasting the output.**

## Steps to reproduce

pdm add ortools -v

## Actual behavior

```
Adding packages to default dependencies: ortools
======== Start resolving requirements ========
  ortools
  Adding requirement ortools
	Found matching candidates:
	  <Candidate ortools 9.0.9048 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 8.2.8710 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 8.1.8487 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 8.0.8283 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 7.8.7959 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 7.7.7810 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 7.6.7691 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 7.5.7466 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 7.4.7247 from https://pypi.org/simple/ortools/>
	  <Candidate ortools 7.3.7083 from https://pypi.org/simple/ortools/>
	  ... [268 more candidate(s)]
======== Starting round 0 ========
  Adding requirement protobuf>=3.15.8(from ortools 9.0.9048)
	Found matching candidates:
	  <Candidate protobuf 3.17.3 from https://pypi.org/simple/protobuf/>
	  <Candidate protobuf 3.17.2 from https://pypi.org/simple/protobuf/>
	  <Candidate protobuf 3.17.1 from https://pypi.org/simple/protobuf/>
	  <Candidate protobuf 3.17.0 from https://pypi.org/simple/protobuf/>
	  <Candidate protobuf 3.16.0 from https://pypi.org/simple/protobuf/>
	  <Candidate protobuf 3.15.8 from https://pypi.org/simple/protobuf/>
  Adding requirement absl-py>=0.11(from ortools 9.0.9048)
	Found matching candidates:
	  <Candidate absl-py 0.13.0 from https://pypi.org/simple/absl-py/>
	  <Candidate absl-py 0.12.0 from https://pypi.org/simple/absl-py/>
	  <Candidate absl-py 0.11.0 from https://pypi.org/simple/absl-py/>
[...]

PR #679, opened 2021-10-12
Update installer version to fix some wheel packages failing to install

SmileyChris:
## Pull Request Check List

- [ ] ~Test cases added for changed code.~

Installer was just released (pradyunsg/installer#78) which fixes pdm breaking on installing some packages with "interesting" wheels. One notable example is the most recent version of the `cryptography` package.

(btw: getting to use the development version of pdm to update its own requirements is fantastically meta)

### DIFF
--- a/news/653.dep.md
+++ b/news/653.dep.md
@@ -1,0 +1,1 @@
+Update `installer` to `0.3.0`, fixing a bug that broke installation of some packages with unusual wheel files.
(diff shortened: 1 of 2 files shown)


frostming:
Thanks, can you update the issue number in the news fragment to refer to #653 ?

SmileyChris:
Done, made it a `.dep` news fragment too, as that's probably more accurate.

frostming:
The failure has nothing to do with this PR. Merging it now. Thanks @SmileyChris

PR #680, opened 2021-10-13
Strip trailing whitespace with pdm-show

blueyed:
## Pull Request Check List

pdm-show uses `project.core.ui.display`, which adds trailing spaces to all the
lines (based on the longest one), which results in them being displayed/wrapped
when the terminal is not wide enough for the longest line:

```
% pdm show coverage|cat -A
Name:                  coverage                                       …$
Latest version:        6.0.2                                          …$
…
Project URLs:          Documentation: https://coverage.readthedocs.io …$
                       Funding: https://example.com/?utm_source=foobar…$
…
```

This patch just adds `rstrip()`, which gets rid of it, while still
keeping the maybe too long underline for the header (not used with
pdm-show).

Maybe this could use $COLUMNS / the terminal width to limit the length
of the header underline, but that is not really needed / not trivial.

### DIFF
--- a/news/680.bugfix.md
+++ b/news/680.bugfix.md
@@ -1,0 +1,1 @@
+Remove trailing whitespace with terminal output of tables (via `project.core.ui.display_columns`), fixing unnecessary wrapping due to / with empty lines full of spaces in case of long URLs in the last column.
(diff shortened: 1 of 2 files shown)


frostming:
A news fragment is required, thanks

PR #726, opened 2021-11-09
[RFC/WIP] Fix find_candidates with no requires_python

blueyed:
This fixes installation/resolving of "coveralls", which would pick
version 1.11.1 without a required Python version (`requires-python` in
`pyproject.toml`'s `[project]` section'), since this is the last
version not requiring a specific Python version.

Note that pdm 1.9 changed the code in this area, but with the default of
`allow_all=False` would apply the same behavior there.
Ref: https://github.com/pdm-project/pdm/pull/658 (6be985aa5ca)

## Pull Request Check List

### DIFF
--- a/pdm/models/repositories.py
+++ b/pdm/models/repositories.py
@@ -120,8 +120,13 @@ class BaseRepository:
             if requirement.specifier.contains(  # type: ignore
                 c.version, allow_prereleases  # type: ignore
             )
-            and requires_python.is_subset(c.requires_python)
         ]
+        if requires_python:
+            applicable_cans = [
+                c
+                for c in applicable_cans
+                if requires_python.is_subset(c.requires_python)
+            ]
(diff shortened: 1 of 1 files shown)


frostming:
I didn't quite get what is going wrong by your description.

blueyed:
The problem is that with a missing `requires-python` entry it will skip candidates that require a specific version of Python.

This confused me when I was trying to create a reproducer for another issue, and it would pick an older version for "coveralls" (the last one without a `requires-python` setting).

I am not sure if ignoring Python version constraints is the desired behavior in this case, but wanted to get the patch out for comments and see if maybe some test(s) fail due to this change.

frostming:
The resolution has been improved, please check if your problem still persists, and feel free to open a new PR if you come up with some working implementation.

PR #749, opened 2021-11-19
Feat: add support for Conda environments

agoose77:
## Pull Request Check List

This PR adds support for `CONDA_PREFIX` where virtual environments are implemented using Conda.

### DIFF
--- a/pdm/models/candidates.py
+++ b/pdm/models/candidates.py
@@ -26,6 +26,7 @@ from pdm.utils import (
     create_tracked_tempdir,
     expand_env_vars_in_auth,
     get_rev_from_url,
+    get_venv_like_prefix,
     path_replace,
     populate_link,
     url_without_fragments,
@@ -409,10 +410,14 @@ class Candidate:
         elif self.req.editable:
             if self.environment.packages_path:
                 src_dir = self.environment.packages_path / "src"
-            elif os.getenv("VIRTUAL_ENV"):
-                src_dir = Path(os.environ["VIRTUAL_ENV"]) / "src"
             else:
-                src_dir = Path("src")
+                venv_prefix = get_venv_like_prefix(
+                    self.environment.interpreter.executable
+                )
+                if venv_prefix is not None:
+                    src_dir = venv_prefix / "src"
+                else:
+                    src_dir = Path("src")
             if not src_dir.is_dir():
                 src_dir.mkdir()
             ireq.ensure_has_source_dir(str(src_dir))
--- a/pdm/utils.py
+++ b/pdm/utils.py
@@ -436,12 +436,38 @@ def is_path_relative_to(path: str | Path, other: str | Path) -> bool:
 
 
 def is_venv_python(interpreter: str | Path) -> bool:
-    """Check if the given interpreter path is from a virtualenv"""
+    """Check if the given interpreter path is from a virtualenv,
+    and return the prefix if found.
+    """
     interpreter = Path(interpreter)
     if interpreter.parent.parent.joinpath("pyvenv.cfg").exists():
         return True
+
     virtual_env = os.getenv("VIRTUAL_ENV")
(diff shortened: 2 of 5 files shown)


pdm/utils.py
```diff
@@ -513,3 +516,15 @@ def pdm_scheme(base: str) -> dict[str, str]:
             "headers": "{base}/include",
         }
     return sysconfig.get_paths("pdm", vars={"base": base}, expand=True)
+
+
+def get_env_prefix() -> Optional[Path]:
```
frostming, line 521: Make the name more obviously related to "venv"

frostming (changes requested):

pdm/utils.py
frostming: also `-> Path | None` to keep consistent

pdm/models/candidates.py
```diff
@@ -409,10 +410,12 @@ def _populate_source_dir(self, ireq: pip_shims.InstallRequirement) -> None:
         elif self.req.editable:
             if self.environment.packages_path:
                 src_dir = self.environment.packages_path / "src"
-            elif os.getenv("VIRTUAL_ENV"):
-                src_dir = Path(os.environ["VIRTUAL_ENV"]) / "src"
             else:
-                src_dir = Path("src")
+                virtual_env_prefix = get_env_prefix()
```
frostming, line 414: I am thinking it shouldn't be testing the env var only, but it should check `is_venv_python(self.environment.interpreter.executable)`

pdm/utils.py
agoose77: Will do - I didn't notice the future import, so I wasn't sure about using the new union syntax!

pdm/models/candidates.py
agoose77: I think you're right. In fact, I don't think we should really be checking `venv` outside of `Project`, because once we have chosen the interpreter, we want to use the same sysconfig information everywhere.

frostming: This is to keep the same behavior of `pip`, so that users know where to find the source when using an editable install.

1. venv pyhon: `<venv_prefix>/src`
2. non-venv python: `$(pwd)/src`
3. (PEP 582 specific): `__pypackages__/$major.$minor/src`

agoose77: Will reply below in comments :)

agoose77:
I don't have time to work on this *right* now, but I'm thinking about the following changes:

- Add `is_venv` to `PythonInfo` or `Environment`, to replace (remove) `pdm.utils.is_venv_python`
- Replace call to `get_env_prefix` in https://github.com/pdm-project/pdm/blob/661a4b41a59a403d3d2de4513193366a0ff28ad3/pdm/models/candidates.py#L414-L416 with a check of `python/environment.is_venv`

This would change the behaviour — right now, as you point out, we don't check if `project.python.interpreter` is a virtual environment.

frostming:
> I don't have time to work on this _right_ now, but I'm thinking about the following changes:
> 
> * Add `is_venv` to `PythonInfo` or `Environment`, to replace (remove) `pdm.utils.is_venv_python`
> * Replace call to `get_env_prefix` in https://github.com/pdm-project/pdm/blob/661a4b41a59a403d3d2de4513193366a0ff28ad3/pdm/models/candidates.py#L414-L416
>    with a check of `python/environment.is_venv`
> 
> This would change the behaviour — right now, as you point out, we don't check if `project.python.interpreter` is a virtual environment.

That is a good change, go ahead with it!

agoose77:
I think we *might* want to refactor the virtual-environment support here, but I think we should do so in a different PR.

I experimented with adding this information to the `PythonInfo` class, because I feel that this knowledge belongs with the interpreter. However, it's a bigger change than just adding Conda support, so I'm holding off on it.

frostming (approved):

PR #752, opened 2021-11-19
Feat: add `minimum` save strategy

agoose77:
## Pull Request Check List

This PR adds `--save-minimum` to `pdm add` for cases where the user doesn't want an upper bound (e.g. non SemVer uses)

See also
- https://bernat.tech/posts/version-numbers/
- https://hynek.me/articles/semver-will-not-save-you/


news/feat-add-lb-strategy.md
```diff
@@ -0,0 +1 @@
+    Add support for saving only the lower bound `x <= VERSION` when adding dependencies.
```
pawamoy, line 1: I think you meant `>=`?
```suggestion
    Add support for saving only the lower bound `x >= VERSION` when adding dependencies.
```

agoose77: Oh no... yes! Thanks for the quick catch.

agoose77:
I just want to take the time to say how cool this project is. It's really easy to navigate the source code, and I like the fact that the scope of `pdm` is quite small. I don't use PEP 582, and it doesn't feel like I'm punished for that, which is also nice!

frostming:
@agoose77 Welcome to PDM! Care to update the completion scripts under `pdm/cli/completions`? It is okay if you won't, I will do that instead.

agoose77:
> @agoose77 Welcome to PDM! Care to update the completion scripts under `pdm/cli/completions`? It is okay if you won't, I will do that instead.

Darn, I thought I'd caught everything! Will do.

agoose77:
@frostming I've updated the completions, and modified the completions script to support ZSH and powershell. Was this a deliberate omission, or am I OK to keep the fix in this PR?

frostming:
oh, just keep them and add a single option manually

agoose77:
> oh, just keep them and add a single option manually

OK, will revert the change. Out of interest, why do we not use the completions generator for powershell/zsh?

frostming:
> > oh, just keep them and add a single option manually
> 
> 
> 
> OK, will revert the change. Out of interest, why do we not use the completions generator for powershell/zsh? I noticed some spelling errors in the hand-crafted ones.

because they are hand written, for some value completion. no autocomplete engine can generate such customized contents.

agoose77:
Right, makes sense. I think this PR is done from my end, are there any outstanding change to be made? :)

frostming:
LGTM, thanks for fixing all the typos

frostming (approved):

PR #760, opened 2021-11-22
Refactor: split Environment into `PEP582Environment` and `GlobalEnvironment`

agoose77:
## Pull Request Check List

This PR tries to separate PEP-582 logic from the non-582 logic in the existing `Environment` base class. I've tried to use duck-typing rather than `isinstance`, but mypy doesn't seem to know how to narrow after a `hasattr` guard, so in some places `isinstance` or `type: ignore` is used.

When modifying the environment, it became apparent that the `Project` and `Environment` classes are quite inter-dependent. For now, I've just made the interpreter aspect more evident by passing it in as an argument. This provides a motivation in the source code for re-creating the environment if the interpreter is changed.

The `Environment` classes only depend on a few project attributes:
- `interpreter` (passed in as arg now)
- `cache_dir`
- `sources`
- `root`

I think we can eliminate some of these by moving the `get_finder` functionality into a new/different class. This would mean changing `Environment` to be responsible for the Python (virtual) environment abstraction only.

Meanwhile, the `Project` class only depends upon environment in order to pass it to other classes, which refer back to the project indirectly via `environment.project`. I think we can decouple these by moving `get_finder` elsewhere as discussed above.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+layout_anaconda pdm
--- a/pdm/cli/actions.py
+++ b/pdm/cli/actions.py
@@ -31,14 +31,13 @@ from pdm.exceptions import NoPythonVersion, PdmUsageError, ProjectError
 from pdm.formats import FORMATS
 from pdm.formats.base import array_of_inline_tables, make_array, make_inline_table
 from pdm.models.candidates import Candidate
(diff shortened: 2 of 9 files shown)


frostming:
Feel free to open a new PR if you come up with some working implementation.

PR #762, opened 2021-11-27
lint using flake8-simplify

danieleades:


pdm/builders/base.py
```diff
@@ -90,7 +90,7 @@ def log_subprocessor(
         outstream.stop()
 
 
-class _Prefix:
+class _Prefix:  # noqa: SIM119
```
danieleades, line 93: `flake8-simplify` suggests using a dataclass here. that's not a terrible idea, but it's a larger refactor, so i've just suppressed it for now

pawamoy:
Nice improvements! Big fan of `contextlib.suppress` :slightly_smiling_face: 
The equality operations are more readble that way indeed.
I have to say -1 on using dataclasses though. I've found them not that useful, but especially not that flexible. If you need to customize the logic a bit more (default values, changing values at init, pos-only/kw-only args, etc.), it won't always do.
For the with statements, the lower indentation is nice, but the statements themselves are not more readable. Too bad the multiline with statements are only available on Python 3.10!

PR #764, opened 2021-11-27
lint using pyupgrade

danieleades:


danieleades:
any ideas why this is failing?

frostming:
I think I have explained this to you the last time you try to add `pyupgrade`: pdm has some compatibility module that needs to run on Python < 3.6

danieleades:
> I think I have explained this to you the last time you try to add `pyupgrade`: pdm has some compatibility module that needs to run on Python < 3.6

in that case, should the project specify python 3.6 as it's minimum interpreter version? (it's currently specified as 3.7).
Or, if this is limited to a specific module, then I can exclude that from pyupgrade's attentions

PR #765, opened 2021-11-27
lint using pep8-naming

danieleades:
enforces pep8-compliant object names

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,7 @@ repos:
       - id: flake8
         additional_dependencies:
           - flake8-bugbear
+          - pep8-naming
 
   - repo: https://github.com/pycqa/isort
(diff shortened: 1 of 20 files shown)


frostming:
I am a bit conservative in adding too many addons in pre-commit linting. Most of them are "can have" instead of "should have", I accept some while rejecting others.

Welcome to contribute code improvements, thanks.

danieleades:
> I am a bit conservative in adding too many addons in pre-commit linting. Most of them are "can have" instead of "should have", I accept some while rejecting others.
> 
> Welcome to contribute code improvements, thanks.

note that the non-pep8-compliant PascalCase variable names picked up by this lint are `also` a mypy "strict" checking error, so you would also need to suppress the mypy error if you wanted to use the stricter type-checking

PR #779, opened 2021-12-08
Feature: prerelease option for add and update

### LINKED ISSUE #774: Support per-package `--prerelease` option

Poetry provides the user with the ability to specify `--prerelease` on a per `poetry add` basis. This is likely more useful than a per-project setting, because (in my experience) most of the time it's only certain packages that should be pre-releases.

A  `--prerelease` option for `pdm add` which allows the user to install pre-releases for the given dependencies. This would require an entry in `tool.pdm`, e.g. `tool.pdm.pre-dependences` to store this for future resolutions.

frostming:
## Pull Request Check List

Close #774


prudnikov:
This does not work with 
```
[tool.pdm.resolution]
allow-prereleases = false
```

PR #806, opened 2021-12-17
Small documentation improvements

lyz-code:
docs: clarify the use of `-d` without `-G`
It's clearer that you can use `pdm add --dev pytest`, because when you
arrive this point in the documentation, you don't know that `-d` means
`--dev`.

docs: add the `minimum` save strategy

It's the default and it was not mentioned! :P

docs: mark the `reuse` update strategy as the default

docs: fix typos

docs: add the `pdm-publish` as an option to upload packages

docs: remove duplicated install cache version

It's already documented [here](project.md#cache-the-installation-of-wheels).

docs: Update the Python versions of the github action snippet

## Pull Request Check List

Just documentation, so no need to add anything under `news` or `tests`

### DIFF
--- a/docs/docs/pyproject/pep621.md
+++ b/docs/docs/pyproject/pep621.md
@@ -43,7 +43,7 @@ PDM can also read version from SCM tags. If you are using git or hg as the versi
 version = {use_scm = true}
 ```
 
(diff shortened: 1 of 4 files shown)


frostming (approved):
LGTM

PR #816, opened 2021-12-22
Use symlinks to cache installed packages when it is supported by the file system

### LINKED ISSUE #814: Prefer symlink when possible to make links to install cache

frostming:
## Pull Request Check List

Close #814


pdm/utils.py
```diff
@@ -534,3 +534,20 @@ def pdm_scheme(base: str) -> dict[str, str]:
...
+def fs_supports_symlink() -> bool:
+    if not hasattr(os, "symlink"):
+        return False
+    if sys.platform == "win32":
+        with tempfile.NamedTemporaryFile(prefix="TmP") as tmp_file:
+            temp_dir = os.path.dirname(tmp_file.name)
+            dest = os.path.join(temp_dir, "{}-{}".format(tmp_file.name, "b"))
+            try:
+                os.symlink(tmp_file.name, dest)
+                return False
```
pawamoy, line 549: `return True`? 😄

frostming: Oh, you are right

PR #822, opened 2021-12-29
Improve the install cache with symlinks method

### LINKED ISSUE #820: File exists error when symlinking (namespace packages)

Related to #817

## Steps to reproduce

`pdm install -v` with cache feature enabled.

```
add sphinxcontrib-serializinghtml failed:
  File "/home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.9.9/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/pawamoy/.local/pipx/venvs/pdm/lib/python3.9/site-packages/pdm/installers/synchronizers.py", line 190, in install_candidate
    self.manager.install(can)
  File "/home/pawamoy/.local/pipx/venvs/pdm/lib/python3.9/site-packages/pdm/installers/manager.py", line 38, in install
    installer(candidate.build(), self.environment, candidate.direct_url())
  File "/home/pawamoy/.local/pipx/venvs/pdm/lib/python3.9/site-packages/pdm/installers/installers.py", line 160, in install_wheel_with_cache
    dist_info_dir = _install_wheel(
  File "/home/pawamoy/.local/pipx/venvs/pdm/lib/python3.9/site-packages/pdm/installers/installers.py", line 259, in _install_wheel
    destination.finalize_installation(
  File "/home/pawamoy/.local/pipx/venvs/pdm/lib/python3.9/site-packages/pdm/installers/installers.py", line 83, in finalize_installation
    target.symlink_to(child)
  File "/home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.9.9/lib/python3.9/pathlib.py", line 1403, in symlink_to
    self._accessor.symlink(target, self, target_is_directory)
  File "/home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.9.9/lib/python3.9/pathlib.py", line 456, in symlink
    return os.symlink(a, b)
FileExistsError: [Errno 17] File exists: '/home/pawamoy/.cache/pdm/packages/sphinxcontrib_serializinghtml-1.1.5-py2.py3-none-any/lib/sphinxcontrib' -> '/media/data/dev/mkdocstrings/__pypackages__/3.7/lib/sphinxcontrib'
```

## Actual behavior

Namespace packages expose directories with the same name. If namespace package N.A is already installed, trying to install N.B fails, because directory N already exists.

## Expected behavior

In the case of namespace packages, instead of symlinking the top directory, it should mkdir the top directory and symlink the namespace packages instead. Probably edgy.

An alternative would be to provide a configuration option to disable symlinks (pth files seemed to work better).

## Environment Information

```bash
PDM version:        1.12.1
Python Interpreter: /home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.7.12/bin/python3.7 (3.7)
Project Root:       /media/data/dev/mkdocstrings
Project Packages:   /media/data/dev/mkdocstrings/__pypackages__/3.7
{
  "implementation_version": "3.7.12",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.15.11-arch2-1",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT Wed, 22 Dec 2021 09:23:54 +0000",
  "python_full_version": "3.7.12",
  "python_version": "3.7",
  "sys_platform": "linux"
}
```

frostming:
## Pull Request Check List

Close #820

### DIFF
--- a/pdm/installers/installers.py
+++ b/pdm/installers/installers.py
@@ -29,13 +29,13 @@ if TYPE_CHECKING:
 
 
 @lru_cache()
-def _is_python_package(root: str) -> bool:
+def _is_python_package(root: str | Path) -> bool:
     for child in Path(root).iterdir():
         if child.is_file():
             if child.suffix in (".py", ".pyc", ".pyo", ".pyd"):
                 return True
         else:
-            if child.joinpath("__init__.py").exists():
+            if _is_python_package(child):
                 return True
     return False
 
--- a/pdm/installers/installers.py
+++ b/pdm/installers/installers.py
@@ -29,9 +29,23 @@ if TYPE_CHECKING:
 
 
 @lru_cache()
-def _is_namespace_package(root: str, files: frozenset[str]) -> bool:
-    if "__init__.py" not in files:
-        return any(f.endswith(".py") for f in files)  # PEP 420 style
+def _is_python_package(root: str) -> bool:
+    for child in Path(root).iterdir():
+        if child.is_file():
+            if child.suffix in (".py", ".pyc", ".pyo", ".pyd"):
+                return True
+        else:
+            if child.joinpath("__init__.py").exists():
+                return True
+    return False
+
+
+@lru_cache()
+def _is_namespace_package(root: str) -> bool:
+    if not _is_python_package(root):
+        return False
+    if not os.path.exists(os.path.join(root, "__init__.py")):  # PEP 420 style
+        return True
     int_py_lines = [
         line.strip()
         for line in Path(root, "__init__.py").open(encoding="utf-8")
@@ -66,7 +80,7 @@ def _create_symlinks_recursively(source: str, destination: str) -> Iterable[str]
             continue
         relpath = os.path.relpath(root, source)
         destination_root = os.path.join(destination, relpath)
-        if not is_top and not _is_namespace_package(root, frozenset(files)):
+        if not is_top and not _is_namespace_package(root):
             # A package, create link for the parent dir and don't proceed
             # for child directories
             if os.path.exists(destination_root):
--- a/pdm/installers/installers.py
+++ b/pdm/installers/installers.py
@@ -77,6 +77,7 @@ def _create_symlinks_recursively(source: str, destination: str) -> Iterable[str]
         )
         bn = os.path.basename(root)
         if bn == "__pycache__" or bn.endswith(".dist-info"):
+            dirs[:] = []
             continue
         relpath = os.path.relpath(root, source)
         destination_root = os.path.join(destination, relpath)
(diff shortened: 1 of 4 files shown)


pdm/installers/installers.py
```diff
@@ -25,6 +25,85 @@
...
+            # A package, create link for the parent dir and don't proceed
+            # for child directories
+            if os.path.exists(destination_root):
+                os.remove(destination_root)
+            os.symlink(root, destination_root, True)
+            yield relpath
+            dirs[:] = []
+            continue
+        # Otherwise, the directory is likely a namespace package,
+        # mkdir and create links for all files inside.
```
pawamoy, line 94: Very nice1 From glancing at the code, it seems you support nested namespace packages?

frostming: Yeah, but I guess it won't work if the namespace packages are nested in more than 2 levels:

```
a   # <- this will be detected as normal directory and create link for it
    b
        c
           __init__.py
```

But I think it is enough.

pawamoy: Ah I see. Too bad, I actually have a doubly nested namespace package 😅 
But it's OK, pth files work fine, and I'm planning to change that to a level-1 namespace package later anyway.
Thanks!

frostming: it is easy for a recursive detection, but i am not sure if a deeply nested package can still be called a 'package'. However, if you agree every py files inside a wheel must belong to a package, the answer would be Yes.

pawamoy: Well, that's how PEP420 packages work, so I'd say yes, with `a/b/c/d/e.py`, `a` is a (namespace) package (and b, c, and d are sub-namespace packages, or namespace sub-packages haha).

pdm/installers/installers.py
```diff
@@ -25,6 +25,86 @@
...
 
+    from pdm.models.environment import Environment
+
+
+@lru_cache()
+def _is_python_package(root: str | Path) -> bool:
+    for child in Path(root).iterdir():
+        if (
+            child.is_file()
+            and child.suffix in (".py", ".pyc", ".pyo", ".pyd")
```
pawamoy, line 36: I wonder if `.so` files should be checked as well.

frostming: I've thought of that, but it was not that easy to tell a `.so` is a python extension module or not. So this case can be ignored given how rare it is to put a `.so` inside a PEP 420 namespace package.

pawamoy: I've just encountered that case :joy: That's actually why I'm commenting this :joy:
But yeah, more than the extension, the whole file name would need to be checked against some regular expression like `^(?P<module_name>\w+).(?P<build_info>>.+).so$` (from my limited knowledge)😕

Issue #874, opened 2022-01-30
pip 22.0.1 broke `pip_shims.unpack_url`


KerberosMorphy:
## Steps to reproduce

Hard to give a way to reproduce since I make lots of experimentation in Docker at the same time.
I've tried do recreate a small example but I couldn't.

In all case, the code speak for itself.

## Current situation

Error raised since pip expect `verbosity` as fourth positional argument.

```
[TypeError]: unpack_url() missing 1 required positional argument: 'verbosity'
```

https://github.com/pdm-project/pdm/blob/afb3244b43b128ee52bb2faf88e1b47fd64aa443/pdm/models/candidates.py#L257-L262

## Change in pip code

They change the behavior of `unpack_url` from 21.3.1 to 22.0.1:

```diff
def unpack_url(
    link: Link,
    location: str,
    download: Downloader,
+   verbosity: int,
    download_dir: Optional[str] = None,
    hashes: Optional[Hashes] = None,
) -> Optional[File]:
```

https://github.com/pypa/pip/blob/c73ac8d6bcf4f64041cafeccd2125cca052abed9/src/pip/_internal/operations/prepare.py#L174-L181

## Environment Information

```bash
PDM version:        1.12.6
Python Interpreter: /usr/local/bin/python (3.9)
Project Root:       /app
Project Packages:   /app/__pypackages__/3.9
{
  "implementation_name": "cpython",
  "implementation_version": "3.9.10",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.10.60.1-microsoft-standard-WSL2",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Wed Aug 25 23:20:18 UTC 2021",
  "python_full_version": "3.9.10",
  "platform_python_implementation": "CPython",
  "python_version": "3.9",
  "sys_platform": "linux"
}
```

H-Plus-Time:
In the meantime, a workaround is to specify the pip dependency in your install-pdm step:
`curl -sSL https://raw.githubusercontent.com/pdm-project/pdm/main/install-pdm.py | python3 - --dep "pip<22.0.1"`

That or the other ways to specify that dependency override (the PDM_DEPS env var for instance).

CharString:
Maybe we should consider pinning the pdm dependencies. That would prevent these breakages, at least when installed with `pipx`.

pawamoy:
> Maybe we should consider pinning the pdm dependencies. That would prevent these breakages

[...]

PR #875, opened 2022-01-31
Add missing verbosity flag

q0w:
#874

## Pull Request Check List

### DIFF
--- a/pdm/models/candidates.py
+++ b/pdm/models/candidates.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import os
 import re
 import sys
@@ -35,9 +36,9 @@ from pdm.utils import (
 )
 
 if sys.version_info >= (3, 8):
-    from importlib.metadata import Distribution, PathDistribution, version
+    from importlib.metadata import Distribution, PathDistribution
 else:
-    from importlib_metadata import Distribution, PathDistribution, version
+    from importlib_metadata import Distribution, PathDistribution
 
 if TYPE_CHECKING:
     from pdm.models.environment import Environment
@@ -255,7 +256,7 @@ class Candidate:
             self._populate_source_dir(ireq)
             if not self.link.is_existing_dir():
                 assert ireq.source_dir
-                if tuple(map(int, version("pip").split("."))) >= (22, 0):
+                if "verbosity" in inspect.getfullargspec(pip_shims.unpack_url).args:
                     unpack_url = partial(pip_shims.unpack_url, verbosity=0)
                 else:
                     unpack_url = partial(pip_shims.unpack_url)
--- a/pdm/models/candidates.py
+++ b/pdm/models/candidates.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
-import inspect
 import os
 import re
 import sys
-from functools import partial
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast, no_type_check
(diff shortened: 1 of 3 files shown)


pdm/models/candidates.py
```diff
@@ -258,6 +258,7 @@ def prepare(self, allow_all: bool = False) -> None:
                     self.link,
                     ireq.source_dir,
                     downloader,
+                    0,
```
CharString, line 261: I'd prefer `verbosity=0` to be explicit.

CharString:
This would solve #874

pawamoy:
Is the flag backported to previous pip versions? If not, and if always using it in PDM, then I think the minimum required pip version in pyproject.toml should also be upgraded?

frostming:
You need to ensure the fix works for the specified version range of `pip` in `dependencies`. But before that, I will cut a new release pinning `pip<22` to fix the breakage.

Thanks

q0w:
Should I check if `unpack_url` has `verbosity` flag in function signature?

pawamoy:
This, or check pip's version using importlib.metadata.

pdm/models/candidates.py
```diff
@@ -254,7 +255,11 @@ def prepare(self, allow_all: bool = False) -> None:
             self._populate_source_dir(ireq)
             if not self.link.is_existing_dir():
                 assert ireq.source_dir
-                downloaded = pip_shims.unpack_url(
+                if tuple(map(int, version("pip").split("."))) >= (22, 0):
```
q0w, line 258: Its wrong, how I should parse version?

frostming (changes requested):
This need to be done in the `pip_shims` model, and other models just import and use it. That is what the name `pip_shims` stands for.

q0w:
> This needs to be done in the `pip_shims` module, and other modules just import and use it. That is what the name `pip_shims` stands for.

Ok. Should it be tested somehow?

pdm/models/pip_shims.py
```diff
@@ -79,3 +80,10 @@ def get_package_finder(
 
     build_kwargs["ignore_requires_python"] = ignore_requires_python
     return install_cmd._build_package_finder(**build_kwargs)  # type: ignore
+
+
+unpack_url = (
+    partial(unpack_url, verbosity=0)
+    if "verbosity" in inspect.getfullargspec(unpack_url).args
+    else partial(unpack_url, verbosity=0)
```
pawamoy, line 88: I think you meant

```suggestion
    unpack_url
```
?

q0w: @pawamoy mypy fails if just `else unpack_url`

pawamoy: `# type: ignore` it, or use a partial but without `verbosity=0` I guess?

pawamoy: Ah, no, better: since there's only one thing to do, remove the ternary expression:

```python
if "verbosity" in inspect.getfullargspec(unpack_url).args:
    unpack_url = partial(unpack_url, verbosity=0)
```

q0w: Is it right in the global scope?

pawamoy: Sure, why not?

q0w: Ok.

frostming (approved):

frostming:
Thanks

PR #882, opened 2022-02-02
Ignore uninstalled plugin configs

### LINKED ISSUE #872: pdm plugin config is not cleaned up after removal

## Steps to reproduce

```bash
❯ pdm plugin add pdm-venv
✔ Installation succeeds.
❯ pdm config venv.in_project True
❯ pdm plugin remove pdm-venv -y
✔ Uninstallation succeeds.
❯ pdm config -v
...
  File "/Users/*/Library/Python/3.10/bin/pdm", line 10, in <module>
    sys.exit(main())
  File "/Users/*/Library/Application Support/pdm/venv/lib/python3.10/site-packages/pdm/core.py", line 219, in main
    return Core().main(args)
  File "/Users/*/Library/Application Support/pdm/venv/lib/python3.10/site-packages/pdm/core.py", line 161, in main
  File "/Users/*/Library/Application Support/pdm/venv/lib/python3.10/site-packages/pdm/core.py", line 156, in main
    f(options.project, options)
  File "/Users/*/Library/Application Support/pdm/venv/lib/python3.10/site-packages/pdm/cli/commands/config.py", line 32, in handle
    self._list_config(project, options)
  File "/Users/*/Library/Application Support/pdm/venv/lib/python3.10/site-packages/pdm/cli/commands/config.py", line 48, in _list_config
    "# " + project.global_config._config_map[key].description
KeyError: 'venv.in_project'
```

## Actual behavior

After executing `pdm plugin remove <plugin_name>`, if the plugin has added some configuration to pdm and user has updated any of those configurations, it creates an entry either in local or global **config.toml** based on the user flag used. These config entries stay persistent and then lead to this `KeyError`.

## Expected behavior

After executing `pdm plugin remove <plugin_name>` all the plugin config should also be cleaned up.

## Environment Information

```bash
PDM version:        1.12.6
Python Interpreter: /usr/local/opt/python@3.10/bin/python3.10 (3.10)
Project Root:       /Users/*/demo
Project Packages:   /Users/*/demo/__pypackages__/3.10
{
  "implementation_version": "3.10.1",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "20.6.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 20.6.0: Wed Jan 12 22:22:42 PST 2022; root:xnu-7195.141.19~2/RELEASE_X86_64",
  "python_full_version": "3.10.1",
  "python_version": "3.10",
  "sys_platform": "darwin"
}
```

xanterx:
## Pull Request Check List

Close #872

### DIFF
--- a/news/872.bugfix.md
+++ b/news/872.bugfix.md
@@ -1,1 +1,1 @@
-- pdm now ignores config of uninstalled plugins
+PDM now ignores configuration of uninstalled plugins.
--- a/news/872.bugfix.md
+++ b/news/872.bugfix.md
@@ -1,0 +1,1 @@
+- pdm now ignores config of uninstalled plugins
--- a/pdm/cli/commands/config.py
+++ b/pdm/cli/commands/config.py
@@ -44,14 +44,12 @@ class Command(BaseCommand):
         with ui.indent("  "):
             for key in sorted(project.global_config):
                 try:
-                    project.global_config._config_map[key]
+                    global_config_item = project.global_config._config_map[key]
                 except KeyError:
                     # ignoring removed plugin keys
                     continue
                 ui.echo(
-                    termui.yellow(
-                        "# " + project.global_config._config_map[key].description
-                    ),
+                    termui.yellow("# " + global_config_item.description),
                     verbosity=termui.DETAIL,
                 )
                 ui.echo(f"{termui.cyan(key)} = {project.global_config[key]}")
@@ -63,14 +61,12 @@ class Command(BaseCommand):
         with ui.indent("  "):
             for key in sorted(project.project_config):
                 try:
-                    project.project_config._config_map[key]
+                    project_config_item = project.project_config._config_map[key]
                 except KeyError:
                     # ignoring removed plugin keys
                     continue
                 ui.echo(
-                    termui.yellow(
-                        "# " + project.project_config._config_map[key].description
-                    ),
+                    termui.yellow("# " + project_config_item.description),
                     verbosity=termui.DETAIL,
                 )
                 ui.echo(f"{termui.cyan(key)} = {project.project_config[key]}")
--- a/pdm/cli/commands/config.py
+++ b/pdm/cli/commands/config.py
@@ -43,6 +43,11 @@ class Command(BaseCommand):
         ui.echo("Home configuration ({}):".format(project.global_config._config_file))
         with ui.indent("  "):
             for key in sorted(project.global_config):
+                try:
+                    project.global_config._config_map[key]
+                except KeyError:
+                    # ignoring removed plugin keys
+                    continue
                 ui.echo(
                     termui.yellow(
                         "# " + project.global_config._config_map[key].description
@@ -57,6 +62,11 @@ class Command(BaseCommand):
         )
         with ui.indent("  "):
             for key in sorted(project.project_config):
+                try:
+                    project.project_config._config_map[key]
+                except KeyError:
+                    # ignoring removed plugin keys
+                    continue
                 ui.echo(
                     termui.yellow(
                         "# " + project.project_config._config_map[key].description
--- a/pdm/project/config.py
+++ b/pdm/project/config.py
@@ -195,7 +195,12 @@ class Config(MutableMapping[str, str]):
             tomlkit.dump(toml_data, fp)  # type: ignore
 
     def __getitem__(self, key: str) -> Any:
-        env_var = self._config_map[key].env_var
+        try:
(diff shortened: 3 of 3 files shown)


news/872.bugfix.md
```diff
@@ -0,0 +1 @@
+- pdm now ignores config of uninstalled plugins
```
pawamoy, line 1:
```suggestion
PDM now ignores configuration of uninstalled plugins.
```

pdm/project/config.py
```diff
@@ -195,15 +195,20 @@ def _save_config(self) -> None:
             tomlkit.dump(toml_data, fp)  # type: ignore
 
     def __getitem__(self, key: str) -> Any:
-        env_var = self._config_map[key].env_var
+        try:
+            config = self._config_map[key]
+        except KeyError:
+            # ignoring removed plugin keys
+            return
+        env_var = config.env_var
```
frostming, line 203: I think it would be better to keep the KeyError for `__getitem__` instead of returning a `None`.

pdm/cli/commands/config.py
```diff
@@ -43,6 +43,11 @@ def _list_config(self, project: Project, options: argparse.Namespace) -> None:
         ui.echo("Home configuration ({}):".format(project.global_config._config_file))
         with ui.indent("  "):
             for key in sorted(project.global_config):
+                try:
+                    project.global_config._config_map[key]
```
frostming, line 47: Why not assign to an intermediate variable as it is used later.

frostming (changes requested):

pdm/project/config.py
xanterx: Raising the KeyError here will fall back to the previous behaviour.  With this traceback
```bash
  File "D:\projects\pdm\pdm\project\config.py", line 199, in __getitem__
    config = self._config_map[key]
KeyError: 'venv.in_project'

During handling of the above exception, another exception occurred:

  File "C:\Users\**\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\**\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\projects\pdm\.venv\Scripts\pdm.exe\__main__.py", line 7, in <module>
  File "D:\projects\pdm\pdm\core.py", line 219, in main
    return Core().main(args)
  File "D:\projects\pdm\pdm\core.py", line 172, in main
    if options.project.config["check_update"]:
  File "D:\projects\pdm\pdm\project\core.py", line 130, in config
    result = dict(self.global_config)
  File "D:\projects\pdm\pdm\project\config.py", line 202, in __getitem__
    raise KeyError
```
Return ignore will still tag invalid key as before.

pdm/cli/commands/config.py
xanterx: Updated

pdm/project/config.py
frostming: That is because `__iter__()` returns all presented config items, you need to exclude those that can't be recognized.

xanterx: Ok let me check that.

xanterx: So something like this 
```python
def __getitem__(self, key: str) -> Any:
    # check if keys are valid
    if key in (self._data.keys() - self._config_map.keys()):
        return
    config = self._config_map[key]
    env_var = config.env_var
```
will ignore those keys which are presented in the config file but is not a part of valid `_config_map` 
So, `__getitem__` will now still raise a key error if it is not a part of config file.

frostming: The `KeyError` is desired, don't ignore it.

You need to change `__iter__()` method to exclude keys not in `_config_map`.

xanterx: Done moved the ignore logic to `__iter__()` method

pdm/cli/commands/config.py
```diff
@@ -43,10 +43,13 @@ def _list_config(self, project: Project, options: argparse.Namespace) -> None:
         ui.echo("Home configuration ({}):".format(project.global_config._config_file))
         with ui.indent("  "):
             for key in sorted(project.global_config):
+                try:
+                    global_config_item = project.global_config._config_map[key]
```
frostming, line 47: Since the key is already excluded in `__iter__` so this shouldn't raise KeyError anymore. But keeping this is fine and  I will merge it.
Thanks

Issue #939, opened 2022-02-26
Ansi codes are not properly handled on Windows Terminal


julie777:
- [x ] I have searched the issue tracker and believe that this is not a duplicate.

**Make sure you run commands with `-v` flag before pasting the output.**

## Steps to reproduce
run git bash in Windows Terminal
run `pdm -h`

## Actual behavior
Terminal codes are output raw.

```
PDM - Python Development Master

←[33m←[1mCommands←[0m:
←[36m  {add,build,cache,completion,config,export,import,info,init,install,list,lock,plugin,remove,run,search,show,sync,update,use}
←[0m←[36m    add                 ←[0mAdd package(s) to pyproject.toml and install them
←[36m    build               ←[0mBuild artifacts for distribution
←[36m    cache               ←[0mControl the caches of PDM
←[36m    completion          ←[0mGenerate completion scripts for the given shell
```

## Expected behavior

I expected output to be formatted and colored like it is in other environments.

## Environment Information

```bash
# Paste the output of `pdm info && pdm info --env` below:
$ pdm info && pdm info --env
PDM version:        1.13.3
Python Interpreter: C:\Users\julie\AppData\Local\Programs\Python\Python310\python.EXE (3.10)
Project Root:       C:/tmp/a
Project Packages:   C:\tmp\a\__pypackages__\3.10
{
  "implementation_name": "cpython",
  "implementation_version": "3.10.2",
  "os_name": "nt",
  "platform_machine": "AMD64",
  "platform_release": "10",
  "platform_system": "Windows",
  "platform_version": "10.0.19043",
  "python_full_version": "3.10.2",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "win32"
}
```

julie777:
I will submit a pull request shortly, but I want to document the issue here.
Windows Terminal emulates and xterm unlike the Windows Console used for `cmd.exe` and `powershell`. This means that just checking for `platform=='win32'` is not sufficient to handle terminal configuration.

Colorama documents that calling `initialise.py:init` with no arguments should cause Colorama to do nothing if not on Windows. This fails to account for Windows Terminal.

[...]

PR #940, opened 2022-02-26
Wt bash fix

julie777:
## Pull Request Check List

I'm sorry but I don't understand how to do this.

- [x ] Test cases added for changed code.

Fixed Ansi codes are not properly handled on Windows Terminal #939.
The issue #939 has all the details of my understanding of the problem and how I fixed it.

I ran all pytest tests on 
- cmd.exe with Windows Console
- cmd.exe with Windows Terminal
- bash with Windows Terminal 
- powershell using Windows Terminal
- bash - Ubuntu 20 - Windows Terminal - Python 3.8.
    - 3 tests failed because only Python 3.8.10 is installed

5 tests failed on Windows because I only have Python 3.10.2 installed.

### DIFF
--- a/install-pdm.py
+++ b/install-pdm.py
@@ -167,7 +167,7 @@ def support_ansi() -> bool:
             os.getenv("ANSICON") is not None
             or os.getenv("WT_SESSION") is not None
             or "ON" == os.getenv("ConEmuANSI")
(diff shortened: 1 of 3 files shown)


frostming (changes requested):
Please update the patch according to [the comment](https://github.com/pdm-project/pdm/issues/939#issuecomment-1054887221)

PR #953, opened 2022-03-05
Fix for issue #939 ansi colors don't work on Windows Terminal

julie777:
## Pull Request Check List

I don't know how to do this
Not required

I found that a call to os.system("") with no command will set python on Windows to correctly support ANSI output.
ref: https://bugs.python.org/issue40134

### DIFF
--- a/pdm/__init__.py
+++ b/pdm/__init__.py
@@ -19,6 +19,10 @@ __all__ = (
     "Core",
 )
 
+# This is a fix for Issue 939 ansi colors don't work correctly with Windows Terminal.
+# Apparently Python doesn't initialize correctly for Windows Terminal to support ANSI
+# escape codes. The call below forces that initialization.
+# Ref: https://bugs.python.org/issue40134
 os.system("")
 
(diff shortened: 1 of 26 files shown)


pdm/__init__.py
```diff
@@ -17,6 +18,8 @@
     "Core",
 )
 
+os.system("")
```
pawamoy, line 21: It would probably be good to add a link to the Python issue in a comment above that statement :slightly_smiling_face:

frostming (changes requested):
Again, the original issue https://bugs.python.org/issue40134 only occurs on windows cmd and Windows Terminal works fine. This fix doesn't make sense.

There should be some issue with your terminal setup.

pdm/__init__.py
julie777: will do

frostming:
@julie777 ping, Are you still working on this? I hope you understand my concern behind those review comments.

frostming:
This issue should be fixed after the UI framework is switched to rich. Close it for now.

PR #967, opened 2022-03-12
docs: Add notes about using Pylance/Pyright

GreyElaina:
## Pull Request Check List

- [ ] ~~Test cases added for changed code.~~ it's a change about document.

i added a note about using pylance/pyright.

the user may see a ton of errors or warnings after they configured `"python.analysis.diagnosticMode": "workspace"`.
pyright follows the "pyrightconfig.json", and it could solve this problem.


frostming:
Thanks

PR #984, opened 2022-03-21
PDM logo use cdn

RatChips:
## Pull Request Check List

I change PDM logo's url  to the url provided by cdn in README_zh.

Most Chinese people can't access raw.githubusercontent.com, so I use cdn in the README_zh to make the image accessible properly.

Also, raw.githubusercontents.com is a free cdn for raw.githubusercontent.com, please feel free to use it.

### DIFF
--- a/README_zh.md
+++ b/README_zh.md
@@ -4,7 +4,7 @@
 
 一个现代的 Python 包管理器，支持 [PEP 582]。[English version README](README.md)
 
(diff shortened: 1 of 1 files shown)


frostming:
Thanks, I didn't know `raw.githubusercontents.com` has any difference from `raw.githubusercontent.com` and the former is a CDN.

PR #1012, opened 2022-04-02
fixed the issue where license-expression is read as None

k4yt3x:
## Pull Request Check List

## The Issue

In the latest version of PDM (1.13.6), `license` shows up as `None` even if `license-expression` is defined in `pyproject.toml`. I came across this issue while running `pdm show` in my own repo:

image

## The Fix

The key for `license-expression` is uses a hyphen `-` instead of an underscore `_`. The changes in this PR produces the following effects:

- If both `license-expression` and `license` are defined, prioritize `license-expression`
- If only `license` is present, use `license`
- If neiter are present, use `None`

When both are defined, `license-expression` is prioritized:

image

If only `license` exists, use the value of `license`:

image

If neither are present, license is `None`:

image

### DIFF
--- a/news/1012.bugfix.md
+++ b/news/1012.bugfix.md
@@ -1,0 +1,1 @@
+Fixed the bug where license is `None` even if `license-expression` is defined.
--- a/pdm/models/project_info.py
+++ b/pdm/models/project_info.py
@@ -51,7 +51,9 @@ class ProjectInfo:
             "summary": str(metadata.description),
             "author": str(metadata.author),
             "email": str(metadata.author_email),
-            "license": str(metadata.get("license-expression", metadata.get("license"))),
+            "license": str(
+                metadata.get("license-expression", getattr(metadata, "license", ""))
+            ),
             "requires-python": str(metadata.requires_python),
             "platform": "",
             "keywords": ", ".join(metadata.keywords or []),
--- a/pdm/models/project_info.py
+++ b/pdm/models/project_info.py
@@ -45,16 +45,13 @@ class ProjectInfo:
         }
 
     def _parse_self(self, metadata: Metadata) -> dict[str, Any]:
-        license_expression = getattr(metadata, "license_expression", None)
-        if license_expression is None:
-            license_expression = getattr(metadata, "license", "")
         return {
             "name": str(metadata.name),
             "version": str(metadata.version),
             "summary": str(metadata.description),
             "author": str(metadata.author),
             "email": str(metadata.author_email),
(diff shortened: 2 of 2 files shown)


frostming:
What does `pyproject.toml` look like?

`metadata.license_expression` is a property that reads the license value from the proper field. and `metadata.license` is for fallback.

k4yt3x:
Here's the pyproject file: https://github.com/k4yt3x/video2x/blob/master/pyproject.toml

k4yt3x:
I checked this format against [this article](https://peps.python.org/pep-0639/#advanced-example).

image

frostming:
Can you verify if the following fix works?
```diff
if license_expression is None:
-    license_expression = getattr(metadata, "license", "")
+    license_expression = getattr(metadata, "license_type", "")
```

frostming:
Here is how `pdm-pep517` reads the license:

```python
    @property
    def license_expression(self) -> Optional[str]:
        if "license-expression" in self._metadata:
            if "license" in self._metadata:
                raise MetadataError(
                    "license-expression",
                    "Can't specify both 'license' and 'license-expression' fields",
                )
            return normalize_expression(self._metadata["license-expression"])
        elif "license" in self._metadata and "text" in self._metadata["license"]:
            # warnings.warn(
            #     "'license' field is deprecated in favor of 'license-expression'",
            #     PDMWarning,
            #     stacklevel=2,
            # )
            # TODO: do not validate legacy license text,
            # remove this after PEP 639 is finalized
            return self._metadata["license"]["text"]
        elif "license-expression" not in (self.dynamic or []):
            warnings.warn("'license-expression' is missing", PDMWarning, stacklevel=2)
        return None
```

k4yt3x:
1.13.6 doesn't work for me out of the box:

image

---

> Can you verify if the following fix works?

Doesn't seem to work:

image

image

k4yt3x:
Some additional context:

- Python: 3.10.4
- OS: Arch; kernel 5.16.17

k4yt3x:
I'm quite baffled -- it works fine in a container. It's rather late now so I'll have to leave this for the moment. I'll come back to this tomorrow. I think it might be something specific to my environment. I'll cross-check on another Arch.

image

frostming:
It seems the `pdm-pep517` version installed with `pdm` is pretty old, try upgrading it.

k4yt3x:
Ah good call. Updated it and it works. I'll pin pdm-pep517 to a newer version. Thanks for the help.
Since there's no bug I'll close this PR.

PR #1039, opened 2022-04-14
Fix use_cache.json with corrupted python causes `pdm use` error

linw1995:
## Pull Request Check List

- [x] A news fragment is added in `news/` describing what is new.
- [ ] Test cases added for changed code.

## Describe what you have changed in this PR.

Before

```plain
[⚡ PIPE] ❯ pdm use -f 3.10 -v
Traceback (most recent call last):
  File "/Users/linw1995/.local/bin/pdm", line 8, in <module>
    sys.exit(main())
  File "/Users/linw1995/Documents/opensources/pdm/pdm/core.py", line 233, in main
    return Core().main(args)
  File "/Users/linw1995/Documents/opensources/pdm/pdm/core.py", line 168, in main
    raise cast(Exception, err).with_traceback(traceback)
  File "/Users/linw1995/Documents/opensources/pdm/pdm/core.py", line 163, in main
    f(options.project, options)
  File "/Users/linw1995/Documents/opensources/pdm/pdm/cli/commands/use.py", line 29, in handle
    actions.do_use(
  File "/Users/linw1995/Documents/opensources/pdm/pdm/cli/actions.py", line 571, in do_use
    cached_python = PythonInfo.from_path(use_cache.get(python))
  File "/Users/linw1995/Documents/opensources/pdm/pdm/models/python.py", line 27, in from_path
    raise InvalidPyVersion(f"Invalid Python interpreter: {path}")
pdm.exceptions.InvalidPyVersion: Invalid Python interpreter: /Users/linw1995/.pyenv/versions/3.10.0/bin/python3.10
```

After

image

PR #1046, opened 2022-04-19
Another approach for fixing errors that pdm use corrupted python

linw1995:
## Pull Request Check List

another approach for #1039

### DIFF
--- a/pdm/models/python.py
+++ b/pdm/models/python.py
@@ -7,24 +7,20 @@ from typing import Any
 from findpython import PythonVersion
 from packaging.version import Version
 
-from pdm.exceptions import InvalidPyVersion
-
 
 class PythonInfo:
     """
     A convenient helper class that holds all information of a Python interepreter.
     """
 
-    def __init__(self, py_version: PythonVersion) -> None:
+    def __init__(self, py_version: PythonVersion, valid: bool) -> None:
         self._py_ver = py_version
+        self.valid = valid
 
     @classmethod
     def from_path(cls, path: str | Path) -> "PythonInfo":
         py_ver = PythonVersion(Path(path))
-        if py_ver.executable.exists() and py_ver.is_valid():
-            return cls(py_ver)
-        else:
-            raise InvalidPyVersion(f"Invalid Python interpreter: {path}")
+        return cls(py_ver, valid=py_ver.executable.exists() and py_ver.is_valid())
 
     def __hash__(self) -> int:
         return hash(self._py_ver)
--- a/pdm/models/python.py
+++ b/pdm/models/python.py
@@ -13,7 +13,7 @@ class PythonInfo:
     A convenient helper class that holds all information of a Python interepreter.
     """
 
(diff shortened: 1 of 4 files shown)


linw1995:
All tests passed except the `3.11.0-alpha` version of Python.

pdm/models/python.py
```diff
@@ -7,24 +7,20 @@
...
-from pdm.exceptions import InvalidPyVersion
-
 
 class PythonInfo:
     """
     A convenient helper class that holds all information of a Python interepreter.
     """
 
-    def __init__(self, py_version: PythonVersion) -> None:
+    def __init__(self, py_version: PythonVersion, valid: bool) -> None:
```
frostming, line 16: better to give a default valid=True

pdm/models/python.py
```diff
@@ -7,24 +7,20 @@
...
         self._py_ver = py_version
+        self.valid = valid
 
     @classmethod
     def from_path(cls, path: str | Path) -> "PythonInfo":
         py_ver = PythonVersion(Path(path))
-        if py_ver.executable.exists() and py_ver.is_valid():
-            return cls(py_ver)
-        else:
-            raise InvalidPyVersion(f"Invalid Python interpreter: {path}")
```
frostming, line 27: does it mean we don't want it to raise for all other cases than cached selection?

frostming:
While I am trying to figure out the original problem, it seems simply a try-except around the line `cached_python = PythonInfo.from_path(use_cache.get(python))` does the trick, doesn't it? Or am I missing any context of the issue?

linw1995:
> While I am trying to figure out the original problem, it seems simply a try-except around the line `cached_python = PythonInfo.from_path(use_cache.get(python))` does the trick, doesn't it? Or am I missing any context of the issue?

The `from_path` calls in the `find_interpreters` method may raise errors too. There are many places that need to be wrapped in try-except. @frostming

pdm/models/python.py
linw1995: Yes, it will raise errors for all other cases when it actually being used, such as retrieving the Python version or executable path.

frostming:
> The `from_path` calls in the `find_interpreters` method may raise errors too. There are many places that need to be wrapped in try-except. @frostming

> Yes, it will raise errors for all other cases when it actually being used, such as retrieving the Python version or executable path.

Then it seems you just delay the validation of PythonInfo, it will raise anyway. So it is the same as catching the exception in cache selection only

linw1995:
> Then it seems you just delay the validation of PythonInfo, it will raise anyway. So it is the same as catching the exception in cache selection only

@frostming But the cache selection of python is handled via logging.  

https://github.com/pdm-project/pdm/pull/1046/files#diff-c4ffae29bb40d3aec3b55fb7c063191b2645e01c113c8387122dead05e2983d9R585-R591

Other cases will raise exceptions.

frostming:
The API still needs to be refined, but otherwise LGTM.

PR #1072, opened 2022-05-08
Use tomllib on Python 3.11

hauntsaninja:


frostming:
It is better to make it in a compatibility module and import from it. A bit ahead of time though.

frostming (changes requested):
It is better to make it in a compatibility module and import from it. A bit ahead of time though.

frostming:
I will add this to the release 2.0 milestone, thanks

frostming (approved):
LGTM

PR #1101, opened 2022-05-29
Give a default version if the version is dynamic in setup.cfg

frostming:
## Pull Request Check List

### DIFF
--- a/tests/models/test_setup_parsing.py
+++ b/tests/models/test_setup_parsing.py
@@ -1,0 +1,114 @@
+import pytest
+
+from pdm.models.setup import Setup
+
+
+@pytest.mark.parametrize(
+    "content, result",
+    [
+        (
+            """[metadata]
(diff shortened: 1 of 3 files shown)


tests/models/test_setup_parsing.py
```diff
@@ -0,0 +1,114 @@
...
+[options]
+python_requires = >=3.6
+install_requires =
+    click
+    requests
+[options.extras_require]
+tui =
+    rich
+""",
+            Setup("foo", "0.1.0", ["click", "requests"], {"tui": ["rich"]}, ">=3.6"),
```
dashsatish, line 37: Will this work if a version is just `2.0`. Can you please add this scenario too?

frostming: They are the same case, as long as it is a valid version string.

PR #1109, opened 2022-06-01
Move the test artifacts to a submodule

### LINKED ISSUE #1084: 1.15.0: pytest is failing
I'm trying to package your module as an rpm package. So I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix>
- run pytest with PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>

frostming:
## Pull Request Check List

Close #1084

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-tests/fixtures/artifacts/* filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,9 @@ on:
       - main
       - dev
     paths-ignore:
-      - 'docs/**'
-      - 'news/**'
-      - '*.md'
+      - "docs/**"
+      - "news/**"
+      - "*.md"
       - Dockerfile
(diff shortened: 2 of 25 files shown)


kloczek:
All what is still missing is version tag in https://github.com/pdm-project/test-artifacts corresponding with `pdm` tag.
Still to be honest I do not understand why those files are no in `pdm` repo.
IMO it would be better to keep all in single basket.

frostming:
They do not share the tags because the artifacts rarely changes. The best way is to checkout with git submodule. Or do you want it to be listed in the release artifacts in the github releases page.

kloczek:
Really .. IMO it would better merge those files or tag the repo the same as `pdm`.
Please do not use submodules as github is not able to provide autogenerated tar ball across multiple repos.

If those files rarely would be changing this it should be no problem with keep that stuff in `pdm` tree.
Isn't int?

Issue #1111, opened 2022-06-02
InvalidPyVersion raised when trying to add a `coverage` dependency


gtors:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

```
 pdm add coverage -vv                                                                                                                                                                                                                                                                                              [1]
Adding packages to default dependencies: coverage
======== Start resolving requirements ========
  coverage
  python>=3.10
  Adding requirement coverage
    Found matching candidates:
      <Candidate coverage 6.4.1 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.4 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.3.3 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.3.2 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.3.1 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.3 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.2 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.1.2 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.1.1 from https://pypi.org/simple/coverage/>
      <Candidate coverage 6.1 from https://pypi.org/simple/coverage/>
      ... [1435 more candidate(s)]
  Adding requirement python>=3.10
======== Starting round 0 ========
Resolving: new pin python >=3.10
======== Ending round 0 ========
======== Starting round 1 ========
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.10/site-packages/pdm/models/versions.py", line 28, in __init__
    tuple(int(v) if v != "*" else v for v in version_str.split("."))[
  File "/home/user/.local/lib/python3.10/site-packages/pdm/models/versions.py", line 28, in <genexpr>
    tuple(int(v) if v != "*" else v for v in version_str.split("."))[
ValueError: invalid literal for int() with base 10: '0a6'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user/.local/bin/pdm", line 8, in <module>
    sys.exit(main())
  File "/home/user/.local/lib/python3.10/site-packages/pdm/core.py", line 232, in main
    return Core().main(args)
  File "/home/user/.local/lib/python3.10/site-packages/pdm/core.py", line 167, in main
    raise cast(Exception, err).with_traceback(traceback)
[...]

PR #1112, opened 2022-06-02
Fix crash if dependencies have prerelease requires-python versions

Hnasar:
Recently, coverage 6.4.1 listed its requirements with an alpha version of
python and this broke PDM's version parsing:

    extras_require={
        'toml': ['tomli; python_full_version<="3.11.0a6"'],
    },

Prerelease `requires-python` versions *are* valid, per these specifications:

* https://peps.python.org/pep-0440/
* https://peps.python.org/pep-0621/#requires-python
* https://packaging.python.org/en/latest/specifications/core-metadata/#requires-python

Therefore this commit adds missing parsing support for `{a|b|rc}[N]`
pre-release specifiers which are used by python language releases:

* https://docs.python.org/3/faq/general.html#how-does-the-python-version-numbering-scheme-work

This bug meant that projects that directly or indirectly depended on coverage
were unable to update pdm.lock using commands like `update` `add` `lock` and
`install` because `pdm.models.versions.Version` would raise:

    pdm.exceptions.InvalidPyVersion: 3.11.0a6: Prereleases or postreleases are not supported for python version specifers.

Until this is fixed, projects can workaround this by depending on:

    "coverage<6.4",
    "coverage[toml]<6.4",

Fixes pdm-project/pdm#1111

### DIFF
--- a/news/1111.bugfix.md
+++ b/news/1111.bugfix.md
@@ -1,0 +1,1 @@
+Fix a bug where dependencies with `requires-python` pre-release versions caused `pdm update` to fail with `InvalidPyVersion`.
--- a/pdm/models/specifiers.py
+++ b/pdm/models/specifiers.py
@@ -69,8 +69,6 @@ class PySpecSet(SpecifierSet):
             self._analyze_specifiers()
 
     def _analyze_specifiers(self) -> None:
-        # XXX: Prerelease or postrelease specifiers will fail here, but I guess we can
-        # just ignore them for now.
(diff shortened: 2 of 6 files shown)


frostming:
Thanks for doing that, mostly LGTM. Can you remove the workaround [here](https://github.com/pdm-project/pdm/blob/b93ec61934a02db00d872670c5f23589ee2fb040/pdm/models/specifiers.py#L109-L114) since prereleases are now supported in the python specifier.

PR #1115, opened 2022-06-05
Fix #1095

### LINKED ISSUE #1095: Error with pygments and `Ctrl+C`

## Steps to reproduce
1. Download and run [mathicsscript](https://github.com/Mathics3/mathicsscript) (though this is also reproducible with ipython, or any REPL that uses pygments). 

## Actual behavior
If a command takes a while to run, if you press Ctrl+C, the process will exit, though you can not see anyhting you type until you execute `reset` or `tset`.

## Expected behavior
Whatever the program is programmed to do.

## Environment Information

```bash
PDM version:        1.14.0
Python Interpreter: /usr/bin/python3.10 (3.10)
Project Root:       /home/system/PDM/ipython
Project Packages:   /home/system/PDM/ipython/__pypackages__/3.10
{
  "implementation_version": "3.10.4",
  "os_name": "posix",
  "platform_machine": "aarch64",
  "platform_release": "5.17.1-1-aarch64-ARCH",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Mon Mar 28 23:54:20 UTC 2022",
  "python_full_version": "3.10.4",
  "python_version": "3.10",
  "sys_platform": "linux"
}

```

DUOLabs333:
## Pull Request Check List

In `pdm run`, ignores any SIGINT (Ctrl+C) to the main `pdm` process and pass it into the actual process.

### DIFF
--- a/pdm/cli/commands/run.py
+++ b/pdm/cli/commands/run.py
@@ -151,10 +151,14 @@ class TaskRunner:
                 process_env["NO_SITE_PACKAGES"] = "1"
 
         cwd = project.root if chdir else None
-        def send_int(exitnum,handler):
+
+        def send_int(exitnum, handler):
             process.send_signal(signal.SIGINT)
+
         s = signal.signal(signal.SIGINT, send_int)
-        process = subprocess.Popen(expanded_args, cwd=cwd, env=process_env, shell=shell,bufsize=0)
+        process = subprocess.Popen(
+            expanded_args, cwd=cwd, env=process_env, shell=shell, bufsize=0
+        )
         process.wait()
         signal.signal(signal.SIGINT, s)
         return process.returncode
--- a/pdm/cli/commands/run.py
+++ b/pdm/cli/commands/run.py
@@ -151,11 +151,12 @@ class TaskRunner:
                 process_env["NO_SITE_PACKAGES"] = "1"
 
         cwd = project.root if chdir else None
-        process = subprocess.Popen(expanded_args, cwd=cwd, env=process_env, shell=shell)
-        try:
-            process.wait()
-        except KeyboardInterrupt:
-            pass
+        def send_int(exitnum,handler):
+            process.send_signal(signal.SIGINT)
+        s = signal.signal(signal.SIGINT, send_int)
+        process = subprocess.Popen(expanded_args, cwd=cwd, env=process_env, shell=shell,bufsize=0)
+        process.wait()
+        signal.signal(signal.SIGINT, s)
         return process.returncode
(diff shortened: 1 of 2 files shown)


Hnasar:
@DUOLabs333 you can see your change caused a mypy error which failed the [pre-commit ci job](https://results.pre-commit.ci/run/github/230369286/1654393438.UkRL317TSSSiDILQVkkADQ)

If you want to set up pre-commit (after pip-installing it) instructions are in the [`CONTRIBUTING.md`](https://github.com/pdm-project/pdm/blob/main/CONTRIBUTING.md#code-style)

pdm/cli/commands/run.py
```diff
@@ -151,11 +152,16 @@ def _run_process(
...
-        process = subprocess.Popen(expanded_args, cwd=cwd, env=process_env, shell=shell)
-        try:
-            process.wait()
-        except KeyboardInterrupt:
-            pass
+
+        def send_int(exitnum, handler):
+            process.send_signal(signal.SIGINT)
+
+        s = signal.signal(signal.SIGINT, send_int)
```
frostming, line 159: A lambda is enough: `lambda signum, frame: process.send_signal(signum)`

DUOLabs333: Yeah, I didn't know if that was important enough to annotate it.

frostming: Yeah, if you use a normal function you need to annotate it, the signature is `(signum: int, frame: types.FrameType|None) -> None`

frostming:
Need a news fragment under `news/` folder and we are good to go 💯

DUOLabs333:
Done.

frostming (approved):

pdm/cli/commands/run.py
```diff
@@ -151,11 +152,15 @@ def _run_process(
...
-            pass
+
+        s = signal.signal(
+            signal.SIGINT, lambda signum, frame: process.send_signal(signum)
+        )
+        process = subprocess.Popen(
+            expanded_args, cwd=cwd, env=process_env, shell=shell, bufsize=0
+        )
+        process.wait()
+        signal.signal(signal.SIGINT, s)
```
blueyed, line 163: Wouldn't it also work to just `process.send_signal(signum)` in the previous code's `except KeyboardInterrupt:` block? (without (un)installing a custom signal handler)

frostming: This means the parent process will catch the interrupt and end the function, unless we `process.wait()` again

PR #1117, opened 2022-06-06
Feature: composite scripts

noirbizarre:
## Pull Request Check List

This PR adds a "composite script" feature allowing to run multiple scripts one by one.

A classical use case would be:

```toml
[tool.pdm.scripts]
lint = "flake8 mypackage"
test = "pytest -v tests/"
ci = {composite = ["lint", "test"]}
```

Then `pdm run ci` will run the `lint` command and then the `test` command. Execution stops and returns on first error.

Both `env` and `env_file` defined in `composite` action will override those defined in called tasks.

So with the following:

```toml
[tool.pdm.scripts]
echo = {shell = "echo $VAR", env = { VAR = "value"}}
composite = {composite = ["echo"], env = { VAR = "overriden"}}
```

you will have:

```console
$ pdm run echo
value
$ pdm run composite
overriden
```

### DIFF
--- a/pdm/cli/commands/run.py
+++ b/pdm/cli/commands/run.py
@@ -25,6 +25,19 @@ class TaskOptions(TypedDict, total=False):
     site_packages: bool
 
 
+def exec_opts(*options: TaskOptions | None) -> dict[str, Any]:
+    return dict(
+        env={k: v for opts in options if opts for k, v in opts.get("env", {}).items()},
+        **{
+            k: v
+            for opts in options
+            if opts
+            for k, v in opts.items()
+            if k not in ("env", "help")
+        },
+    )
+
+
 class Task(NamedTuple):
     kind: str
     name: str
@@ -38,7 +51,7 @@ class Task(NamedTuple):
 class TaskRunner:
     """The task runner for pdm project"""
 
-    TYPES = ["cmd", "shell", "call"]
+    TYPES = ["cmd", "shell", "call", "composite"]
     OPTIONS = ["env", "env_file", "help", "site_packages"]
 
     def __init__(self, project: Project) -> None:
@@ -156,9 +169,10 @@ class TaskRunner:
             pass
         return process.returncode
 
-    def _run_task(self, task: Task, args: Sequence[str] = ()) -> int:
+    def _run_task(
+        self, task: Task, args: Sequence[str] = (), opts: TaskOptions | None = None
+    ) -> int:
         kind, _, value, options = task
-        options.pop("help", None)
         shell = False
         if kind == "cmd":
             if not isinstance(value, list):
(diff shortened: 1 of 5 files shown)


pdm/cli/commands/run.py
```diff
@@ -196,10 +212,46 @@ def _run_task(self, task: Task, args: Sequence[str] = ()) -> int:
             err=True,
             verbosity=termui.DETAIL,
         )
+        if kind == "composite":
+            env = options.get("env", {})
+            env_file = options["env_file"]
+            for script in value:
+                if script is None:
```
frostming, line 219: Why does this case exist?

pdm/cli/commands/run.py
```diff
@@ -196,10 +212,46 @@ def _run_task(self, task: Task, args: Sequence[str] = ()) -> int:
...
+                    return 1
+                code = self._run_subtask(script, env, env_file)
```
frostming, line 221: Ideally, the composite task should accept all options and pass them to the subtasks, rather than supporting `env` and `env_file` only.

pdm/cli/commands/run.py
```diff
@@ -196,10 +212,46 @@ def _run_task(self, task: Task, args: Sequence[str] = ()) -> int:
...
+                    return code
+            code = self._run_task(task.clone_with(env, env_file))
+            if code != 0:
+                return code
+            post_task = self._get_task(f"post_{command}")
+            if post_task is not None:
+                code = self._run_task(post_task.clone_with(env, env_file))
+            return code
+        else:
+            return self._run_process(
```
frostming, line 248: Oh, this is another hidden feature -- we also support string-like `cmd` scripts in composite scripts, so why don't we support all(except for `composite` to avoid recursion) if we are doing so?

frostming (changes requested):
Hi, this feature is good to have, but since there is still room for improvement and we are developing on v2.0, new features should go there. I will change the base to `dev` branch if you don't mind.

pdm/cli/commands/run.py
noirbizarre: Removed, part of a previous version

pdm/cli/commands/run.py
noirbizarre: Changed: `_run_subtask` and `run` have been merged and take a `TaskOptions` as optionnal parameter

pdm/cli/commands/run.py
noirbizarre: We now have exact match between tasks and subtask in term of features, including shortcuts because they now use the same function

noirbizarre:
Given it's now planned for 2.0 and we have time to craft a proper solution, I did the following:
- merged `run()` and `_run_subtask()`
- make `run()` and `_run_task()` accept an optional `TaskOptions` parameter
- environment variables are cascading
I think I'm going to add:
- support for passing argument to subtasks
- more tests on environment handling (at least for cascading)

frostming (approved):
Nice work! LGTM

Issue #743, opened 2021-11-18
Towards release 2.0


frostming:
## Planned changes in version 2.0

- [x] ~[PEP 665](https://www.python.org/dev/peps/pep-0665/) support~ Rejected
- [x] Switch the UI framework to [rich](https://github.com/willmcgugan/rich), drop `click`. #1066 #1091 
- [x] Remove support of legacy pyproject.toml format: #1157 
- [x] Drop the dependency of `pip` #1096
- [x] Decouple PDM from any build backend, making it a pure fronend
- [x] #1150 
- [x] #593 
- [x] Restructure the documentation about PEP 621 metadata, tool settings and backend settings
- [ ] (Possible) #46 
  Experiment on sub-resolvers to address the multi specifications for one dependency

~~It won't happen shortly, at least before PEP 665's finalization.~~

See the [Release 2.0 milestone](https://github.com/pdm-project/pdm/milestone/4) for the latest progress.

pawamoy:
Care to elaborate on the reasons that make you want to switch to rich :slightly_smiling_face: ? Isn't it going to be a *big* refactor?

frostming:
> Care to elaborate on the reasons that make you want to switch to Rich 🙂 (I personnally don't like Click but didn't think Rich would offer similar CLI features)? Isn't it going to be a _big_ refactor?

For many reasons:

1. The API to colorize output is annoying when nested, and it always resets all styles when the color ends: `yellow(f"Hello, {red('World')}. Python")` produces `Python` uncolored. As a comparison, rich supports [color themes](https://rich.readthedocs.io/en/latest/style.html#style-themes)
2. PDM uses a patched version of halo to provide parallel spinners because the project is inactive and the [PR is not merged](https://github.com/manrajgrover/halo/pull/155) for a long time.
3. Even so, the parallel spinners could have a race condition in [rare cases](https://github.com/pdm-project/pdm/issues/677)
4. The column display implementation is very simple and the line will break when exceeding the max terminal width.
5. All these are provided by Rich out of the box, strong and well tested, the author of Rich engages full-time OSS so it is well-maintained.
6. The only thing that stops me from adopting rich is it adds a number of dependencies. The size of `pygments` alone is 4MB. If only Rich were splitter into smaller packages.

pawamoy:
I see, thanks! Ah, the constant battle of using libraries while reducing dependencies ^^

harkue:
look forward to this feature:


> New in version 1.12.0
```
[...]

PR #1151, opened 2022-06-22
Proposal: Fix for multilines and composite scripts listing

noirbizarre:
## Pull Request Check List

While the current script listing is working fine when your command are short one-liners, it was broken for multilines scripts on `main` (where this PR was initially implemented).
Now that `dev` switched on a `rich` table display, I'm not sure whether it is improvement or not so I submit it anyway,

This PR improvement proposal was to merge the `Script` and `Description` column into a single `Description` column displaying the `help` if available and if not:
- display the sequence of calls for `composite`
- display the first line of script with a `⮨` ellipsis if it has been truncated

Also, the list is now sorted.

## Example

Here the same output for `main`, `dev` and this PR for this given `tool.pdm.scripts`:

```toml
[tool.pdm.scripts]
all = {composite = ["format", "lint", "cover"]}
test = "pytest better/ tests/"
format = { shell = """\
isort better/
black better/
""" }
doc = "sphinx-build docs docs/_build/html"
"doc:watch" = "sphinx-autobuild docs docs/_build/html"

[tool.pdm.scripts.lint]
shell = """\
  echo "Flakes"
  flake8 better
  echo "isort"
  isort --check better
  echo "black"
  black --check better
  echo "MyPy"
  mypy better
"""

[tool.pdm.scripts.cover]
help = "Run tests with coverage"
cmd = """
pytest
  --cov-report=term
  --cov=better
  --cov-report=html:reports/coverage
  --cov-report=xml:reports/coverage.xml
  --junitxml=reports/tests.xml
"""
```

### Main
Capture d’écran du 2022-06-22 13-21-27

Note: the `composite` task has been removed to be able to run the command

### Dev
Capture d’écran du 2022-06-22 13-22-03

### Proposal
Capture d’écran du 2022-06-22 13-22-53

### DIFF
--- a/pdm/cli/commands/run.py
+++ b/pdm/cli/commands/run.py
@@ -50,6 +50,22 @@ class Task(NamedTuple):
     def __str__(self) -> str:
         return f"<task [cyan]{self.name}[/]>"
 
+    @property
+    def short_description(self) -> str:
+        """
+        A short one line task description
+        """
+        if self.kind == "composite":
+            fallback = f" {termui.Emoji.ARROW_SEPARATOR} ".join(self.args)
+        else:
+            lines = [
+                line.strip() for line in str(self.args).splitlines() if line.strip()
+            ]
+            fallback = (
+                f"{lines[0]}{termui.Emoji.ELLIPSIS}" if len(lines) > 1 else lines[0]
+            )
+        return self.options.get("help", fallback)
+
 
 class TaskRunner:
     """The task runner for pdm project"""
@@ -260,9 +276,9 @@ class TaskRunner:
     def show_list(self) -> None:
         if not self.project.scripts:
             return
-        columns = ["Name", "Type", "Script", "Description"]
+        columns = ["Name", "Type", "Description"]
         result = []
-        for name in self.project.scripts:
+        for name in sorted(self.project.scripts):
             if name == "_":
                 continue
             task = self._get_task(name)
@@ -271,8 +287,7 @@ class TaskRunner:
                 (
                     f"[green]{name}[/]",
                     task.kind,
-                    str(task.args),
-                    task.options.get("help", ""),
+                    task.short_description,
                 )
             )
         self.project.core.ui.display_columns(result, columns)
--- a/news/1151.feature.md
+++ b/news/1151.feature.md
@@ -1,0 +1,1 @@
+Shorter scripts listing, especially for multilines and composite scripts.
--- a/pdm/termui.py
+++ b/pdm/termui.py
@@ -101,11 +101,15 @@ class Emoji:
         LOCK = " "
         CONGRAT = " "
         POPPER = " "
+        ELLIPSIS = "..."
+        ARROW_SEPARATOR = ">"
     else:
         SUCC = ":heavy_check_mark:"
         FAIL = ":heavy_multiplication_x:"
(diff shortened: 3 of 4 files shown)


frostming:
The ⮨ ellipsis doesn't look good on my machine(M1 Mac with WezTerm, Rec Mono). It doesn't even render on Chrome.

Better to choose another character.

noirbizarre:
Ok then let's change that. It was ⮨, here a list of alternatives (so you can test it and validate it is working before i update the PR) I see:
1. The classical ellipsis: `…`
2. Same arrow but thinner: `↵`
3. Same but a bit curvier: `⤶`
4. The fat one: `⮰`

Or if you have any other suggestion or preference.
I can render the porposal with each to choose if you want.

frostming:
I prefer the classical ellipsis and fallback to three dots `...` on Windows cmd. 😞

noirbizarre:
Updated with a classical ellipsis and a fallback on 3 dots for legacy windows terminals.

@frostming it's why I love open-source: you find an awesome tools, open source gives you the opportunity to make more awesome to you. Given I'm a full time developer, and Python packaging have been so painful those last year, I'm glad PDM exists solving my main Python irritation point and I even more glad to contribute to make it more awesome !!! (And BTW, I apologize in advance because I know I will be flooding this repository with pull requests for some time)

frostming:
> (And BTW, I apologize in advance because I know I will be flooding this repository with pull requests for some time)

Oh great, what big would you like to propose? BTW, would you like an alpha release or are you using the dev version already?

pdm/cli/commands/run.py
```diff
@@ -50,6 +50,21 @@ class Task(NamedTuple):
     def __str__(self) -> str:
         return f"<task [cyan]{self.name}[/]>"
 
+    @property
+    def short_description(self) -> str:
+        """
+        A short one line task description
+        """
+        if self.kind == "composite":
+            fallback = " ➤ ".join(self.args)
```
frostming, line 59: A fallback to `>` as well?

frostming (approved):

noirbizarre:
And it's updated 👍🏼 

An alpha release would be great (I can use the dev version but I prefer a released version, easier issue tracking)

Between others, I'd like to submit PRs for:
- yarn-like root script (in progress): make scripts available as root cmd as long as they don't conflict with an existing one (aka. the `test` script should be available as `pdm test`
- a bit of documentation about possible usages with `direnv`, writing plugins...
- some plugins idea: 
  - `pdm-create`: bootstrap projects using templates (like cookie-cutter or copier but builtin, same as `yarn create`, should bring more value than a simple `pdm new = mkdir + cd + pdm init`
  - `pdm-releaser`: like `go-releaser` but for and based on `pdm`
  - `pdm-vscode`: create/update vscode workspace/directory settings automatically
- and contribute on the remaining tasks on the path to 2.0 (edtiables management, XDG dirs...)

I don't know in which order and when, yarn-like root scripts is already started so it might be the next one but after that I think I will help for the 2.0 critical path, plugins can wait

frostming:
>and contribute on the remaining tasks on the path to 2.0 (editables management, XDG dirs...)

I've updated the meta issue #743 about the remaining tasks, FYI edtiable management has been finished.

PR #1159, opened 2022-06-26
feat(CLI): Yarn-like root scripts fallback

noirbizarre:
## Pull Request Check List

This PR allow to execute PDM scripts without the `run` command prefix as long as it doesn't conflict with any builtin command or any plugin-provided command.
In case of error, it produces the exact same output, especially in case of shell command not found (ie. instead of having the command not found error from your shell as `run` does, you still have the PDM usage and error).

In the process I made the "no parameters provided" behavior the same as any others parsing error: usage (help in in this case) is printed on `stderr` instead of `stdout` [here](https://github.com/pdm-project/pdm/pull/1159/files#diff-3a198f24ba7555b1c4cf4d06730a48e6d7494772cda928b78676a76f0f1a0cd9L158-R184).

**Note:** the `click.testing.CliRunner` have some known issues with `stderr` (`stderr` is unbuffered but `CliRunner` mock it as a buffered one making error message lost in the process) and to be able to test error messages, I had to switch on a simpler home made runner only based on Pytest fixture. Same signature, less `click` specific in the process but we gain proper `stderr` handling (as well as the possibility to have the execution output for debug with the `RunResult.print()` method. I kept the dependency in  `pyproject.toml` to avoid locking in this PR and I wasn't sure about side-effects as some tests install `click` and having it as a dependency make sure that it is present in the cache.

**Note:** there has been 3 versions before compatibility with Python 3.7 and Python 3.8 was reached:
- first one was using `argparse`'s `exit_on_error` but this was added in Python 3.9
- second one was trying to backport `argparse`'s `exit_on_error`, was working but MyPy disagreed
- this one which now rely on catching `SystemExit` instead of `argparse.ArgumentError`: less fine but tests ensure that we don't silently catch other errors

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -58,10 +58,8 @@ nav:
       - dev/benchmark.md
 
 markdown_extensions:
-  - admonition
   - pymdownx.highlight:
       linenums: true
-  - pymdownx.superfences
   - pymdownx.tabbed:
       alternate_style: true
   - pymdownx.details
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,14 +7,14 @@ import os
 import shutil
 import sys
 from contextlib import contextmanager
-from io import BytesIO
+from dataclasses import dataclass
+from io import BytesIO, StringIO
 from pathlib import Path
-from typing import Callable, Dict, Iterable, List, Optional, Tuple
+from typing import Callable, Dict, Iterable, List, Mapping, Optional, Tuple
 from urllib.parse import unquote, urlparse
 
 import pytest
 import requests
-from click.testing import CliRunner
 from packaging.version import parse as parse_version
 from unearth.vcs import Git, vcs_support
 
@@ -33,6 +33,7 @@ from pdm.models.requirements import (
 )
 from pdm.models.session import PDMSession
 from pdm.project.config import Config
+from pdm.project.core import Project
 from pdm.utils import normalize_name, path_to_url
 from tests import FIXTURES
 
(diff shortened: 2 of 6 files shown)


frostming:
> This PR allow to execute PDM scripts without the run command prefix as long as it doesn't conflict with any builtin command or any plugin-provided command.

I would prefer to restrict it to user scripts only, so `pdm flask run` shouldn't work.

noirbizarre:
I hesitated on this one.
Updated to restrict to user scripts only ✅

docs/mkdocs.yml
```diff
@@ -58,6 +58,7 @@ nav:
       - dev/benchmark.md
 
 markdown_extensions:
+  - admonition
```
frostming, line 61: Duplicated. It is already included below.

frostming (changes requested):

docs/mkdocs.yml
noirbizarre: Sorry,
I just noticed that `pymdownx.superfences` is also duplicated in the list. I will fix that 👍🏼

tests/conftest.py
```diff
@@ -380,15 +380,67 @@ def is_dev(request):
     return request.param
 
 
-@pytest.fixture()
-def invoke(core):
-    runner = CliRunner(mix_stderr=False)
+class RunResult:
```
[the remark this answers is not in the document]
noirbizarre: Glad to switch to `dataclass` 👍🏼

frostming (approved):
LGTM, also thanks for your detailed technical explanation.

noirbizarre:
You're welcome 🙏🏼 
Do you know why sometimes a python version is missing on `test_integration::test_bsaic_integration` ? Do you think it's something we can fix somehow or is it only a github worker issue ?

frostming:
> You're welcome 🙏🏼 Do you know why sometimes a python version is missing on `test_integration::test_bsaic_integration` ? Do you think it's something we can fix somehow or is it only a github worker issue ?

Yeah, I just don't have the energy to dig it, it seems quite flaky. I doubt it is related to the GitHub Action tools cache.

noirbizarre:
I'll take a look if I have time

frostming:
Besides, it randomly errors
```
unearth: Skip https://pypi.org/simple/pytz/ because of NotHTML: only HTML is   
E       supported but its content-type is ..
```
I don't know why the integration test is so flaky. May need some improvement on the test architecture of integration tests.

PR #1160, opened 2022-06-26
Adds a tox.ini file for easier local testing against all Python versions

noirbizarre:
## Pull Request Check List

Just add a `tox.ini` file for easier local testing against all supported Python versions.
It requires `tox-pdm`, can be either run using `tox` diurectly or through PDM with `pdm run tox` (depends on the new `tox` development dependency group)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,6 +75,20 @@ pdm run test
 
 The test suite is still simple and needs expansion! Please help write more test cases.
 
+!!! note
+    You can also run your test suite against all supported Python version using `tox` with the `tox-pdm` plugin.
+    You can either run it by yourself with:
+ 
+    ```shell
+    tox
+    ```
+ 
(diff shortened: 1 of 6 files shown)


frostming:
I am afraid `command = test` won't work on the latest `tox-pdm`, yes, I removed most of the magic on the new version since it can be very tricky. Just the normal venv and run.

frostming:
> Actually, with [pdm-project/tox-pdm#27](https://github.com/pdm-project/tox-pdm/pull/27) which is included in the `0.5` version it's working perfectly👌🏼 (I tested the Yarn-like root scripts PR with it). I updated the PR with a `>=0.5` lower bound version specifier for `tox-pdm`

Oh yes, I forgot the feature. Can you also update the `pdm.lock`?

noirbizarre:
`pdm.lock` updated ✅

frostming (approved):

PR #1189, opened 2022-06-30
Add option `fallback_verbose`

wikiped:
## Pull Request Check List

- [ ] Test cases added for changed code *.

* - Did not find any test cases checking for output of `global_project.fallback`.

Add config `global_project.fallback_verbose` defaulting to `True`. When set to `False` disables message `Project is not found, fallback to the global project`


docs/docs/configuration.md
```diff
@@ -8,6 +8,7 @@ The following configuration items can be retrieved and modified by [`pdm config`
 | `cache_dir`                   | The root directory of cached files                                        | The default cache location on OS                                          | No                   |                          |
 | `check_update`                | Check if there is any newer version available                             | True                                                                      | No                   |                          |
 | `global_project.fallback`     | Use the global project implicitly if no local project is found            | `False`                                                                   | No                   |                          |
+| `global_project.fallback_verbose`     | If True show message when global project is used implicitly            | `True`                                                                   | No                   |                          |
```
frostming, line 11: I wonder if `silent` is a better name, and the value would need to be reversed.

frostming: I mean just `silent`, without `fallback_`

wikiped: I guess it depends on the future plans and whether you foresee any future additions that might "muddy the waters" or require grouping / splitting options.

In theory `silent` / `quite` / `verbose` could be applied to other options that might (if not now then in the future) emit some use messages and then having specificity (i.e. `fallback_` prefix) helps to distinguish one from another. 

My concern is that `global-project.silent` is not specific enough with regards to what it does right now.

frostming: Mostly agree. There won't be other configurations that need verbosity config in the short future. General verbosity can be controlled by `-v` option. `global_project` is special that it is verbose by default because switching the project may lead to surprising results so it is better to inform users.

PR #1191, opened 2022-06-30
feat: add `--no-keep` option to sync command

frostming:
## Pull Request Check List

### DIFF
--- a/docs/docs/usage/dependency.md
+++ b/docs/docs/usage/dependency.md
@@ -202,7 +202,8 @@ pdm remove -dG test pytest-cov
 There are two similar commands to do this job with a slight difference:
 
 - [`pdm install`](cli_reference.md#exec-0--install) will check the lock file and relock if it mismatches with project file, then install.
-- [`pdm update`](cli_reference.md#exec-0--update) installs dependencies in the lock file and will error out if it doesn't exist. Besides, [`pdm sync`](cli_reference.md#exec-0--sync) can also remove unneeded packages if `--clean` option is given.
+- [`pdm update`](cli_reference.md#exec-0--update) installs dependencies in the lock file and will error out if it doesn't exist.
+- [`pdm sync`](cli_reference.md#exec-0--sync) can remove unneeded packages not included in the lockfile if `--clean` option is given. When `--pure` is given, only selected packages will be kept.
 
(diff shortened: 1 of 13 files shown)


Hnasar:
I found the behavior of `pure` to be a bit hard to understand. A name like `--only` or `--only-keep` would be more intuitive. For example:

```
pdm sync --only-keep --prod
```

Some examples in the docs might help too.

docs/docs/usage/dependency.md
```diff
@@ -202,7 +202,8 @@ pdm remove -dG test pytest-cov
 There are two similar commands to do this job with a slight difference:
 
 - [`pdm install`](cli_reference.md#exec-0--install) will check the lock file and relock if it mismatches with project file, then install.
-- [`pdm update`](cli_reference.md#exec-0--update) installs dependencies in the lock file and will error out if it doesn't exist. Besides, [`pdm sync`](cli_reference.md#exec-0--sync) can also remove unneeded packages if `--clean` option is given.
+- [`pdm update`](cli_reference.md#exec-0--update) installs dependencies in the lock file and will error out if it doesn't exist.
+- [`pdm sync`](cli_reference.md#exec-0--sync) can remove unneeded packages not included in the lockfile if `--clean` option is given. When `--pure` is given, only selected packages will be kept.
```
Hnasar, line 206:
```suggestion
There are a few similar commands to do this job with a slight difference:

- [`pdm sync`](cli_reference.md#exec-0--sync) installs the packages in the lock file.
- [`pdm update`](cli_reference.md#exec-0--update) will update the lock file, then `sync`.
- [`pdm install`](cli_reference.md#exec-0--install) will check the project file for changes, update the lock file if needed, then `sync`.

Note: `sync` has a few options to manage installed packages:
   * `--clean`: remove packages no longer in the lockfile
   * `--only-keep`: only selected packages (using options like `-G` or `--prod`) will be kept.
```

frostming: Thanks, updated

LukeDRussell:
Keen to see this in a pre-release tag. 
I've been trying to migrate a small project to PDM. I've had to delete the venv every time to acheive a strict match to the 'working set' of the intended groups.

PR #1193, opened 2022-07-01
feat: pdm venv integration

frostming:
## Pull Request Check List

1. `pdm venv` commands integration

2. `pdm init` will ask to create virtualenv if not exists
`pdm init`, when choose to create virtualenv:
image
`pdm init`, when choose not to create virtualenv:
image

3. `pdm use` will list pythons in central virtualenv location
4. `pdm install` will create virtualenv if python is not chosen yet and `__pypackages__` doesn't exist
5. venv will be created at `.venv` in project root, while configurable.


thiagola92:
First, i'm loving pdm! Thank you very much!  

Second, why the change to use `venv` by default?  

Observation: Is `False` in `configuration.md`  
https://github.com/pdm-project/pdm/blob/main/docs/docs/usage/configuration.md?plain=1#L20

frostming:
@thiagola92 [here](https://frostming.com/2022/pdm-2-en/#virtualenv-becomes-the-default) are some explanations. And the doc definitely needs to be updated, would you like to contribute?

PR #1203, opened 2022-07-06
feat: make lockfile contain urls instead of filenames

betaboon:
## Pull Request Check List

This PR changes the lockfile to contain full urls instead of just the filenames.

## Context

I'm coming from efforts in the bazel world to properly integrate cross-platform python-lockfiles. (the effort in question is [here](https://github.com/jvolkman/rules_pycross) )
There we stumbled upon the problem that it is hard to resolve filenames in lockfiles to urls and it would be alot easier if the lockfiles would already include the urls, as the tool creating the lockfile should already have the knowledge about the urls.

I hacked together a little [patch](https://gist.github.com/betaboon/39369dd88384ff268ea8f41d2a5e0e98) the does that with pdm and talked to @frostming on discord, who signaled interest in this change.

so here we are ;)


betaboon:
> Hello @betaboon, is there an open issue that discusses this change? If so, could you link to it in the PR description? I'd like to understand the reasoning behind it slightly_smiling_face I think NPM puts the resolved URL in the lock file as well, maybe you got inspired by other package managers?

@pawamoy i added some context-explanation to the description. hope that clears it up a little :)

frostming:
@pawamoy The discussion actually happens in [discord](https://discord.com/channels/824472774965329931/824472777762275361/993105765613183026) and it SGTM as the `filename` is not being used after all by now.

@betaboon Can you bump the minor version at https://github.com/pdm-project/pdm/blob/cc2ad1e2a156f877686f8ed3e2052815a9400037/pdm/project/core.py#L65 ?

frostming:
> > @pawamoy The discussion actually happens in [discord](https://discord.com/channels/824472774965329931/824472777762275361/993105765613183026) and it SGTM as the `filename` is not being used after all by now.
> > @betaboon Can you bump the minor version at
> > https://github.com/pdm-project/pdm/blob/cc2ad1e2a156f877686f8ed3e2052815a9400037/pdm/project/core.py#L65
> > 
> > ?
> 
> done

I am sorry, it seems we should bump the major version, let's make it `"4"`, since it can't be read by old PDM versions. Or if you keep the `filename` it can be `"3.2"`

betaboon:
Which one do you prefer? I can do either.

I think minor (aka adding URL instead of replacing filename) might prevent some migration problems for ppl.

On the other hand since you're aiming for a 2.0 release might be the perfect time to introduce a "breaking" change

frostming:
> Which one do you prefer? I can do either.

At this point it is okay to make a major bump, the PDM can regenerate the lock file automatically if it sees an incompatible lock file version when doing installation.

pawamoy:
@betaboon Thanks!
@frostming oh I see, thanks, I rarely check Discord 🙂

betaboon:
bumped `LOCKFILE_VERSION` to `4.0`

tests/fixtures/projects/demo-package-has-dep-with-extras/pdm.lock
```diff
@@ -49,26 +49,26 @@ summary = "HTTP library with thread-safe connection pooling, file post, and more
 
 [metadata]
 lock_version = "3.1"
```
frostming, line 51: Also please update the lock_version

tests/fixtures/projects/demo-package/pdm.lock
```diff
@@ -32,101 +32,144 @@ dependencies = [
...
-name = "Werkzeug"
+name = "werkzeug"
 version = "1.0.1"
 requires_python = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
 summary = "The comprehensive WSGI web application library."
 
 [metadata]
-lock_version = "3"
-content_hash = "sha256:e7d5fc5ef44897e132127714303b47d189d905df437d07d987c8bf772bb02d1c"
+lock_version = "3.1"
```
frostming, line 47: Ditto

tests/fixtures/projects/demo-package/requirements.txt
```diff
@@ -1,14 +1,22 @@
 click==7.1.2 \
     --hash=sha256:dacca89f4bfadd5de3d7489b7c8a566eee0d3676333fbb50030263894c38c0dc \
+    --hash=sha256:dacca89f4bfadd5de3d7489b7c8a566eee0d3676333fbb50030263894c38c0dc \
```
frostming, line 3: The hash seems duplicating, maybe some bug?

betaboon: i noticed that yesterday to.
this is actually due to: https://github.com/pdm-project/pdm/blob/main/pdm/project/config.py#L36-L39

since there are two default-repositories, there are two hashes for each file (one for each repository).

the way to prevent duplicate hashes would be using a `set` [here](https://github.com/pdm-project/pdm/blob/main/pdm/formats/requirements.py#L190). but that would change the order of hashes.

tests/fixtures/projects/demo-package/pdm.lock
betaboon: done

tests/fixtures/projects/demo-package/requirements.txt
frostming: > this is actually due to: https://github.com/pdm-project/pdm/blob/main/pdm/project/config.py#L36-L39

I doubt that, though a bit confusing, the default repositories are for **uploading packages**, and have nothing to do with downloading and fetching

betaboon: if you take a look at the generated [`pdm.lock`](https://github.com/pdm-project/pdm/blob/aa272a97957356155523148e2cbc138e05322bdb/tests/fixtures/projects/demo-package/pdm.lock#L52-L55) it also contains files from both repositories

frostming: Should we prefer the link from the prior source(i.e. perform deduplication against filename, the first comes wins)?
The extra sources act as complementary sources. Only those not found in the prior sources take the link from these sources.

betaboon: ah yeah you're right. it's due to `pyproject.toml` in the example projects.

i think we should only do deduplication of hashes in the `requirements.txt`
i see value in having multiple urls for the same package in the `pdm.lock` as it allows fallback. imagine the case where the first source is available at lock-time, not not available at install time.

frostming: >imagine the case where the first source is available at lock-time, not not available at install time.

Sounds sensible, then we will implement the behavior I mentioned in the gist:

> When installing from the lock file, take the links stored with the lock file.

frostming: First, of course
Second, of course

betaboon: done

frostming (approved):
LGTM, can you please add a release note under `news` folder?

PR #1210, opened 2022-07-08
pre-commit: Add flake8-return

cclauss:
## Pull Request Check List

* https://pypi.org/project/flake8-return

I am unsure which of these you want to fix vs. ignore.


frostming:
IMO the functionality the plugin provides a relatively trivial and brings minor benefit. There are many flake8 plugins and we just can't adopt all of them.

Sorry I am going to close this PR.

PR #1225, opened 2022-07-15
fix: write lockfile before post_lock hook

### LINKED ISSUE #1224: Write lockfile before calling the "post_lock" hook
First, thanks for the nice work with PDM. 😄 

I've been trying to come up with a way to `pdm export` a requirements.txt file every time an update to the lockfile happens. As some vulnerability scanning services and other CI tools do not yet support parsing/reading/interpreting pdm.lock files, this would allow us to adopt the PDM without compromising other parts of the workflow.

At first, we thought using PDM lifecycle hooks functionality would suffice. Something like
```toml
[tool.pdm.scripts]
post_lock = {shell = "pdm export -f requirements > requirements.txt"}
```
Unfortunately, that did not work, as the `post_lock` hook is called **before** the lockfile is written to the disk (snippet below)
https://github.com/pdm-project/pdm/blob/7d9c0f3d108c7418cd93e6a020b0f56333b8936b/pdm/cli/actions.py#L120-L125

Having the lockfile written to the disk before calling the `post_lock` hook would enable this functionality and make PDM's adoption easier for workflows that rely, at some point, on the standard requirements.txt file. 😃

lecardozo:
## Pull Request Check List

Now lockfile is written before `post_lock` hook. Fixes #1224

### DIFF
--- a/news/1224.bugfix.md
+++ b/news/1224.bugfix.md
@@ -1,0 +1,1 @@
+Write lockfile before calling 'post_lock' hook
--- a/pdm/cli/actions.py
+++ b/pdm/cli/actions.py
@@ -120,9 +120,8 @@ def do_lock(
         else:
             data = format_lockfile(project, mapping, dependencies)
             ui.echo(f"{termui.Emoji.LOCK} Lock successful")
+            project.write_lockfile(data, write=not dry_run)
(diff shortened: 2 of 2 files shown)


frostming:
thank you :

PR #1267, opened 2022-07-23
feat: Introduce `pypi.ca_certs` config entry

### LINKED ISSUE #1240: Feature Request: Make CA bundle configurable
## Problem description

- Some users (e.g. me) are using package repositories other than pypi.org.
- These are (of course) accessed over an encrypted HTTPS connection.
- The repository server's certificate might be signed by a private CA.
- In this case, PDM will fail to access the repository via HTTPS.

## There is a known workaround

As the requests are being made using the [requests](https://pypi.org/project/requests/) library, I can [override the cert store using `REQUESTS_CA_BUNDLE`](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification) so that my system's CA store is used instead of [certifi](https://pypi.org/project/certifi/). E.g. for Debian:

```bash
export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
```

## The solution I would prefer

I'm seeing some problems with the workaround above:

 - As a user, I might not want to have `REQUESTS_CA_BUNDLE` set for every terminal session. In that case, I have to remember to set it manually when using PDM.
 - *If I would want to set the env variable globally*, I still have to know how to do that properly.

So, I would prefer to be able to set the CA bundle in PDM's configuration, similar to `pip`'s `global.cert` option, e.g. `pypi.cert` or `pypi.ca`. This could be set to a file path (e.g. `/etc/ssl/certs/ca-certificates.crt` for Debian).

sleiner:
## Pull Request Check List

Fix #1240.


sleiner:
depends on frostming/unearth#16

sleiner:
@frostming Do you have a suggestion for a test case?

frostming:
> @frostming Do you have a suggestion for a test case?

It is okay to not include a test, we just reuse the ability of `requests`, which should have covered :)

pdm/project/config.py
```diff
@@ -199,6 +200,11 @@ class Config(MutableMapping[str, str]):
         "pypi.verify_ssl": ConfigItem(
             "Verify SSL certificate when query PyPI", True, coerce=ensure_boolean
         ),
+        "pypi.ca_certs": ConfigItem(
+            "Path to a CA certificate bundle used for verifying the identity "
+            "of the PyPI server",
+            certifi.where(),
```
frostming, line 206: This default value will show in the output of `pdm config`, which will be a bit noisy IMO. I wonder if it will be better if we just use None here and get the default path when used.

sleiner: > This default value will show in the output of `pdm config`, which will be a bit noisy IMO.

I agree 👍🏻

---

>. I wonder if it will be better if we just use None here and get the default path when used.

I have updated the code to do just that. I'm wondering though whether we should introduce something like `hide_default` into `ConfigItem` since especially in this case, `config.py` seems to be the right place for specifying the default (at least to me). What do you think?

frostming: > whether we should introduce something like hide_default into ConfigItem since especially in this case

Good suggestion, but I think it is okay for now since `ca_certs` is the only one. And implementing such a behavior needs a bit refactor around the logic of retrieving the config value(you need to distinguish whether it comes from default or input). So thanks for your work, it LGTM.

sleiner: Perfect, I will rebase it then 👍🏻

pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "virtualenv>=20",
     "pep517>=0.11.0",
     "requests-toolbelt",
-    "unearth>=0.5.2,<1",
+    "unearth @ git+https://github.com/frostming/unearth.git@main",
```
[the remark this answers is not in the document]
frostming: Can you help update the version?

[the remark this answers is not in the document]
sleiner: done 👍🏻

PR #1292, opened 2022-08-01
feat: Add config for client cert and key for PyPI

### LINKED ISSUE #1283: pdm completion breaks after upgrade to 2.0

## Steps to reproduce

run `pdm completion zsh` or `pdm completion powershell`. The output breaks coz of changed to rich.

## Actual behavior

`[word]` will be parsed as rich markup which is not expected. zsh and powershell completion scripts will be broken.

## Expected behavior

should work like pdm 1.x only `%{python_executable}` is replaced.

## Environment Information

```bash
PDM version:
  2.0.2
Python Interpreter:
  D:\Programs\Python\Python38\python.exe (3.8)
Project Root:
  D:/Projects/ykdlweb
Project Packages:
  D:\Projects\ykdlweb\__pypackages__\3.8
{
  "implementation_version": "3.8.10",
  "os_name": "nt",
  "platform_machine": "AMD64",
  "platform_release": "10",
  "platform_system": "Windows",
  "platform_version": "10.0.19044",
  "python_full_version": "3.8.10",
  "python_version": "3.8",
  "sys_platform": "win32"
}
```
---

I have digged into the source files. Didn't find an elegant solution for a pr. `print` maybe naive :)

Maybe add and `escape=False` parameter to `UI.echo`

GooseYArd:
Support mutual TLS

## Pull Request Check List

Adds a pair of configuration directives, pypi.client_cert and pypi.client_key, which are applied to the session objects cert attribute to enable mutual TLS to private repositories.


pdm/models/environment.py
```diff
@@ -143,6 +143,10 @@ def get_finder(
                 self.project.config.get("pypi.ca_certs", certifi.where())
             ),
         )
+        if certfn := self.project.config.get("pypi.client_cert"):
```
frostming, line 146: Walrus operator is py38 and above, which is not allowed according to the `requires-python` range.

frostming (changes requested):

pdm/models/environment.py
GooseYArd: whoops, got it!

PR #1299, opened 2022-08-03
Add an override option to the env_file script option

pgjones:
This allows the user to specify that the env_file should override any
existing environment variables. This is not the default as the
environment the code runs it should take precedence.

## Pull Request Check List

### DIFF
--- a/pdm/cli/commands/run.py
+++ b/pdm/cli/commands/run.py
@@ -21,9 +21,13 @@ from pdm.project import Project
 from pdm.utils import is_path_relative_to
 
 
+class EnvFileOptions(TypedDict, total=True):
+    override: str
+
+
 class TaskOptions(TypedDict, total=False):
     env: Mapping[str, str]
-    env_file: str | None
+    env_file: EnvFileOptions | str | None
     help: str
     site_packages: bool
 
@@ -119,12 +123,19 @@ class TaskRunner:
         shell: bool = False,
         site_packages: bool = False,
         env: Mapping[str, str] | None = None,
-        env_file: str | None = None,
+        env_file: EnvFileOptions | str | None = None,
     ) -> int:
         """Run command in a subprocess and return the exit code."""
         project = self.project
-        process_env = {}
-        if env_file:
+        process_env = os.environ.copy()
+        if env_file is not None:
+            if isinstance(env_file, str):
+                path = env_file
+                override = False
+            else:
+                path = env_file["override"]
+                override = True
+
             import dotenv
 
             project.core.ui.echo(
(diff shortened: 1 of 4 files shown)


pgjones:
The test failures seem unrelated.

pdm/cli/commands/run.py
```diff
@@ -21,9 +21,14 @@
 from pdm.utils import is_path_relative_to
 
 
+class EnvFileOptions(TypedDict, total=True):
+    path: str
+    override: bool | None
```
frostming, line 26: I recommend:

```python
class EnvFileOverride(TypedDict, total=True):
    override: str
```
This means when user uses a table, `env_file.override = ".env"` , then the env will override. If he doesn't want to override, he will use a plain text `env_file = ".env"`

pgjones: I can't get this to work (I might be missing something), the following errors with `env_file` key already defined:
```
{ cmd = "...", env_file = ".env", env_file.override = true}
```

frostming: Not overriding:
```toml
{ cmd = "...", env_file = ".env"}
```
Overriding:
```toml
{ cmd = "...", env_file.override = ".env"}
```

frostming: You are stuck in with a boolean `override`, haha. The above however uses different structures to decide whether to override.

pgjones: I see, I've updated.

pdm/cli/commands/run.py
```diff
@@ -119,23 +123,31 @@ def _run_process(
...
                 err=True,
                 verbosity=termui.Verbosity.DETAIL,
             )
-            process_env = dotenv.dotenv_values(
-                project.root / env_file, encoding="utf-8"
-            )
-        process_env.update(os.environ)
+            dotenv_env = dotenv.dotenv_values(project.root / path, encoding="utf-8")
+            if override:
+                process_env = process_env | dotenv_env
```
frostming, line 148: This syntax is Python 3.9+, please use `{**this, **other}` instead.

frostming (changes requested):
Just a minor issue. Could you also add some docs in `docs/usage/script.md`?

pdm/cli/commands/run.py
pgjones: Done.

pgjones:
I've added docs as well.

frostming (approved):

pgjones:
Thanks

choinhet:
image
this is not working for me :(
neither if I do `start.env_file` for example

PR #1313, opened 2022-08-10
fix: Work on FIPS 140-2 Enabled Systems

linville:
## Pull Request Check List

This PR fixes pdm to work on FIPS 140-2 enabled systems using Python 3.9+.

The change adds the `usedforsecurity=False` kwarg to the usage of `hashlib.md5(..)` in `cli/commands/venv/utils.py`. This does not appear to look like security related code (but someone else more familiar with this area should probably chime-in on this).

The `usedforsecurity` kwarg was added in Python 3.9, to maintain compatibility with Python 3.7 this uses [`hashlib.new` generic constructor](https://docs.python.org/3/library/hashlib.html#hashlib.new). **Note:** pdm will continue to fail using Python 3.7 and 3.8 on FIPS 140-2 enabled systems (swapping out for another hash supported by older Pythons like sha256 will likely work, but I'm unsure all what expects the existing hash to be what it is).

## Background / FIPS Restrictions

On FIPS 140-2 enabled systems, running a command like `pdm build` will lead to an error:

```sh
$ pdm build 
Building sdist... 
See /tmp/pdm-build-rp3wfnc5.log for detailed debug log. 
[ValueError]:  disabled for FIPS 
```


frostming:
I think it is better to switch to another hashing algorithm, a hash string of `{algo}:{hash_value}` can be understood by older versions of PDM.

linville:
I just realized this is an issue that arose when the venv capability was included directly in pdm 2.0.0 and the default behavior was switched to use venvs rather than pep582. Setting the default back to pep582 (`python.use_venv false`) also works around the issue (and avoiding any `pdm venv` commands). (I'm new to pdm and this wasn't quite clear to me until now; this is more of a note to anyone else new to pdm running into this issue).

frostming (approved):
Thanks, we might need to switch to other hash algorithms in the future.

Issue #1364, opened 2022-09-07
Documentation page on what to commit to version control


gjoseph92:
## Is your feature request related to a problem? Please describe.

For beginners, a clear section instructing them what to track (and not track) in version control might be helpful. For comparison, poetry has this: https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control.

## Describe the solution you'd like

A "Working with version control" section under https://pdm.fming.dev/latest/usage/project/ recommending that you:
* Always commit `pyproject.toml`
* Almost always commit `pdm.lock` so other devs can contribute to the project in an identical environment, so you can re-use that environment in CI, etc.
* Add `.pdm.toml` to your `.gitignore` and don't commit it (I'm not actually sure about this—is it guaranteed to only contain machine-specific config?)

Beartime234:
Suggested section in PR above. Umm'd and Ahh'd about the wording so feel free to critque/change.
#1373

PR #1373, opened 2022-09-08
doc: Add working with version control section

Beartime234:
## Pull Request Check List

~- [x] Test cases added for changed code.~

Added a section under project in the documentation on how to correctly use a version control system like git with a PDM based project. This resolves issue #1364 

_Sorry first time committing to with open source so sorry if there are any mistakes._

### DIFF
--- a/docs/docs/usage/project.md
+++ b/docs/docs/usage/project.md
@@ -247,6 +247,20 @@ pdm export -f setuppy -o setup.py
 !!! NOTE
     You can also run `pdm export` with a [`.pre-commit` hook](advanced.md#hooks-for-pre-commit).
 
+## Working with version control
+
+Ensure the `pdm.lock` and `pyproject.toml` files are committed when using a version control system with PDM. Do not commit the `.pdm.toml` file.
+
+By committing the `pdm.lock` file, you ensure that all installers are using the same versions of dependencies.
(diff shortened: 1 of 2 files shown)


docs/docs/usage/project.md
```diff
@@ -247,6 +247,20 @@ pdm export -f setuppy -o setup.py
 !!! NOTE
     You can also run `pdm export` with a [`.pre-commit` hook](advanced.md#hooks-for-pre-commit).
 
+## Working with version control
+
+Ensure the `pdm.lock` and `pyproject.toml` files are committed when using a version control system with PDM. Do not commit the `.pdm.toml` file.
```
frostming, line 252: `pyprojec.toml` is a MUST, while `pdm.lock` is a SHOULD. There are practices you don't commit the `pdm.lock` file, especially for a library. `.pdm.toml` is a MUST NOT.

frostming (commented):
Nice done! No big deals, only a small recommendation.

docs/docs/usage/project.md
Beartime234: Awesome, Ya that makes sense was not aware of a scenario where you would not want to commit the lock file. 

Have updated accordingly. Had to do some rewording for it to make sense but lemme know if any more concerns.

frostming (approved):

PR #1395, opened 2022-09-22
feat: Introduce `ca_certs` / `--ca-certs` to `pdm publish` repos

### LINKED ISSUE #1392: Feature Request: Make publish repository CA bundle configurable
## Problem Statement

In the [feature adding CA bundle configuration](https://github.com/pdm-project/pdm/pull/1267), the configuration was added for dependency resolution; however, when attempting to publish built packages to a custom PyPI repository (e.g. a local Artifactory PyPI repository), with a self-signed certificate, the upload fails with the following:

```bash
[SSLError]: HTTPSConnectionPool(host='custom-artifactory', port=443): Max retries exceeded with url: /artifactory/api/pypi/pypi-local (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate 
verify failed: self signed certificate in certificate chain (_ssl.c:1129)')))
```

This happens even when using:

`pdm config pypi.ca_certs /path/to/custom/ca.pem`

## There is a known workaround

Just as in the [feature request for making CA bundles configurable](https://github.com/pdm-project/pdm/issues/1240), the underlying [REQUESTS_CA_BUNDLE environment variable](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification) may be set to point to the correct bundle. For example, when using a Unix shell like bash:

```bash
export REQUESTS_CA_BUNDLE=/path/to/custom/ca.pem
```

## The solution I would prefer

- The primary problem with using `REQUESTS_CA_BUNDLE` (outside of needing to set it in new shell windows) is that it affects *both* dependency resolution as well as publishing.
   - Therefore, if using, for example, the standard PyPI for resolution, but a custom upload location, the CA bundle used must be the standard [certifi](https://github.com/certifi/python-certifi) bundle with the custom bundle tacked on.
- For publishing, it seems that the best option would be to:
   1. Add a `ca_bundle` configuration option for a publish repository in the global config (this matches the naming of `pypi.ca_bundle`)
   3. Add a `--ca-bundle` option to the `pdm publish` sub-command to allow overriding from the command line
   4. Allow the `--ca-bundle` override to also be provided via the `PDM_PUBLISH_CA_BUNDLE` environment variable

**Note**: if the proposed solution is acceptable, I would be happy to contribute the new feature.

crbeckle:
## Pull Request Check List

Fix #1392.

### DIFF
--- a/docs/docs/usage/project.md
+++ b/docs/docs/usage/project.md
@@ -134,7 +134,7 @@ password = "<secret>"
 ca_certs = "/path/to/custom-cacerts.pem"
 ```
 
-A PEM-enoded certificate authority cert bundle (`ca_certs`) can be used for local / custom PyPI repositories where the server certificate is not signed by the standard [certifi](https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem) CA bundle.
+A PEM-encoded Certificate Authority bundle (`ca_certs`) can be used for local / custom PyPI repositories where the server certificate is not signed by the standard [certifi](https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem) CA bundle.
 
 !!! NOTE
     You don't need to configure the `url` for `pypi` and `testpypi` repositories, they are filled by default values.
--- a/docs/docs/usage/project.md
+++ b/docs/docs/usage/project.md
@@ -131,10 +131,16 @@ password = "<secret>"
 url = "https://pypi.company.org/legacy/"
 username = "frostming"
 password = "<secret>"
+ca_certs = "/path/to/custom-cacerts.pem"
 ```
+
+A PEM-enoded certificate authority cert bundle (`ca_certs`) can be used for local / custom PyPI repositories where the server certificate is not signed by the standard [certifi](https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem) CA bundle.
 
 !!! NOTE
     You don't need to configure the `url` for `pypi` and `testpypi` repositories, they are filled by default values.
+
+!!! TIP
+    The username, password, and certificate authority bundle can be passed in from the command line for `pdm publish` via `--username`, `--password`, and `--ca-certs`, respectively.
 
 To change the repository config from the command line, use the [`pdm config`](cli_reference.md#exec-0--config) command:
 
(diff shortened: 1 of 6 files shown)


docs/docs/usage/project.md
```diff
@@ -131,18 +131,25 @@ password = "<secret>"
 url = "https://pypi.company.org/legacy/"
 username = "frostming"
 password = "<secret>"
+ca_certs = "/path/to/custom-cacerts.pem"
 ```
 
+A PEM-enoded certificate authority cert bundle (`ca_certs`) can be used for local / custom PyPI repositories where the server certificate is not signed by the standard [certifi](https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem) CA bundle.
```
pawamoy, line 137:
```suggestion
A PEM-encoded Certificate Authority bundle (`ca_certs`) can be used for local / custom PyPI repositories where the server certificate is not signed by the standard [certifi](https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem) CA bundle.
```

src/pdm/cli/commands/publish/__init__.py
```diff
@@ -68,6 +68,12 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
             dest="build",
             help="Don't build the package before publishing",
         )
+        parser.add_argument(
+            "--ca-certs",
+            dest="ca_certs",
+            help="The path to a PEM-encoded certificate authority cert bundle to use"
```
pawamoy, line 74:
```suggestion
            help="The path to a PEM-encoded Certificate Authority bundle to use"
```

crbeckle:
Thanks for the review, @pawamoy! I updated the two fixes noted in a new commit. [One of the checks failed on the initial commits](https://github.com/pdm-project/pdm/actions/runs/3107438474/jobs/5035548860), but it doesn't appear to have anything to do with my changes. This seems to be the actual error, repeated twice:

```
pdm.termui:base.py:59 Using pip 22.2.2 from /Users/runner/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/pip (python 3.8)
pdm.termui:base.py:59 WARNING: Skipping page https://pypi.org/simple/pdm-pep517/ because the GET request got Content-Type: Unknown. The only supported Content-Types are application/vnd.pypi.simple.v1+json, application/vnd.pypi.simple.v1+html, and text/html
pdm.termui:base.py:59 ERROR: Could not find a version that satisfies the requirement pdm-pep517 (from versions: none)
pdm.termui:base.py:59 ERROR: No matching distribution found for pdm-pep517
```

Seems like a fluke or issue with either pip or the upload of pdm-pep517? Is that something that has been breaking in past workflows?

frostming (approved):

frostming:
LGTM, thanks for working on this.

PR #1419, opened 2022-10-05
fix: switch to the public api of importlib.metadata

frostming:
## Pull Request Check List


frostming:
This doesn't work

PR #1433, opened 2022-10-13
Draft licenses CLI feature

JohnHardy:
## Pull Request Check List

Added a new CLI command `licenses` that allows you to output the license documentation for all the deps.  This usually requires 3rd party projects that may or may not be compatible with whatever packaging system is used.  This CLI command implements common use cases that I have seen in the wild.

🙏🙏 **If you are interested in merging this PR in, please let me know so I will invest the time to lint / write test cases.  However, before that, I wanted to make sure that you were happy with: (1) the existence of the feature, (2) the implementation details, my first time working on a package manager, (3) the ux.** 🙏🙏 

Features are:

* Licenses can be gathered from the working set or resolved using the resolver/lockfile.  
* Output to CSV
* Output to JSON
* Output to console in the fancy PDM tables
* Output as markdown (including license text files included in dist-info)
* If a `License` field is empty, it will try to recover it from the Trove classifiers.
* Highlight the group a requirement is defined in (eg. dev deps) 
* Distinguish between specified deps and sub-dependencies 

Here are some use cases:

`pdm licenses`
image

`pdm licenses --resolve --skip default,tox,workflow,sub --fields group,identifier,version,licenses --sort licenses`
image

`pdm licenses --sort licenses`
image

`pdm licenses --sort licenses --csv`
image

`pdm licenses --sort licenses --markdown > mylicenses.md`
image


JohnHardy:
@frostming sounds good to me, but it will impact other functions inside pdm like `show_dependency_graph` , `actions.py`, etc.  I was reluctant to pollute `do_list` since I don't know the codebase.  What do you prefer?

Shall I keep the switches for selecting fields, skipping dep groups, changing the sort column, etc?

frostming:
> @frostming sounds good to me, but it will impact other functions inside pdm like `show_dependency_graph` , `actions.py`, etc. I was reluctant to pollute `do_list` since I don't know the codebase. What do you prefer?
> 
> Shall I keep the switches for selecting fields, skipping dep groups, changing the sort column, etc?
> 
> EDIT: It could also live in a plugin - happy to follow your guidance and suggestions from others.

The column selection is great, let's revisit the current options and the options to be added, if it is to be integrated in to `pdm list`:

| Format/Option | `--fields` | `--sort` | `--resolve` | `--license` | `--reverse` | `--include/--exclude`[^1] |
| ----- | ---- | ---- | ---- | ---- |  ---- | --- |
| table | ✅ | ✅ | ✅ | ✅ | ⛔ | ✅ | 
| csv| ✅ | ✅ | ✅ | ✅ | ⛔ | ✅ | 
| json | ✅ | ✅ | ✅ | ✅ | ⛔ | ✅ | 
| freeze | ⛔ | ⛔ | ⛔[^2] | ⛔ | ⛔ | ✅ | 
| graph[^3] | ⛔ | ⛔ | ⛔ | ⛔ | ✅ | ✅ | 

[^1]: To replace the proposed `--skip`, this should only work with `--resolve`, and `include` is prior to `exclude`
[^2]: should use `pdm export --format requirements` instead
[^3]: Can be given with `--json`

I hope that doesn't make the feature too complex.

JohnHardy:
I had another quick go at this again to match your feature matrix, let me know what you think!

Let me know if I missed anything or you would like something to be done differently.

**Questions:**

1. Currently using `--include *` to mean include everything.  Do you want it to match `pdm install -G:all` style instead?
2. Currently using `--exclude sub` to mean exclude sub-dependencies, but `sub` seems like an unclear choice of word.  Any suggestions?
3. At the moment there is the case where `cachecontrol` and `cachecontrol[filecache]` reference the `cachecontrol` package (seems right).  Is there a case where they should be distinguished by name?  Is there a way of mapping distributions back to requirement identifiers.  
4. At the moment, the resolver works on ALL groups BEFORE being filtered and listed.  However, we might want to send less packages to the resolver.  Only I thought that could potentially resolve different versions / licenses.  Probably nobody will care, but just wondering.

If you are happy, I can make some test cases.


PDM list

frostming:
> 1. Currently using `--include *` to mean include everything.  Do you want it to match `pdm install -G:all` style instead?

The current implementation LGTM.


> 2. Currently using `--exclude sub` to mean exclude sub-dependencies, but `sub` seems like an unclear choice of word.  Any suggestions?

Two options, 1) use a special identifier to avoid conflicting with existing groups, such as `:sub` 2) use another option flag `--exclude-sub`



> 3. At the moment there is the case where `cachecontrol` and `cachecontrol[filecache]` reference the `cachecontrol` package (seems right).

IMO it is okay to ignore the entries with extras `[]`

JohnHardy:
1. 👍
2. Done, `:sub` is used.
3. Sounds good.  FYI: I noticed that the `echo` function will trim `word[here]` to `word` even if the string does not contain `[/]` to close it.
4. I am updating the tests now.  Please can you advise how to mock the Command functions?

For example:

```
def test_list_command(project, invoke, mocker):  # previous test
    do_list = mocker.patch.object(actions, "do_list")
    invoke(["list"], obj=project)
    do_list.assert_called_once()   # <--- works OK in old codebase
    
def test_list_command(project, invoke, mocker):  # proposed new test
    handle = mocker.patch.object(Command, "handle")
    invoke(["list"], obj=project)
    handle.assert_called_once()   # <--- list.Command.handle is not called
```

frostming:
> 3\. FYI: I noticed that the `echo` function will trim `word[here]` to `word` even if the string does not contain `[/]` to close it.

Yeah, that is the markup syntax of rich console, `[]` need to be escaped by `\`

JohnHardy:
@frostming I've added a bunch of tests and sorted the linting.

Can you advise on how I can test `--resolve`  with the mock test fixtures?

If it is not necessary, then I think it is ready for review.

tests/cli/test_list.py
```diff
@@ -1,96 +1,147 @@
...
+from tests import FIXTURES
+from tests.conftest import Distribution
 
 
 def test_list_command(project, invoke, mocker):
-    do_list = mocker.patch.object(actions, "do_list")
-    invoke(["list"], obj=project)
-    do_list.assert_called_once()
+    # Calls the correct handler within the Command
+    with patch.object(Command, "handle_list") as m:
```
frostming, line 17:
```suggestion
    with mocker.patch.object(Command, "handle_list") as m:
```

tests/cli/test_list.py
```diff
@@ -1,96 +1,147 @@
...
+        invoke(["list"], obj=project)
+        m.assert_called_once()
 
 
 @pytest.mark.usefixtures("repository", "working_set")
-def test_list_dependency_graph(project, capsys):
+def test_list_dependency_graph(project, invoke):
+    # Calls the correct handler within the list command
+    with patch.object(Command, "handle_graph") as m:
```
frostming, line 25: ditto

frostming:
> Can you advise on how I can test `--resolve` with the mock test fixtures?

Use the `repository` fixture to mock the package source and `working_set` to mock the installation. Then run `pdm add requests` to generate a lock file and then it should be able to generate a package list by `--resolve`.

All packages in the mocked source are defined in `tests/fixtures/pypi.json` and feel free to add more packages with dependencies if necessary.

tests/cli/test_list.py
JohnHardy: I can't get the mocker fixture to work at all, not sure why yet!  Will take another look tomorrow.

tests/cli/test_others.py
```diff
@@ -63,12 +62,13 @@ def test_global_project_other_location(invoke, project):
...
-    result = invoke(["list"])
+    mocker.patch.object(actions, "do_lock", side_effect=RuntimeError("test error"))
+    result = invoke(["lock"])
     assert "[RuntimeError]: test error" in result.stderr
 
-    result = invoke(["list", "-v"])
-    assert isinstance(result.exception, RuntimeError)
+    with pytest.raises(RuntimeError) as err:
+        result = invoke(["lock"], strict=True)
+    assert "test error" in str(err)
```
frostming, line 71: This is to test the behavior of `-v` to raise the error directly, why not keep the old way.

```python
assert isinstance(result.exception, RuntimeError)
```

tests/test_integration.py
```diff
@@ -51,5 +51,5 @@ def test_basic_integration(python_version, core, tmp_path, invoke):
 def test_actual_list_freeze(project, local_finder, invoke):
     invoke(["config", "-l", "install.parallel", "false"], obj=project, strict=True)
     invoke(["add", "first"], obj=project, strict=True)
-    r = invoke(["list", "--freeze"], obj=project)
+    r = invoke(["export", "--format", "requirements"], obj=project)
```
frostming, line 54: This doesn't seem correct to me, they don't work the same way.

src/pdm/cli/commands/list.py
```diff
@@ -1,37 +1,369 @@
...
+SUBDEP_GROUP_LABEL = ":sub"
+
 
 class Command(BaseCommand):
     """List packages installed in the current working set"""
 
     def add_arguments(self, parser: argparse.ArgumentParser) -> None:
-        group = parser.add_mutually_exclusive_group()
-        group.add_argument(
+        graph = parser.add_mutually_exclusive_group()
```
frostming, line 26: You seem to leave the `--freeze` option out, that isn't what I mean in the matrix. `--freeze` is supposed work without any other options. `pdm list --freeze` prints the **currently installed** requirements list compatible with `pip`'s requirements file format.

tests/cli/test_list.py
```diff
@@ -276,4 +331,398 @@ def test_list_json_with_circular_reverse(project, capsys, repository):
...
+    assert expected == result.output
+
+
+# # TODO: resolve with graph?
+# # TODO: how to fix this?
+# # "[CandidateNotFound]: No candidate is found for `requests` that
+# # matches the environment or hashes\nAdd '-v' to see the detailed traceback\n"
+# @pytest.mark.usefixtures("working_set", "repository")
+# def test_list_bare_sorted_version_resolve(project, invoke, repository):
+#     # project.environment.python_requires = PySpecSet(">=3.6")
```
frostming, line 497: Shouldn't this line fix the `CandidateNotFound` issue?

frostming (changes requested):

tests/cli/test_list.py
JohnHardy: Sorted 👍

tests/cli/test_list.py
JohnHardy: Sorted 👍

src/pdm/cli/commands/list.py
JohnHardy: Ah sorry, I thought you meant to remove it.  I will add it back 👍

tests/cli/test_list.py
JohnHardy: I thought so, but I cannot make it work.  Do you have any ideas?

tests/test_integration.py
JohnHardy: Sorted 👍 will push asap

src/pdm/cli/commands/list.py
JohnHardy: Sorted 👍 will push asap

tests/cli/test_others.py
JohnHardy: Ah ok, I think I had a problem locally.  I reverted and changed from `do_list` to `do_lock`.   Is that still ok?

tests/conftest.py
```diff
@@ -189,11 +189,13 @@ def build_env():
...
 class Distribution:
-    def __init__(self, key, version, editable=False):
+    def __init__(self, key, version, editable=False, metadata=None):
         self.version = version
         self.link_file = "editable" if editable else None
         self.dependencies = []
-        self.metadata = {"Name": key}
+        self.metadata = {"Name": key, "Version": version}
+        if metadata:
+            self.metadata = {**self.metadata, **metadata}
```
JohnHardy, line 198: Would you prefer `name` and `version` each to be a `@property` reading from `self.metadata`?

tests/cli/test_list.py
JohnHardy: I tried quite a lot of different things here, and I am not sure what to try next.  I would appreciate any advice :)

tests/cli/test_list.py
```diff
@@ -1,96 +1,161 @@
...
+    )
+    expects = "".join(expects)
+    assert expects == result.outputs
+
+    # Only include the dev dep
+    result = invoke(["list", "--graph", "--include", "dev"], obj=project)
+    expects = (
+        "demo 0.0.1 [ Not required ]\n",
+        "+-- chardet [ not installed ] [ required: Any ]\n",  # installed 3.0.4 ?
+        "`-- idna [ not installed ] [ required: Any ]\n",  # installed 2.7 ?
```
JohnHardy, line 74: Here the `chardet` and `idna` packages are showing as `[ not installed ]`.   I am not sure why this is showing like this, because `--resolve` is not used, so these packages should all be from the working set.

JohnHardy:
Added a few of comments on the code where I need some help/advice about what the correct behaviour should be.  Thanks!

tests/cli/test_list.py
frostming: You need to mock the `.metadata` cuz it will be accessed during the command, which will try to find a package link and fail.
https://github.com/pdm-project/pdm/blob/cd2f1a98513d03f03fcb784804b9eff10fb37871/src/pdm/cli/commands/list.py#L144-L146

tests/cli/test_list.py
```diff
@@ -1,96 +1,161 @@
...
+        "demo 0.0.1 [ Not required ]\n",
+        "+-- chardet 3.0.4 [ required: Any ]\n",
+        "`-- idna 2.7 [ required: Any ]\n",
+        "requests 2.19.1 [ Not required ]\n",
+        "+-- certifi 2018.11.17 [ required: >=2017.4.17 ]\n",
+        "+-- chardet 3.0.4 [ required: <3.1.0,>=3.0.2 ]\n",
+        "+-- idna 2.7 [ required: <2.8,>=2.5 ]\n",
+        "`-- urllib3 1.22 [ required: <1.24,>=1.21.1 ]\n",
+    )
+    expects = "".join(expects)
```
frostming, line 66: It seems we don't need to join the lines if we remove the trailing commas.

JohnHardy: No problem!

tests/cli/test_list.py
JohnHardy: I explicitly added `idna` and `chardet` them as dev deps in the test and made a note.  I think for the purpose of the test it is reasonable/clearer than a complicated mock.

src/pdm/cli/commands/list.py
```diff
@@ -1,37 +1,417 @@
...
+    ) -> None:
+        if options.csv:
+            raise PdmUsageError("--csv cannot be used with --graph")
+        if options.markdown:
+            raise PdmUsageError("--markdown cannot be used with --graph")
+        if options.sort:
+            raise PdmUsageError("--sort cannot be used with --graph")
+
+        dep_graph = build_dependency_graph(
+            packages, project.environment.marker_environment  # type: ignore
```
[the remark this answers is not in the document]
JohnHardy: Changed to `Mapping[str, im.Distribution]`

src/pdm/cli/commands/list.py
```diff
@@ -1,37 +1,417 @@
...
+        if not all(g in valid_groups for g in include):
+            raise PdmUsageError(
+                f"--include groups must be selected from: {valid_groups}"
+            )
+        exclude = parse_comma_separated_string(options.exclude, lowercase=False)
+        if exclude and not all(g in valid_groups for g in exclude):
+            raise PdmUsageError(
+                f"--exclude groups must be selected from: {valid_groups}"
+            )
+        selected_groups = set(g for g in include if g not in exclude)
```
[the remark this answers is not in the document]
frostming: Not exactly, `include` should be a strong yes, when it is given, only those specified remain, except for `:sub`, which will be absent only when it appears in `exclude`.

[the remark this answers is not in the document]
JohnHardy: Ah, I see, thanks!  I will try to get to it tomorrow :)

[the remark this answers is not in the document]
JohnHardy:
```
valid = set([a, b, c, s])

include=[set()], exclude=[set()] ==> selected=[{'c', 'a', 'b', 'sub'}]
include=[set()], exclude=[{'a'}] ==> selected=[{'c', 'b', 'sub'}]
include=[set()], exclude=[{'a', 'sub'}] ==> selected=[{'c', 'b'}]
include=[{'a'}], exclude=[{'a', 'sub'}] ==> selected=[{'a'}]
include=[{'a', 'b'}], exclude=[set()] ==> selected=[{'a', 'b', 'sub'}]
include=[{'a', 'b'}], exclude=[{'sub'}] ==> selected=[{'a', 'b'}]
```

tests/cli/test_list.py
JohnHardy: I have not worked out how to mock this yet.  How can I access and mock the candidates within the test fixture?

frostming:
@JohnHardy Thanks for the contribution, it may still need to be refined. I will merge it first and continue the work on another branch.

frostming (approved):

JohnHardy:
Fantastic, thank you for making last refinements.

I look forward to seeing the changes - please ping me if I can help in the future👍

PR #1469, opened 2022-10-28
feat: enhance the pdm list command

frostming:
## Pull Request Check List

- Improve `--sort` to accept a comma-separated string
- Packages are included or excluded in the tree approach, when a package is excluded, all its dependencies are excluded, too.
- Modify test cases according to the expected behavior

cc @JohnHardy

### DIFF
--- a/src/pdm/cli/commands/list.py
+++ b/src/pdm/cli/commands/list.py
@@ -288,7 +288,22 @@ class Command(BaseCommand):
         elif options.markdown:
             body = [f"# {project.name} licenses"]
             body.extend(row.markdown(fields) for row in records)
-            ui.echo("\n".join(body), highlight=True)
+            text_body = "\n".join(body)
+            try:
+                ui.echo(text_body, highlight=True)
+            except UnicodeEncodeError:
+                ui.echo(
+                    "Markdown output contains non-ASCII characters. "
+                    "Setting env var PYTHONIOENCODING to 'utf8' may fix this.",
+                    err=True,
+                    style="error",
+                )
+                ui.echo(
+                    text_body.encode().decode("ascii", errors="ignore"), highlight=True
+                )
+                ui.echo(
+                    "**Problem decoding file as UTF-8.  Some characters may be omit.**"
+                )
 
         # Write nice table format.
         else:
--- a/src/pdm/cli/commands/list.py
+++ b/src/pdm/cli/commands/list.py
@@ -1,37 +1,430 @@
 import argparse
+import json
+from collections import defaultdict
+from typing import DefaultDict, Dict, List, Optional, Sequence, Set
 
 from pdm.cli import actions
 from pdm.cli.commands.base import BaseCommand
+from pdm.cli.utils import (
+    build_dependency_graph,
+    check_project_file,
+    get_dist_location,
+    show_dependency_graph,
+)
+from pdm.compat import importlib_metadata as im
+from pdm.exceptions import PdmUsageError
+from pdm.models.requirements import Requirement
 from pdm.project import Project
+
+# Group label for subdependencies
+SUBDEP_GROUP_LABEL = ":sub"
 
 
 class Command(BaseCommand):
     """List packages installed in the current working set"""
 
+    DEFAULT_FIELDS = "name,version,location"
+
     def add_arguments(self, parser: argparse.ArgumentParser) -> None:
-        group = parser.add_mutually_exclusive_group()
-        group.add_argument(
-            "--graph", action="store_true", help="Display a graph of dependencies"
-        )
-        group.add_argument(
+        graph = parser.add_mutually_exclusive_group()
+
+        parser.add_argument(
             "--freeze",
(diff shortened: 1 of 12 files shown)


src/pdm/cli/commands/list.py
```diff
@@ -1,37 +1,428 @@
...
+        # Write JSON
+        elif options.json:
+            json_row = [row.json(fields) for row in records]
+            ui.echo(json.dumps(json_row, indent=4), highlight=True)
+
+        # Write Markdown
+        elif options.markdown:
+            body = [f"# {project.name} licenses"]
+            body.extend(row.markdown(fields) for row in records)
+            ui.echo("\n".join(body), highlight=True)
```
JohnHardy, line 291: A few licences had unusual characters that resulted in unicode errors when printed to a non-tty, for instance the `Arpeggio` package. https://github.com/textX/Arpeggio/blob/master/LICENSE 

What will `ui.echo` do with the character?  As it was reproducing a license document, I felt it was best to add a warning to the document that said that some characters were changed. 

Ref: https://rich.readthedocs.io/en/stable/console.html#printing

frostming: It seems both `ui.echo` and `print` work on my machine, maybe the terminal encoding is fine for me.

JohnHardy: Try piping the output to a file to trigger the error (powershell, windows terminal):

**pdm list --fields name,groups,licenses --markdown -v > test.md**

Here is the stack trace:
```
PS C:\Users\John\Desktop\pdm> pdm run pdm list --fields name,groups,licenses --markdown -v > test.md
  File "C:\Users\John\AppData\Local\Programs\Python\Python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\John\AppData\Local\Programs\Python\Python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\Scripts\pdm.EXE\__main__.py", line 7, in <module>
  File "C:\Users\John\Desktop\pdm\src\pdm\core.py", line 258, in main
    return Core().main(args)
  File "C:\Users\John\Desktop\pdm\src\pdm\core.py", line 190, in main
  File "C:\Users\John\Desktop\pdm\src\pdm\core.py", line 185, in main
    f(project, options)
  File "C:\Users\John\Desktop\pdm\src\pdm\cli\commands\list.py", line 187, in handle
    self.handle_list(packages, name_to_groups, project, options)
  File "C:\Users\John\Desktop\pdm\src\pdm\cli\commands\list.py", line 291, in handle_list
    ui.echo("\n".join(body), highlight=True)
  File "C:\Users\John\Desktop\pdm\src\pdm\termui.py", line 173, in echo
    console.print(message, **kwargs)
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\lib\rich\console.py", line 1694, in print
    self._buffer.extend(new_segments)
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\lib\rich\console.py", line 848, in __exit__
    self._exit_buffer()
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\lib\rich\console.py", line 806, in _exit_buffer
    self._check_buffer()
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\lib\rich\console.py", line 1999, in _check_buffer
    legacy_windows_render(
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\lib\rich\_windows_renderer.py", line 19, in legacy_windows_render
    term.write_text(text)
  File "C:\Users\John\Desktop\pdm\__pypackages__\3.9\lib\rich\_win32_console.py", line 403, in write_text
    self.write(text)
  File "C:\Users\John\AppData\Local\Programs\Python\Python39\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u0107' in position 307: character maps to <undefined>
PS C:\Users\John\Desktop\pdm>
```

To me, it looks like the issue is when it tries to write a non-ascii character to an ascii pipe.

Try:

```
text_body = "\n".join(body)
try:
     ui.echo(text_body, highlight=True)
except UnicodeEncodeError:
    ui.echo(text_body.encode().decode("ascii", errors="ignore"), highlight=True) 
    ui.echo("Problem decoding file as UTF-8.  Some characters may be omit.")
```

frostming: > Try:

Why, if this fails, no content is produced, then it is no different from just popping the exception info.

JohnHardy: Ah - for me it _did_ produce content and left a half complete markdown file.  I thought somebody might not notice (e.g. if running in a CI).

I thought that the exception should not stop the other projects being listed - since the charset in the license file is something the pdm user has no control over.

Issue #1441, opened 2022-10-17
Cached version of current project used when using `pdm lock`


tgolsson:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

## Steps to reproduce

I'm working on a package that that has a self-referential pattern for its own optional dependencies; as described for `foo[socks,jwt]` here: https://pdm.fming.dev/latest/usage/dependency/#add-dependencies. I ran `pdm lock` to generate a lockfile. I then changed the description and ran `pdm lock` again.

The exact config is:
```toml
[project.optional-dependencies]
a = ["some-pkg==1.2.3"]
b = [
    "another-package=1.2.3.4",
]
ci = ["gsutil>=4.66", "pkg[a,b]"]
```

## Actual behavior

After changing the description, the base case `foo` uses the old summary but the target `foo[socks,jwt]` uses the new summary.

Like this:

```toml

[[package]]
name = "pkg"
version = "0.1.0"
requires_python = "~=3.8"
summary = "old summary"
dependencies = [ "..." ]

[[package]]
name = "pkg"
version = "0.1.0"
extras = ["a", "b"]
requires_python = "~=3.8"
summary = "new summary"
dependencies = [ "..." ]
```

## Expected behavior

Both package references use the same summary. Deleting `~/.cache/pdm/metadata/package_meta_NNNNN.json` and running `pdm lock` again solves the issue, making both references use `new summary`.

## Environment Information

```bash
# Paste the output of `pdm info && pdm info --env` below:
PDM version:
  2.1.5
Python Interpreter:
  <REDACTED>/pkg/.venv/bin/python (3.8)
Project Root:
  <REDACTED>/pkg
Project Packages:
  None
{
  "implementation_name": "cpython",
  "implementation_version": "3.8.10",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.15.68.1-microsoft-standard-WSL2",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Mon Sep 19 19:14:52 UTC 2022",
  "python_full_version": "3.8.10",
  "platform_python_implementation": "CPython",
  "python_version": "3.8",
  "sys_platform": "linux"
}
```

frostming:
Is `pkg` a published package? then updating description should bump the version right?

tgolsson:
[...]

PR #1470, opened 2022-10-28
Account for self-requirement in lockfile hash

tgolsson:
## Pull Request Check List

If you have a pyproject where you depend on yourself using the pattern described in #1441, with a config such as this:

```toml
[project]
name = 'pkg'
version = '1.1'

[project.optional-dependencies]
a = ["some-pkg==1.2.3"]
b = [
    "another-package=1.2.3.4",
]
ci = ["gsutil>=4.66", "pkg[a,b]"]
```

Both the name and the version of pkg ends up in the lockfile. However, changing either of those properties under the `project` heading does not currently cause the lockfile hash to change. However rerunning `pdm lock` causes the contents to change, This causes `pdm lock --check` to succeed but a subsequent `pdm sync` fails.

This might even be a problem without an explicit self dependency, this is just mirroring the setup where I found the issue.


tgolsson:
Not sure how to deal with failing tests re: missing name for the example project. Should there be a default? Should we skip adding the name to the hash if there's not a name? Not sure what the PEP says about whether a project without those properties being valid or not. Given the phrasing of the error , it sounds like the test projects are maybe misconfigured.

src/pdm/project/core.py
```diff
@@ -530,9 +530,12 @@ def make_self_candidate(self, editable: bool = True) -> Candidate:
         return make_candidate(req, name=self.meta.name, version=self.meta.version)
 
     def get_content_hash(self, algo: str = "md5") -> str:
-        # Only calculate sources and dependencies groups. Otherwise lock file is
+        # Generate a hash based on the external dependencies, as well
+        # as the identity of the project. Otherwise lock file is
         # considered as unchanged.
         dump_data = {
+            "version": self.meta.version,
```
frostming, line 537: Do we have to count `version` in? That may change in EVERY commit if we are using dynamic versioning from SCM.

tgolsson: Yes, I think so. Or we need to relax what version gets put into the lockfile for the local package.

I put up an isolated repro and observe the issue on 2.1.5, as well as on latest main before my fix. https://github.com/tgolsson/pdm-repros/tree/main/1470

frostming: I don't think it is the correct approach to fixing the issue. The self package, ideally, shouldn't appear in the lockfile.

tgolsson: I've done another implementation which seems to work equally well (also doesn't break the other tests!): https://github.com/pdm-project/pdm/compare/main...tgolsson:pdm:ts/1470-again. Let me know what 

I don't love the way I'm dealing with the group -> candidate info translation, but it seemed like the clearest way of handling it. What I wanted to do was generate a "synthetic" candidate + candidate info, but it seemed like more work than this approach, and I wasn't sure how it'd handle extras since those aren't known until we try to resolve.

frostming: There may exist dynamic variables in the dependencies, so you should use `project.meta.dependencies` and `project.meta.optional_dependencies`, respectively.

And in L460:
```python
        if strip_extras(k)[0] == project.name:
            continue
```

`k` is normalized while `project.name` isn't. You need to normalize it with `normalize_name` and be careful of the possible `None` value.

PR #1481, opened 2022-10-31
Do not add local package to lockfile

tgolsson:
Replaces #1470 

## Pull Request Check List

If you have a pyproject where you depend on yourself using the pattern described in https://github.com/pdm-project/pdm/issues/1441, with a config such as this:

```
[project]
name = "pkg"
version = "0.2.0"
dependencies = []

requires-python = ">=3.8"
license = {text = "MIT"}

[project.optional-dependencies]
foobar = ["pyyaml"]
ci = ["pkg[foobar]"]
```

The project will appear multiple times in the lockfile. Changing the version or name of the local package will desync from the lockfile; leading to `pdm sync` failing. This PR removes the local package from the lockfile, and instead adds a generator for it when running from a `LockedRepository`.

This works and all tests pass. I did however notice that when using this pattern the local package gets installed twice; but that happens on main as well:

```
$ pdm --version
PDM, version 2.2.0
$ pdm install -G ci
Synchronizing working set with lock file: 2 to add, 0 to update, 0 to remove

  ✔ Install pkg 0.2.0 successful
  ✔ Install pyyaml 6.0 successful
Installing the project as an editable package...
  ✔ Install pkg 0.2.0 successful

🎉 All complete!
```

Not sure if there's a good opportunity to fix that here as well.


frostming:
> I did however notice that when using this pattern the local package gets installed twice; but that happens on main as well:

It can be fixed by dropping the self-package from the resolution result. Then no need to skip it in `format_lockfile` anymore.

src/pdm/resolver/core.py
```diff
@@ -2,7 +2,9 @@
 
 from typing import TYPE_CHECKING, Dict, cast
 
+from pdm.cli.utils import normalize_name, strip_extras
```
frostming, line 5: Wrong import path, please fix it

frostming (changes requested):

frostming:
It seems the latest ubuntu runner has some issues with setup-python 3.6. But I can't find a bug report about this.

frostming (approved):

Issue #1496, opened 2022-11-05
PDM doesn't cache wheels based on ETag/Last-Modified headers


tgolsson:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

**Make sure you run commands with `-v` flag before pasting the output.**

## Context

I'm trying to build a file-cache plugin for CI to avoid downloading from PYPI, instead hitting a local GCS bucket (or other file-based backend). This works by running a minimal resolve + link resolution (mimicking what happens up to and during PreparedCandidate.obtain), then filling the cache with only those files. Works awesome 99% of the time - except for a single 200 MB wheel, which is unfortunately the most important one due to the size. Even if I fill it properly, PDM replaces the file with a freshly downloaded variant on each run. 

I'm 99% sure at this point that this is a misconfiguration somewhere leading PDM/Unearth/cachecontrol to ignore the ETag and/or Last-Modified headers. I'm posting here while investigating (and in-case it's a deliberate choice).  

This is an example header configuration:

```
accept-ranges: bytes
age: 683
content-length: 199261216
content-type: binary/octet-stream
date: Sat, 05 Nov 2022 12:35:14 GMT
etag: "7dd8fd1edf6f171be8efc30773e1af3d-24"
last-modified: Thu, 27 Jan 2022 19:11:45 GMT
server: AmazonS3
via: 1.1 432d52d55ad517cddd9081b248b2f116.cloudfront.net (CloudFront)
x-amz-cf-id: 3AIysyU1PZd2Nu7QrgTD12CqDwVjkH-F5WhbifPv8nNCwaVm2sSQgA==
x-amz-cf-pop: ARN54-C1
x-amz-version-id: uTkfkquaVRwOzsDfqejmY0gUwSY.ZOer
x-cache: Hit from cloudfront
```
It looks like etag and and last-modified are constant across invocations from a bunch of different in Europe since yesterday evening, and I'd expect that to be enough to cache the file.

Furthermore; since the content of the file on disk (as written by cachecontrol) changes due to the `date` header; my cache also gets busted - I cache on sha256 sum, so if PDM changes a file on disk during the run I also invalidate my cache.

## Steps to reproduce

Checkout this repository: https://github.com/tgolsson/pdm-repros/tree/main/constant-redownload

Following the readme; run the following:

```bash
$ pdm lock
$ pdm sync -v
$ pdm sync -v
```

[...]

PR #1498, opened 2022-11-06
Use split-body cache to make cached wheels disk-stable

tgolsson:
Hey!

This relates to #1496 since it's one of the things that caused files wheels to "change" when they really were just modifying headers.

Doing a bit of an ugly PR here just to get a check for what you think. The current usage of `cachecontrol` puts both the wheel and headers into a single file, in a way that is very hard to read from an external tool. It also means that any updates to header metadata requires reading the whole file, and then writing it back out to disk twice: both for the cache and for the actual build-dir.

With this change, the body of the response doesn't have to get written back to disk for the cache when only headers change (e.g. a 304 response). This reduces disk pressure and makes it easier to sync the pdm-cache elsewhere. It also makes caches easier to manage without reading the metadata: the sha256 of the body itself is enough to identify whether the cache is valid.

This is what the directory would look like with this change:

```
~/.cache/pdm2/http/2/8/4/c/0/284c004432c396e929f45be954b586ae9cf025296e61667ae6c285bf
~/.cache/pdm2/http/2/8/4/c/0/284c004432c396e929f45be954b586ae9cf025296e61667ae6c285bf.body
```

Unfortunately merging this requires a new release of cachecontrol; because 0.12.11 (latest) has a few bugs for the separate body cache. They can all be solved with inheritance/patching, but I don't think it's worth doing that work unless cachecontrol plans to wait for a long time. I've asked that question here: https://github.com/ionrock/cachecontrol/issues/288.

Another option I also considered was changing to a format like this:

```
~/.cache/pdm2/http/2/8/4/c/0/284c004432c396e929f45be954b586ae9cf025296e61667ae6c285bf/pkg-wheel-file-name
~/.cache/pdm2/http/2/8/4/c/0/284c004432c396e929f45be954b586ae9cf025296e61667ae6c285bf/metadata
```

This makes it easy to find the actual files; *and* the files can be fed directly to PIP for debugging etc. Downside is requiring to parse out the filename from the link. In this case we'd also need to deal with upgrading or explicitly invalidating the cache. In the first case, it will require redownloading the data, but isn't "breaking" beyond that.

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -25,6 +25,8 @@ summary = "Fast, simple object-to-object and broadcast signaling"
 name = "cachecontrol"
 version = "0.12.11"
 requires_python = ">=3.6"
+git = "https://github.com/ionrock/cachecontrol"
+revision = "c05ef9eff1c9ac176481fb99e7a7188aa5b4e17b"
 summary = "httplib2 caching for requests"
 dependencies = [
     "msgpack>=0.5.2",
@@ -36,10 +38,12 @@ name = "cachecontrol"
 version = "0.12.11"
 extras = ["filecache"]
 requires_python = ">=3.6"
+git = "https://github.com/ionrock/cachecontrol"
+revision = "c05ef9eff1c9ac176481fb99e7a7188aa5b4e17b"
 summary = "httplib2 caching for requests"
 dependencies = [
-    "cachecontrol==0.12.11",
-    "lockfile>=0.9",
+    "cachecontrol @ git+https://github.com/ionrock/cachecontrol",
+    "filelock>=3.8.0",
 ]
 
 [[package]]
@@ -118,7 +122,7 @@ summary = "execnet: rapid multi-Python deployment"
 
 [[package]]
 name = "filelock"
-version = "3.7.1"
+version = "3.8.0"
 requires_python = ">=3.7"
 summary = "A platform independent file lock."
 
@@ -188,11 +192,6 @@ summary = "A very fast and expressive template engine."
 dependencies = [
     "MarkupSafe>=2.0",
 ]
-
-[[package]]
-name = "lockfile"
-version = "0.12.2"
-summary = "Platform-independent file locking module"
 
 [[package]]
 name = "markdown"
@@ -678,8 +677,8 @@ requires_python = ">=3.7"
 summary = "Backport of pathlib-compatible object wrapper for zip files"
 
 [metadata]
-lock_version = "4.0"
-content_hash = "sha256:7fa3a9a57646b14e66d07f50ce0ca798d47856f7fa901203db8d703b131c1853"
+lock_version = "4.1"
+content_hash = "sha256:a7391d64ac55042a0568037e58cde1437d81da2cad361bba5cf8e3daff984370"
 
 [metadata.files]
 "arpeggio 1.10.2" = [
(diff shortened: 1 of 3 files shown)


tgolsson:
Not sure why CI is sad here; all pass locally. (Also *very slow, but that is likely pytest being a pain somehow.) 
> 596 passed in 2044.04s (0:34:04)

frostming:
Sorry but it isn't clear to me why this and the latest fix on cachecontrol would help resolve the cache missing.

If I read [this function](https://github.com/ionrock/cachecontrol/blob/c05ef9eff1c9ac176481fb99e7a7188aa5b4e17b/cachecontrol/controller.py#L145) correctly, either `max-age` or `expires` must be present in the response headers(given the status code is 200) or request headers.

Maybe I am just misreading that, thanks in advance for your explanation.

tgolsson:
Let me try to clarify. To start with, the cache *is* working. It isn't very clear, but it works. When those heades are missing, the final fallback for reusing the cache is `Last-Modified` and/or `ETag`s. `cachecontrol` does this, by sending a GET with the ETag from the current cache. If that ETag is still valid, the server will reply with status 304 indicating that we can still use the cache.

However, `cachecontrol` needs to save the new `Date` header from the 304 reply. `cachecontrol` does this by building a "cache file", which looks something like this:

```bytes
cc=4,headers={ 'Date':' ...', 'Age': ..., 'ETag': '...', 'body': '...'}
```

And the body is the whole wheel or sdist. This means that when we get a 304 back, we have to replace the whole cache item just to update the `Date` - both writing it, and more importantly for me changing the sha256. For a 1kB wheel or sdist that doesn't matter. For a 200 MB (or 3GB) torch library it is hugely wasteful. The split cache is meant to solve this. It separates the cache metadata and the cache item in two files. The first one is the same, except the body field is left empty. That is instead stored in a separate file. It'd look like this:

```
/x/y/z/some-hash.metadata 1kB
/x/y/z/some-hash          200MB
```

And only the first file would be updated, and the second wouldn't. For my GCS-cache, that means I can skip uploading the cache again and just (optionally) uploading the metadata file.

frostming:
> Let me try to clarify. To start with, the cache _is_ working. It isn't very clear, but it works. When those heades are missing, the final fallback for reusing the cache is `Last-Modified` and/or `ETag`s. `cachecontrol` does this, by sending a GET with the ETag from the current cache. If that ETag is still valid, the server will reply with status 304 indicating that we can still use the cache.
> 
> However, `cachecontrol` needs to save the new `Date` header from the 304 reply. `cachecontrol` does this by building a "cache file", which looks something like this:
> 
> ```
> cc=4,headers={ 'Date':' ...', 'Age': ..., 'ETag': '...', 'body': '...'}
> ```
> 
> And the body is the whole wheel or sdist. This means that when we get a 304 back, we have to replace the whole cache item just to update the `Date` - both writing it, and more importantly for me changing the sha256. For a 1kB wheel or sdist that doesn't matter. For a 200 MB (or 3GB) torch library it is hugely wasteful. The split cache is meant to solve this. It separates the cache metadata and the cache item in two files. The first one is the same, except the body field is left empty. That is instead stored in a separate file. It'd look like this:
> 
> ```
> /x/y/z/some-hash.metadata 1kB
> /x/y/z/some-hash          200MB
> ```
> 
> And only the first file would be updated, and the second wouldn't. For my GCS-cache, that means I can skip uploading the cache again and just (optionally) uploading the metadata file.
> 
> (And the reason I thought that the cache wasn't working was because when the file is replaced, it both changes the creation timestamp and the sha256sum - so it looks like a fully new file. And all that because it was writing a 10-byte header field.)

Thanks, I got what is happening now, that is the line: https://github.dev/ionrock/cachecontrol/blob/c05ef9eff1c9ac176481fb99e7a7188aa5b4e17b/cachecontrol/controller.py#L277-L280

This is unreleased right?

tgolsson:
Yes, indeed. 

Yep! In this branch I've put the dependency to run from github for testing, and I'll likely switch to running this on our CI until cachecontrol releases their next version and we can merge this (if you want to merge it).

frostming:
> Yes, indeed.
> 
> Yep! In this branch I've put the dependency to run from github for testing, and I'll likely switch to running this on our CI until cachecontrol releases their next version and we can merge this (if you want to merge it).

Maybe alternatively you can implement the `set_body` method for the `SeparateBodyCache` and skip if the input body is None? In this way, we are not blocked by the release yeah?

tgolsson:
Unfortunately the latest release sometimes skips loading the body completely no matter what I did. It's related to this upstream fix: https://github.com/ionrock/cachecontrol/commit/6155c199438bfc3944a833f1dec6fb74134476f1

frostming:
> Unfortunately the latest release sometimes skips loading the body completely no matter what I did. It's related to this upstream fix: [ionrock/cachecontrol@6155c19](https://github.com/ionrock/cachecontrol/commit/6155c199438bfc3944a833f1dec6fb74134476f1)

It seems to belong to a yanked release 0.12.12

tgolsson:
Yeah, 0.12.12 did something weird that ended up breaking a bunch of downstreams, so it should have been released as 0.13 as I understand it.

tgolsson:
FYI this hasn't been abandoned, just no movement upstream. I'm unsure how to proceed if that doesn't get unstuck, I do think this would be worthwhile, especially when using remote caching.

frostming:
> I do think this would be worthwhile, especially when using remote caching.

Agree, it causes caching of the PyTorch index totally unusable.

tgolsson:
Saw the new release! I'm a bit swamped right now with RL stuff, but will see if I can get this up to date and test it again soon~. 🥳

frostming:
@tgolsson Go ahead if you want to take this. Also please switch back to `cachecontrol` since it was replaced by `cacheyou` before.

tgolsson:
Will do! :-)

tgolsson:
Replaced by #1971 .

PR #1512, opened 2022-11-14
fix venv.with_pip syntax error

dev590t:
## Pull Request Check List


frostming:
Thanks

Issue #1531, opened 2022-11-25
Clarify the behavior of includes


gsalvatella:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

The behavior of [includes](https://pdm.fming.dev/latest/pyproject/build/#include-and-exclude-files) is not clear. It seems to imply that any files listed will be *added* to the package itself. However the latter has to be also explicitly included. Either change the behavior to append or clarify in the docs that once `includes` is used the package needs to be included.

Suggestion: use an admonition or annotation

```markdown
!!! info

  When using includes the package itself is not included by default
```

## Steps to reproduce

Use the following setup

```
mypkg/
  __init__.py
  main.py
README.md
pyproject.toml
```

Build with

```bash
pdm build && unzip -l dist/*whl
...
mypkg/__init__.py
mypkg/main.py
mypkg-0.1.0.dist-info/WHEEL
mypkg-0.1.0.dist-info/METADATA
mypkg-0.1.0.dist-info/RECORD
```

Add includes

```pyproject.toml
[tool.pdm.build.includes]
  includes = ['README.md']
```

## Actual behavior

The output is 

```bash
README.md
mypkg-0.1.0.dist-info/WHEEL
mypkg-0.1.0.dist-info/METADATA
mypkg-0.1.0.dist-info/RECORD
```

## Expected behavior

The output is 

```bash
mypkg/__init__.py
mypkg/main.py
README.md
mypkg-0.1.0.dist-info/WHEEL
mypkg-0.1.0.dist-info/METADATA
mypkg-0.1.0.dist-info/RECORD
```

## Environment Information

```bash
PDM version:
  2.2.1
Python Interpreter:
  /usr/bin/python3.10 (3.10)
Project Root:
  /tmp/tmp.vpcmLVlgGv
Project Packages:
  /tmp/tmp.vpcmLVlgGv/__pypackages__/3.10
{
  "implementation_name": "cpython",
  "implementation_version": "3.10.8",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.15.78-1-MANJARO",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT Thu Nov 10 20:50:09 UTC 2022",
  "python_full_version": "3.10.8",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "linux"
}
```

PR #1532, opened 2022-11-25
docs: clarify the behavior of includes

gsalvatella:
The behavior of `includes` seems to imply that any files listed will be *added* to the package itself. However the latter has to be also explicitly included. This can cause confusion when adding `includes` for the first time in an already existing package, whereby the package itself will suddenly disappear from the build. Clarify this in the documentaion.

Fixes https://github.com/pdm-project/pdm/issues/1531

## Pull Request Check List

- [ ] A news fragment is added in `news/` describing what is new. N/A
- [ ] Test cases added for changed code. N/A

### DIFF
--- a/docs/docs/pyproject/build.md
+++ b/docs/docs/pyproject/build.md
@@ -95,6 +95,10 @@ excludes = [
 ]
 ```
 
+!!! note
+
+	When using `includes` the default includes will be overriden. You have to add the package paths manually.
+
 In case you may want some files to be included in source distributions only, use the `source-includes` field:
 
 ```toml


docs/docs/pyproject/build.md
```diff
@@ -95,6 +95,10 @@ excludes = [
 ]
 ```
 
+!!! note
+
+	When using `includes` the package itself is not included by default. You have to add it manually as well.
```
frostming, line 100:
```suggestion
	When using `includes` the default includes will be overridden. You have to add the package paths manually.
```

gsalvatella:
@frostming Another interesting option would be to consider adding code block [annotiations](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations). This is a great example of them being useful to not clutter the main text. Would you fancy the idea?

frostming:
> @frostming Another interesting option would be to consider adding code block [annotations](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations). This is a great example of them being useful to not clutter the main text with admonitions and clarifications. Would you fancy the idea?

Good idea, i didn't know that plugin before

PR #1533, opened 2022-11-25
Allows specifying the insertion position of user provided arguments in scripts

### LINKED ISSUE #1507: Allows Specifying the Insertion Position of Extra Arguments Supplied to `pdm run`

I'm using PDM scripts to both run commands in PDM environment, and also as a shorthand for providing common parameters.

For example, in a Django project, it's common to run commands with `manage.py`. I will use script like

```toml
manage = { cmd = [
    'python', 'manage.py',
] }
```

So that I can use command `pdm manage migrate`, instead of `pdm run python manage.py migrate`. Where the common prefix `python manage.py ...` can be reused.

However, I found that one of my tools (specifically, [`pre-commit`](https://pre-commit.com/)) has commands and arguments in this format:

```shell
pre-commit install --config somewhere/config.yml
pre-commit autoupdate --config somewhere/config.yml
pre-commit run --config somewhere/config.yml
```

The parameters like `--config` must come after commands like `install`. If I create script

```toml
pre-commit = { cmd = [
    'pre-commit', '--config', 'somewhere/config.yml',
] }
```

and execute `pdm run pre-commit install`, the additional argument `install` will always be appeneded to the `cmd` and resulting

```
pre-commit --config somewhere/config.yml install
```

which is invalid for `pre-commit`.

Since the common parts (`pre-commit` and `--config somewhere/config.yml`) of the commands are not a prefix, there is no way for me to create shorthand PDM scripts for them.

**It would be greate if we can specify the insertion position of extra arguments supplied to `pdm run`.**

We can provide an optional placeholder to the `cmd`. e.g.:

```toml
pre-commit = { cmd = [
    'pre-commit', '$args', '--config', 'somewhere/config.yml',
] }
```

noirbizarre:
## Pull Request Check List

Allows specifying the insertion position of user provided arguments in scripts with the `{args[:default]}` placeholder.

The placeholder can be used in two forms:
- `{args}` which will be replaced by user provided extra arguments if any
- `{args:default values}` which will be replaced by user provided extra arguments if any or else by the defaults values

As soon as a placeholder is detected,  the extra user arguments are not automaticcaly appended anymore. This is especially important for composite scripts: a single placeholder deactivate the automatic append on all subscripts giving you control on which subscripts receive the arguments or not.

**Edit**: `call` scripts was initialy added to this PR but has been removed given they have access to `sys.argv` directly to handle such cases.

Fixes #1507

### DIFF
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+import itertools
 import os
 import re
 import shlex
@@ -8,7 +9,7 @@ import signal
 import subprocess
 import sys
 from types import FrameType
-from typing import Any, Callable, Mapping, NamedTuple, Sequence, cast
+from typing import Any, Callable, Iterator, Mapping, NamedTuple, Sequence, cast
 
 from pdm import signals, termui
 from pdm.cli.actions import PEP582_PATH
@@ -44,6 +45,20 @@ def exec_opts(*options: TaskOptions | None) -> dict[str, Any]:
             if k not in ("env", "help")
         },
     )
+
+
+RE_ARGS_PLACEHOLDER = re.compile(r"{args(?::(?P<default>[^}]*))?}")
+
+
+def interpolate(script: str, args: Sequence[str]) -> tuple[str, bool]:
+    """Interpolate the `{args:[defaults]} placeholder in a string"""
+
+    def replace(m: re.Match[str]) -> str:
+        default = m.group("default") or ""
+        return " ".join(args) if args else default
+
+    interpolated, count = RE_ARGS_PLACEHOLDER.subn(replace, script)
+    return interpolated, count > 0
 
 
 class Task(NamedTuple):
@@ -216,12 +231,24 @@ class TaskRunner:
         kind, _, value, options = task
         shell = False
         if kind == "cmd":
-            if not isinstance(value, list):
-                value = shlex.split(str(value))
-            args = value + list(args)
+            if isinstance(value, str):
+                cmd, interpolated = interpolate(value, args)
+                value = shlex.split(cmd)
+            else:
+                agg = [interpolate(part, args) for part in value]
+                interpolated = any(row[1] for row in agg)
+                # In case of multiple default, we need to split the resulting string.
+                parts: Iterator[list[str]] = (
+                    shlex.split(part) if interpolated else [part]
+                    for part, interpolated in agg
+                )
+                # We flatten the nested list to obtain a list of arguments
+                value = list(itertools.chain(*parts))
+            args = value if interpolated else [*value, *args]
         elif kind == "shell":
             assert isinstance(value, str)
-            args = " ".join([value] + list(args))  # type: ignore
+            script, interpolated = interpolate(value, args)
+            args = script if interpolated else " ".join([script, *args])
             shell = True
         elif kind == "call":
             assert isinstance(value, str)
@@ -241,7 +268,6 @@ class TaskRunner:
             ] + list(args)
         elif kind == "composite":
             assert isinstance(value, list)
-            args = list(args)
 
         self.project.core.ui.echo(
             f"Running {task}: [success]{str(args)}[/]",
@@ -249,10 +275,16 @@ class TaskRunner:
             verbosity=termui.Verbosity.DETAIL,
         )
         if kind == "composite":
+            args = list(args)
+            should_interpolate = any(
+                (RE_ARGS_PLACEHOLDER.search(script) for script in value)
+            )
             for script in value:
+                if should_interpolate:
+                    script, _ = interpolate(script, args)
                 split = shlex.split(script)
                 cmd = split[0]
-                subargs = split[1:] + args  # type: ignore
+                subargs = split[1:] + ([] if should_interpolate else args)
                 code = self.run(cmd, subargs, options)
                 if code != 0:
                     return code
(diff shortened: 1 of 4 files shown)


frostming:
> In the case of `call` scripts, the arguments are automatically quoted and comma-separated.
> This mean the following snippets:

I don't think it is necessary to support `call`. Because this is designed to pass a function with optional literal arguments. If it need to accept options from command line, the users should make it a CLI main function(like `click.command` or `argparse.Parser`, then the CLI arguments always go as `sys.argv[1:]`, it doesn't make sense to add a placeholder there.

noirbizarre:
To be honnest, I wasn't initially intending to support `call` scripts but seeing the test suite, there already was some tests with the same argument cases than `shell` and `cmd`  (aka. pyproject provided arguments and user provided extra arguments) so I added `call` support to be consistent.
But you are right, this make no sense given call script can access `sys.argv` directly.
I will happily remove this for `call` support as it was in addition introducing sme uncessary complexity 👍🏼

noirbizarre:
Updated with `call` scripts support removed. I added a quick note on why `call` scripts don't support it.

src/pdm/cli/commands/run.py
```diff
@@ -216,12 +227,21 @@ def _run_task(
...
+            interpolations = 0
+            if isinstance(value, str):
+                cmd, count = interpolate(value, args)
+                interpolations += count
+                value = shlex.split(cmd)
+            else:
+                agg = [interpolate(part, args) for part in value]
+                parts, counts = zip(*agg)
+                interpolations += sum(counts)
+                value = list(parts)
```
frostming, line 239: Given the command: `["foo", "{args}", "bar"]` and `arg=--default --other`, I assume the value would be `["foo", "--default --other", "bar"]`, right?

src/pdm/cli/commands/run.py
```diff
@@ -46,6 +46,17 @@ def exec_opts(*options: TaskOptions | None) -> dict[str, Any]:
     )
 
 
+RE_ARGS_PLACEHOLDER = re.compile(r"{args(?::(?P<default>[^}]*))?}")
+
+
+def interpolate(script: str, args: Sequence[str]) -> tuple[str, int]:
+    def replace(m: re.Match[str]) -> str:
+        default = m.group("default")
```
frostming, line 54: When default is not given this value will be `None`

noirbizarre: Yes, exactly. This is expected as `re.sub(n)` replace `None` returns by `""` which is what we want and so no need to have `default = m.group("default") or ""`

src/pdm/cli/commands/run.py
noirbizarre: Hum, I guess you are right, I haven't tested the `cmd` as list case. I will

src/pdm/cli/commands/run.py
noirbizarre: Changed it to be explicitely `""` when there is no default (also make it consistent with `re.sub(n)` method which if it properly handle the `None` case expects a `Callable[[re.Match[str]], str]`

src/pdm/cli/commands/run.py
noirbizarre: Test cases updated and cases fixed

src/pdm/cli/commands/run.py
```diff
@@ -216,12 +228,24 @@ def _run_task(
         kind, _, value, options = task
         shell = False
         if kind == "cmd":
-            if not isinstance(value, list):
-                value = shlex.split(str(value))
-            args = value + list(args)
+            interpolations = 0
```
frostming, line 231: There seems no need to accumulate this value, no loop is involved, and it is only used in `kind == "cmd"` branch.

src/pdm/cli/commands/run.py
```diff
@@ -216,12 +228,24 @@ def _run_task(
...
+            if isinstance(value, str):
+                cmd, count = interpolate(value, args)
+                interpolations += count
```
frostming, line 234:
```suggestion
                cmd, interpolations = interpolate(value, args)
```

src/pdm/cli/commands/run.py
```diff
@@ -216,12 +228,24 @@ def _run_task(
...
+                value = shlex.split(cmd)
+            else:
+                agg = [interpolate(part, args) for part in value]
+                interpolations += sum(row[1] for row in agg)
```
frostming, line 238:
```suggestion
                interpolations = sum(row[1] for row in agg)
```

src/pdm/cli/commands/run.py
```diff
@@ -216,12 +228,24 @@ def _run_task(
...
+                parts = (
+                    shlex.split(interpolated) if count else [interpolated]
+                    for interpolated, count in agg
+                )
+                value = list(itertools.chain(*parts))
```
frostming, line 243: I can read this, but we may need some minimum comments.

frostming (changes requested):

frostming:
I left some comments and I think it will be the last round of review. Thanks

src/pdm/cli/commands/run.py
noirbizarre: Removed

src/pdm/cli/commands/run.py
noirbizarre: I added both comments to explain and typing for the intermediate `parts` variable

src/pdm/cli/commands/run.py
noirbizarre: I refactored ² interpolate²  to return `tuple[str, bool]` and removed the need for loop and aggregate.
It should be cleaner and easier to read and understand

src/pdm/cli/commands/run.py
noirbizarre: Refactored to be clearer

frostming (approved):

PR #1539, opened 2022-12-02
Fix minimum version requirement for tomlkit

snowwm:
PDM uses the method `unwrap` from `tomlkit` which was added only in `tomlkit-0.11.0`. When I upgraded PDM from an old version and ran it, I got an error because of this.

## Pull Request Check List

I think these tasks are unnecessary for such a small change:

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -682,7 +682,7 @@ summary = "Backport of pathlib-compatible object wrapper for zip files"
 
 [metadata]
 lock_version = "4.1"
-content_hash = "sha256:b262d087293be7bfcf3045800bd0153f1834d2423c5885bdfa8be1f3425fa6e6"
+content_hash = "sha256:2b6158801bac6ee329b8d5c7f51bf9cb3b6504bebc15be29f3ff9d795c579320"
 
 [metadata.files]
 "arpeggio 1.10.2" = [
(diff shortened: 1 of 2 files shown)


frostming:
Can you also update the pdm hash by running `pdm lock --refresh`?

snowwm:
Could you do it yourself please? I have some trouble with that, getting weird errors:
`[RequirementError]: Invalid marker: Invalid marker: '; platform_system == \\"Windows\\""', parse error at '; platfo'`

frostming:
Sure, done.

snowwm:
Sorry, it should be `0.11.1` due to https://github.com/sdispater/tomlkit/commit/1cb4c25288184905807619da1eb2d3576cef2a64

pdm.lock
```diff
@@ -682,7 +682,7 @@ summary = "Backport of pathlib-compatible object wrapper for zip files"
 
 [metadata]
 lock_version = "4.1"
-content_hash = "sha256:b262d087293be7bfcf3045800bd0153f1834d2423c5885bdfa8be1f3425fa6e6"
+content_hash = "sha256:2b6158801bac6ee329b8d5c7f51bf9cb3b6504bebc15be29f3ff9d795c579320"
```
frostming, line 685:
```suggestion
content_hash = "sha256:cb83045fddc1803e87f7adf801bb652258f42867a95ea1dbae1ae389ca77a8d0"
```

PR #1552, opened 2022-12-07
fix: egg segment for local dependency

micaelmalta:
## Pull Request Check List

The command `pdm add -e ./demo --dev` result on a line added to pyproject.toml

```
dev = [
    "-e file:///${PROJECT_ROOT}/demo#egg=demo",
]
```

Any following `pdm add` will result on an error

```
Adding packages to dev dev-dependencies: -e file:///${PROJECT_ROOT}/demo#egg=demo
[RequirementError]: Editable requirement is only supported for VCS link or local directory.
```

### DIFF
--- a/src/pdm/utils.py
+++ b/src/pdm/utils.py
@@ -410,3 +410,8 @@ def is_pip_compatible_with_python(python_version: Version | str) -> bool:
     pip = importlib_metadata.distribution("pip")
     requires_python = get_specifier(pip.metadata["Requires-Python"])
     return requires_python.contains(python_version, True)
+
+
+def path_without_fragments(path: str) -> Path:
+    """Remove egg fragment from path"""
+    return Path(path.split("#")[0])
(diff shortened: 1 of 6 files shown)


src/pdm/utils.py
```diff
@@ -410,3 +410,8 @@ def is_pip_compatible_with_python(python_version: Version | str) -> bool:
     pip = importlib_metadata.distribution("pip")
     requires_python = get_specifier(pip.metadata["Requires-Python"])
     return requires_python.contains(python_version, True)
+
+
+def path_without_fragments(path: str) -> Path:
+    """Remove egg fragment from path"""
+    return Path(path.split("#")[0])
```
pawamoy, line 417: Should you `.rsplit("#", 1)[0]` to support hashes in the file path? Does that make sense?

micaelmalta: In this case how would we know the hash is on the file path or is a end real fragment ?

micaelmalta: Fixed with commit https://github.com/pdm-project/pdm/pull/1552/commits/8a361c9c8c3ea91b18f05d2a46d5585ca44c41f0

frostming:
Thanks 🎉

Issue #1558, opened 2022-12-08
Pin version of packaging and other essential dependencies in PDM


lambda:
## Is your feature request related to a problem? Please describe.

Packaging 22.0 introduced a number of backwards incompatible changes, which can break projects using PDM:

* #1554
* #1556
* pypa/packaging#629

In order to avoid problems with PDM updates breaking my CI pipelines, I have switched to installing PDM as `pip install pdm~=2.3.1`, in order to get bugfix releases but manually opt in to feature releases.

However, PDM doesn't lock its own dependencies, so a major release of Packaging 22.0 introduced several incompatible changes, without any update to the PDM version.

## Describe the solution you'd like

PDM should lock the exact versions of its own dependencies, so that installing the same version of PDM gives the same results each time. Major version updates of dependencies that affect the PDM interface in backwards-incompatible ways, such as `packaging`, should at least only be done on feature releases of PDM.

pawamoy:
I disagree. It's true that PDM is more an app than a library, but pinning dependencies brings more problems than it solves. I recently started developing a PDM plugin, so my project depends on PDM, and I definitely do not want to be restricted by pins in PDM's dependencies.

Packages incompatibilities can always be solved on the final user side, using constraints for example. On the contrary, locking dependencies gives no alternative to downstream users.

Just my opinion!

lambda:
@pawamoy The whole reason we use PDM is so that we can have reproducible builds, without having unexpected dependency updates breaking us, while still being able to iterate quickly and update dependencies when we need to.

With PDM itself not having pinned dependencies, we lose that ability; updates to PDM or its dependencies can make our builds fail when there have been no changes on our side.

Of course, having proper support for lockfiles in the Python packaging ecosystem would address this, in which you could have a locked set of dependencies for usage as an application, while loose dependencies when used as a library, just as PDM allows us to do for packages managed by PDM. Unfortunately, there's a bootstrapping problem when installing PDM itself.

One possible way to work around this would be to have two published PDM packages; one with pinned dependencies for use as an application, and one with loose dependencies for use as a library.

frostming:
Also, note that #1554 and #1556 are different from https://github.com/pypa/packaging/issues/629, they are not regression issues but breaking changes that remove deprecated behaviors.

As a library keeps evolving it will introduce breaking changes or not. If you are still relying on old behaviors you have to use older versions.

franzhaas:
[...]

Issue #1556, opened 2022-12-08
Issues with resolving dependency with python 3.11.1


ketan-deserve:
Getting below error while running 
`pdm sync --clean --prod --no-self --verbose` 

ERROR: 
`resolvelib.resolvers.ResolutionImpossible: [RequirementInformation(requirement=FileRequirement(name=None, marker=None, extras=None, specifier=None, editable=False, prerelease=False, url=None, path=PosixPath('Django>=3.*'), subdirectory=None)]`

StackTrace : 

```
File "/usr/local/bin/pdm", line 8, in <module>
--
922 | sys.exit(main())
923 | ^^^^^^
924 | File "/usr/local/lib/python3.11/site-packages/pdm/core.py", line 258, in main
925 | return Core().main(args)
926 | ^^^^^^^^^^^^^^^^^
927 | File "/usr/local/lib/python3.11/site-packages/pdm/core.py", line 190, in main
928 | raise cast(Exception, err).with_traceback(traceback)
929 | File "/usr/local/lib/python3.11/site-packages/pdm/core.py", line 185, in main
930 | f(project, options)
931 | File "/usr/local/lib/python3.11/site-packages/pdm/cli/commands/sync.py", line 36, in handle
932 | actions.do_sync(
933 | File "/usr/local/lib/python3.11/site-packages/pdm/cli/actions.py", line 197, in do_sync
934 | candidates = resolve_candidates_from_lockfile(project, requirements)
935 | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
936 | File "/usr/local/lib/python3.11/site-packages/pdm/cli/actions.py", line 140, in resolve_candidates_from_lockfile
937 | mapping, *_ = resolve(
938 | ^^^^^^^^
939 | File "/usr/local/lib/python3.11/site-packages/pdm/resolver/core.py", line 35, in resolve
940 | result = resolver.resolve(requirements, max_rounds)
941 | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
942 | File "/usr/local/lib/python3.11/site-packages/resolvelib/resolvers.py", line 521, in resolve
943 | state = resolution.resolve(requirements, max_rounds=max_rounds)
944 | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1568, opened 2022-12-12
fix: Exclude `packaging 22.0` to avoid breakage

frostming:
## Pull Request Check List

It resolves the following issues, while still allows future packaging versions to be installed

- #1556
- #1558

### DIFF
--- a/news/1560.bugfix.md
+++ b/news/1560.bugfix.md
@@ -1,1 +1,1 @@
-Allow relative paths in `build-system.requirese`, since `build` and `hatch` both support it. Be aware it is not allowed in the standard.
+Allow relative paths in `build-system.requires`, since `build` and `hatch` both support it. Be aware it is not allowed in the standard.
--- a/news/1568.dep.md
+++ b/news/1568.dep.md
@@ -1,0 +1,1 @@
+Exclude `package==22.0` from the dependencies to avoid some breakages to the end users.
--- a/pdm.lock
+++ b/pdm.lock
@@ -337,9 +337,12 @@ summary = "MessagePack serializer"
 
 [[package]]
 name = "packaging"
-version = "22.0"
-requires_python = ">=3.7"
+version = "21.3"
+requires_python = ">=3.6"
 summary = "Core utilities for Python packages"
+dependencies = [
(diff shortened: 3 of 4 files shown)


ferminho:
Hi there, won't this happen again if packaging releases another version?

ferminho:
I might not have the full picture of the issue and I'd indeed like future updates. I simply asked to understand if we could face the same issue when, say, packaging 22.1 is released.
If so, the recommendation is to pin packaging in our side in every project? 

Honest questions, no intention to debate. Thanks a lot for your support!

franzhaas:
@ferminho 
There is a discussion about how to have repeatable deployments of pdm over there.: https://github.com/pdm-project/pdm/issues/1558

pawamoy:
> If so, the recommendation is to pin packaging in our side in every project?

Locally, you probably have a single installation of PDM. In that case you could simply restrict its `packaging` dependency to exclude the new, unsupported version of packaging, until PDM excludes it itself or becomes compatible with it.
In CI, you'll need to add an instruction to your steps/workflow to restrict it as well, until, again, PDM excludes or supports it.

ferminho:
> Locally, you probably have a single installation of PDM. In that case you could simply restrict its `packaging` dependency to exclude the new, unsupported version of packaging, until PDM excludes it itself or becomes compatible with it. This will fix it for all of your local projects.

We have recently adopted pdm as our standard package manager and we have many installations in Python services, libraries and also environments such as JupyterLab or Databricks. This is why the issue created a bit of havoc everywhere when an old, unused version of [pyzqm](https://github.com/zeromq/pyzmq) couldn't be parsed as candidate. 

Thanks! moving to the other conversation

frostming:
> This is why the issue created a bit of havoc everywhere when an old, unused version of [pyzqm](https://github.com/zeromq/pyzmq) couldn't be parsed as candidate.

If this is what blocks you, it has been already solved by the latest version of `unearth`, those legacy illegal packages will be skipped.

lyz-code:
@frostming can you please make a release with this fix? all my CI pipelines will thank you for it

PR #1594, opened 2023-01-03
feat(pytest): expose pytest fixtures and helpers for pdm plugins

noirbizarre:
## Pull Request Check List

This PR follows the discussion https://github.com/pdm-project/pdm/discussions/1571 and extract most pytest fixtures into their own public module `pdm.pytest`.

This allows pdm plugin developpers to benefits from them by importing them as a plugin:
```python
pytest_plugins = [
    "pdm.pytest",
]
```

When relying on local file fixtures or any other configurable part, this is done by a dependency fixture.
The PDM test suite rely on those fixtures and configure them with PDM test fixtures (and so the PDM test suite is an usage example of those fixtures)

Given those fixtures are now in `src`, type hinting has been completed to comply with MyPy.
Documentation has been added in a dedicated page to differentiate from the real PDM API. Obviously, given I didn't write those, there might be some issues with the documentation and typing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,3 +32,4 @@ repos:
         additional_dependencies:
           - types-requests
           - types-certifi
+          - pytest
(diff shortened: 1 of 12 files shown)


frostming (approved):

frostming:
Thanks 🎉

PR #1597, opened 2023-01-05
feat: support multiple indexes in config

### LINKED ISSUE #1310: Multiple pypi.url

I need to define multiple pypi mirrors. As far as I understand it, using ``pdm.config pypi.url``, I can set ONE mirror.

There should be a way to define multiple pypi mirrors.

Alternatively, when adding a package, there should be a parameter available that specifies the mirror that will be used to look for the package.

frostming:
## Pull Request Check List

Resolve #1310


lukaszmoroz:
Extra indexes are read only from the global configuration, is that intended? Also section `Available Configurations` doesn't list `pypi.<name>` items.

frostming:
> Extra indexes are read only from the global configuration, is that intended?

If you mean system config, yes. Users can only change the home config and project config(`--local` option) from the command line.

lukaszmoroz:
Isn't that inconsistent with how the `pypi` key is handled? As in, `pypi.url` can be configured in project-level `.pdm.toml`, but `pypi.extra.url` cannot. I would expect this to be the case from this (https://github.com/pdm-project/pdm/commit/66a2a431a8462cdc057cc17f01766fe92d4c62f3#diff-daa5420dff18bd472f2bd76f06565f9bed47a1fdc6a6ebb498c8e620869310aeR1) commit/change and from the current documentation (although it doesn't mention `pypi.extra.url` specifically, only `pypi.url` etc.)

frostming:
> but `pypi.extra.url` cannot.

I don't know why you get this:

```
❯ pdm config -l pypi.extra.url https://example.org
❯ cat .pdm.toml
[pypi.extra]
url = "https://example.org"
```

lukaszmoroz:
I meant that these configs seem not to be read during `pdm install/add` (neither is `pypi.ignore_stored_index`):
https://github.com/pdm-project/pdm/pull/1597/files#diff-bedef3df88548ee12f64dd39370848af367bc83dcc3f49617951dde95d790649R370

Issue #1584, opened 2022-12-27
pdm import does not check for duplicates


quassy:
Running `pdm import` when you already have dependencies defined (maybe because you want to unify multiple requirements) or just by running it for the same file multiple times will result in duplicate entries.

- [x] I have searched the issue tracker and believe that this is not a duplicate.

(“Make sure you run commands with `-v` flag before pasting the output.”; -v does not seem to change the output of `pdm import`.)

## Steps to reproduce

Have a initialized pdm project with `pyproject.toml` and a simple `requirements.txt` file (`echo sqlalchemy > requirements.txt`).

Run `pdm import requirements.txt -f requirements` multiple times.

## Actual behavior

You get duplicate dependencies:

```
[project]
...
dependencies = ["sqlalchemy", "sqlalchemy"]
```

## Expected behavior

Entries already existing should not be added again.

```
[project]
...
dependencies = ["sqlalchemy"]
```

## Environment Information

```bash
$ pdm info && pdm info --env
PDM version:
  2.3.3
Python Interpreter:
  /path/to/project/.venv/bin/python (3.10)
Project Root:
  /path/to/project
Project Packages:
  None
{
  "implementation_name": "cpython",
  "implementation_version": "3.10.9",
  "os_name": "posix",
  "platform_machine": "arm64",
  "platform_release": "22.2.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 22.2.0: Fri Nov 11 02:06:26 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T8112",
  "python_full_version": "3.10.9",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "darwin"
}
```

frostming:
I will mark it as first-time contribution-friendly since it is not urgent nor breaking anything.

frostming:
Closed by #1627

PR #1627, opened 2023-01-17
Fix: Deduplicate lists when running pdm import

GabDug:
## Pull Request Check List

Avoid duplicates in dependencies/authors when importing from requirements, Poetry (#1584)

As there is some logic in the `do_import` action, we may want to add testing for the CLI command `import` or move this logic in the converters.


frostming (approved):

frostming:
Thanks

PR #1652, opened 2023-01-29
Set VIRTUAL_ENV in pdm run

### LINKED ISSUE #1650: pdm run does not activate the project's environment when running the command
- [X] I have searched the issue tracker and believe that this is not a duplicate.

## Steps to reproduce

```
❯ pdm run -v python -c "import sys; print(sys.executable)"
C:\Work\Scratch\Python\repro\.venv\Scripts\python.exe
❯ pdm run -v py -c "import sys; print(sys.executable)"
C:\Users\Gustav\AppData\Local\Programs\Python\Python311\python.exe
```

## Actual behavior

Note that running `python` shows the copy in the venv, but running `py` shows the system Python.

## Expected behavior

When the requested command is run, the `VIRTUAL_ENV` environment variable is set and the environment `bin`/`Scripts` directory is added to the start of `PATH`, in the same way as would happen with an activated virtual environment.

## Environment Information

```bash
❯ pdm info; pdm info --env
PDM version:
  2.4.1
Python Interpreter:
  C:\Work\Scratch\Python\repro\.venv\Scripts\python.exe (3.11)
Project Root:
  C:/Work/Scratch/Python/repro
Project Packages:
  None
{
  "implementation_version": "3.11.0",
  "os_name": "nt",
  "platform_machine": "AMD64",
  "platform_release": "10",
  "platform_system": "Windows",
  "platform_version": "10.0.22621",
  "python_full_version": "3.11.0",
  "python_version": "3.11",
  "sys_platform": "win32"
}
```

pfmoore:
## Pull Request Checklist

If the environment is a virtualenv, `pdm run` will set the `VIRTUAL_ENV` environment variable, which lets tools know that a virtualenv is in use. This means that (for example) the `py` launcher on Windows correctly locates the environment's Python rather than using the system Python.

Fixes #1650

### DIFF
--- a/news/1652.bugfix.md
+++ b/news/1652.bugfix.md
@@ -1,0 +1,1 @@
+Set VIRTUAL_ENV in `pdm run`.
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -182,6 +182,8 @@ class TaskRunner:
         )
         if project_env.packages_path:
             process_env.update({"PEP582_PACKAGES": str(project_env.packages_path)})
+        if project_env.venv_path:
+            process_env.update({"VIRTUAL_ENV": str(project_env.venv_path)})
         if env:
             process_env.update(env)
         if shell:
--- a/src/pdm/models/environment.py
+++ b/src/pdm/models/environment.py
@@ -112,6 +112,13 @@ class Environment:
         return pypackages
 
     @cached_property
+    def venv_path(self) -> Path | None:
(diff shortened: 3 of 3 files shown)


pfmoore:
I'll look into adding some tests, but I need to work out how to get a test to choose "virtualenv" mode, global mode or pypackages mode when running `pdm run` first!

frostming:
> global mode or pypackages mode when running `pdm run` first!

nvm, i will take care of it

PR #1656, opened 2023-01-31
- check requirement format all the time

franzhaas:
## Pull Request Checklist

This is not a clean ready to merge pull request. This is a proof of concept to.:

1.) deal with wheel versions which include a source control version, like a git hash.
2.) to always deal with unsoported version selectors, regardless of weather this is a necessity or not.

please give it a look. In order to make this mergeable at least.:

a proper way to inform the user about the modification of hte specifier is needed
some tests are needed


src/pdm/utils.py
```diff
@@ -26,7 +26,6 @@
 from pdm.compat import Distribution, importlib_metadata
 
 _egg_fragment_re = re.compile(r"(.*)[#&]egg=[^&]*")
-PACKAGING_22 = Version(importlib_metadata.version("packaging")) >= Version("22")
```
frostming, line 29: This variable is also used by tests

frostming:
> 1.) deal with wheel versions which include a source control version, like a git hash.

IMO this isn't a problem, local versions are not acceptable only in comparing operaters: `>=1.0.0+g12134`, but OK otherwise.

franzhaas:
The changes I initially presented here were overkill.

The problem which initially triggered me was that pdm does not work in a zipapp since the PACKAGING_22 variable gets constructed.

So my initial aproach was to get this variable out -> which works for me.

So this is a less invasive suggestion, if importlib metadata fails, lets try a different method. This is similar to the way pdm learns about its own version...

Best regards,
Franz


.

PR #1659, opened 2023-02-02
cut local versions with < or > markers

franzhaas:
## Pull Request Checklist

Since a change to packaging (version 22+) we have an issue with our local wheel versions and ">"/"<".

This is a so far only checked with unit tests. Please have a look.

Best regards,
Franz

### DIFF
--- a/src/pdm/models/requirements.py
+++ b/src/pdm/models/requirements.py
@@ -469,14 +469,17 @@ def parse_as_pkg_requirement(line: str) -> PackageRequirement:
 
     def fix_wildcard(match: Match[str]) -> str:
         operator, _, version = match.groups()
-        if ".*" not in version or operator in ("==", "!="):
-            return match.group(0)
-        version = version.replace(".*", ".0")
-        if operator in ("<", "<="):  # <4.* and <=4.* are equivalent to <4.0
-            operator = "<"
-        elif operator in (">", ">="):  # >4.* and >=4.* are equivalent to >=4.0
-            operator = ">="
-        return f"{operator}{version}"
+        rval = match.group(0)
+        if ".*" in version:
+            version = version.replace(".*", ".0")
+            if operator in ("<", "<="):  # <4.* and <=4.* are equivalent to <4.0
+                operator = "<"
+            elif operator in (">", ">="):  # >4.* and >=4.* are equivalent to >=4.0
+                operator = ">="
+            rval = f"{operator}{version}"
+        if operator in ("<", "<=", ">", ">="):
+            rval = rval.split("+")[0]
+        return rval
 
     try:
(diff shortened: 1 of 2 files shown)


frostming:
Can you show a use case of where this will cause a problem?

siobz:
I have a use case for this. We have young projects shared with other groups in the company. If they now use the newest version of pdm, it is not possible to recreate the environment. That is unpleased, after we convinced them to use pdm as a stable solution. Our files look like this:

```toml
dependencies = [
    "ipython",
    "pdoc",
    "numpy",
    "pandas",
    "ct-data-faker",
    "ams-ct-xray-raw-data-parser-analyser",
    "ams-rw-python-register-helper>=0.1.dev31+g294d274",
]
```

frostming:
`ams-rw-python-register-helper>=0.1.dev31+g294d274` would be a bit too strict. In PEP 440 [local versions are not allowed in comparison](https://peps.python.org/pep-0440/#inclusive-ordered-comparison). It would work if you change the specifier to `>=0.1.dev31`

franzhaas:
Yes this is correct. The problem is that when you work with an old state of one of our projects, you can not do so with a recently installed pdm.

It would help us if pdm would cut this local versions for us. This would allow us to reproduce old project states with current tools.

frostming:
> pdm would cut this local versions for us

We are not doing this, breaking the rule brings a risk that other tools consuming the file may not support it.

After all, if the local version is to be stripped in comparison, why do you want to keep it in the `pyproject.toml`.

frostming:
Sorry, we can't update to fit *your* workflow, you know there are thousands of workflows. Downgrade to packaging < 22 if you want to work with legacy version specifiers.

franzhaas:
I am sorry but I have to disagree.

Today I ran into this issue again and figured out how it happened.

I used a machine where pdm 2.3.2 is installed to add a package which is locally hosted with a local version by.:

`pdm add ams-rw-stdf`

the resulting pyproject.toml had this line in it and worked on that machine.:

`dependencies = ["ams-rw-stdf>=0.1.dev154+ge5720d0",]`

I think this is because the installed packaging==21.3

So from my perspective, this is the recommended workflow. While I can update PDM, we still have a lot or pyproject.toml which have been created when pdm version behaving like that were current.

I think it would be great if pdm could forgive these mistakes from the past.

frostming:
> So from my perspective, this is the recommended workflow. While I can update PDM, we still have a lot or pyproject.toml which have been created when pdm version behaving like that were current.

Yes, it was a bug existing on old PDM versions and were not surfaced until the release of packaging 22. This is subject to two attitudes towards the bugs: to tolerate or to ban. We choose to ban and that is it.

Issue #1672, opened 2023-02-04
pdm init does not accept --ignore-remembered


eltoder:
This leads to confusing behavior:
```bash
$ mkdir proj && cd proj
$ pdm init -n
Creating a pyproject.toml for PDM...
Using the last selection, add '-i' to ignore it.
You are using the PEP 582 mode, no virtualenv is created.
For more info, please visit https://peps.python.org/pep-0582/
Changes are written to pyproject.toml.
$ pdm init -n -i
Usage: pdm [-h] [-V] [-c CONFIG] [-v] [-I] [--pep582 [SHELL]]
           {add,build,cache,completion,config,export,import,info,init,install,list,lock,publish,remove,run,search,self,plugin,show,sync,update,use,venv}
           ...
pdm: error: unrecognized arguments: -i
```
I.e. pdm init suggested to use -i but failed when -i was passed.

PR #1673, opened 2023-02-04
feat: accept --ignore-remembered in pdm init

eltoder:
Closes: #1672

### DIFF
--- a/news/1672.feature.md
+++ b/news/1672.feature.md
@@ -1,0 +1,1 @@
+Always ignore remembered selection in pdm init.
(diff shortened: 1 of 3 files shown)


eltoder:
A simpler alternative is to always pass `ignore_remembered=False` in init, if this is acceptable.

frostming:
Thanks, can you modify to always ignore remembered?

eltoder:
@frostming done, please take a look

frostming:
Please also fix the test case `tests/cli/test_init.py::test_init_non_interactive`

eltoder:
Sorry, missed that. Fixed now.

PR #1687, opened 2023-02-08
doc: improve and restructure the docs

### LINKED ISSUE #1678: Improve requires-python experience

I'm having a hard time specifying a custom correct `requires-python` through `pdm init`, often times ending with an exception `[InvalidSpecifier]: Invalid specifier`. At the same time, [the documentation](https://pdm.fming.dev/latest/usage/project/#how-requires-python-controls-the-project) provides zero examples. My experience has mostly been trial and error, which isn't great.

It would be great to:
- Improve the documentation providing some examples about the most common use cases (something similar to [this](https://python-poetry.org/docs/dependency-specification/#caret-requirements)).
	- `>=`
	- Specific minor (that I managed to achieve as `>= 3.10, < 3.11`, I don't know if there's a faster way)
	- Examples about other operands supported and the version range they are resolved to
- The exception should either:
	a. Provide a correction
	b. Link to the documentation
	c. Provide a few examples

While I understand that the `requires-python` may be part of PEP 621 and so not necessarily your responsibility, it would be great to provide some documentation where needed :)

frostming:
## Pull Request Checklist

Close #1678


hynek:
- I don't think "references" is the plural of "reference articles". The section is just "Reference" or "Reference Guides" (cf. e.g. https://docs.djangoproject.com/en/4.1/). References are people that vouch for you skills when you're asking for a new jobs.
- as for the structure: have you seen <https://diataxis.fr>?

frostming:
> * I don't think "references" is the plural of "reference articles". The section is just "Reference" or "Reference Guides" (cf. e.g. https://docs.djangoproject.com/en/4.1/). References are people that vouch for you skills when you're asking for a new jobs.

Thank you, updated.



> * as for the structure: have you seen https://diataxis.fr?

Yes I have, that is a beautiful doc site. But I didn't have it in mind when I am restructuring. I just think we need a tutorial-like structure for Usage section.

pawamoy:
We should make sure to setup redirects for pages that don't exist anymore, to avoid breaking existing links to the docs. This can be done with mkdocs-redirects for example.

frostming:
> We should make sure to setup redirects for pages that don't exist anymore, to avoid breaking existing links to the docs. This can be done with mkdocs-redirects for example.

Thanks for the recommendation, #1688

PR #1721, opened 2023-02-17
Add pyright/eglot config for emacs

pakelley:
## Pull Request Checklist

- [N/A] Test cases added for changed code.

MS's python lang server (as well as lsp-python-ms) are archived in favor of pyright. Additionally, eglot will be packaged with Emacs once Emacs 29 is released.

### DIFF
--- a/docs/docs/usage/advanced.md
+++ b/docs/docs/usage/advanced.md
@@ -222,6 +222,28 @@ Below is a sample code snippet showing how to make PDM work with [lsp-python-ms]
                (lsp))))  ; or lsp-deferred
 ```
 
+## Work with pyright and eglot in Emacs
+
+Using [pyright](https://github.com/microsoft/pyright) and [eglot](https://github.com/joaotavora/eglot) (which is included in Emacs 29), add the following to your config:
+
+```emacs-lisp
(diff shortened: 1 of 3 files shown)


pakelley:
Adding a news fragment now...

pakelley:
Also, I'm curious why Emacs's config is in a different section than all of the other editors? If it's arbitrary I'd be happy to move it.

pakelley:
Sorry for the noise, but the config listed for Neovim works for Emacs too. If it seems helpful, I could add a note about doing it that way to the Emacs section, since it may be simpler in some cases.

frostming:
@pakelley I am not an expert of emacs and I have no objections about the changes.

pakelley:
Sounds good, well if you don't mind I'll move the emacs section next to the other editors and add a section about configuring via `pyproject.toml`.
Thanks for all of your work on pdm, it's the first python package management tool I've used that isn't painful :)

pakelley:
Alright, as long as you're happy with this, I've made all of the changes I wanted. Let me know if there's anything you'd like updated.

frostming (approved):
LGTM

PR #1735, opened 2023-02-23
fix: in project conda venv activate without argument

ProgramRipper:
## Pull Request Checklist

~~- [ ] Test cases added for changed code.~~ (current tests do not cover `venv activate` without `env`)

For in project venv created by `conda`, using `pdm venv activate` without specifying `env_name` would be a failure:

```bash
(base) programripper:~/conda-test$ echo $CONDA_PREFIX  # Actually it is quite common for conda that isn't installed normally (e.g. via pyenv) that not to have CONDA_PREFIX properly set, which will make pdm fail faster.
/home/programripper/.pyenv/shims/conda
(base) programripper:~/conda-test$ pdm venv list
Virtualenvs created with this project:

*  in-project: /home/programripper/conda-test/.venv
(base) programripper:~/conda-test$ cat .pdm.toml
[venv]
backend = "conda"

[python]
path = "/home/programripper/conda-test/.venv/bin/python"
(base) programripper:~/conda-test$ pdm venv activate
Can't activate a non-venv Python /home/programripper/conda-test/.venv/bin/python, you can specify one with pdm venv activate <env_name>
```

By this PR, user can activate in project venv created by `conda` using `pdm venv activate` as venv created by `venv` or `virtualvenv`.

### DIFF
--- a/src/pdm/utils.py
+++ b/src/pdm/utils.py
@@ -281,19 +281,17 @@ def get_venv_like_prefix(interpreter: str | Path) -> Path | None:
     and return the prefix if found.
     """
     interpreter = Path(interpreter)
-    prefix = interpreter.parent.parent
-    if prefix.joinpath("pyvenv.cfg").exists():
+    prefix = interpreter.parent
+    if prefix.joinpath("conda-meta").exists():
+        return prefix
+
+    prefix = prefix.parent
+    if prefix.joinpath("pyvenv.cfg").exists() or prefix.joinpath("conda-meta").exists():
         return prefix
 
     virtual_env = os.getenv("VIRTUAL_ENV", os.getenv("CONDA_PREFIX"))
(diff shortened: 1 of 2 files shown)


src/pdm/utils.py
```diff
@@ -288,6 +288,12 @@ def get_venv_like_prefix(interpreter: str | Path) -> Path | None:
     virtual_env = os.getenv("VIRTUAL_ENV", os.getenv("CONDA_PREFIX"))
     if virtual_env and is_path_relative_to(interpreter, virtual_env):
         return Path(virtual_env)
+
+    with contextlib.suppress(FileNotFoundError, subprocess.CalledProcessError):
+        envs = json.loads(subprocess.check_output(["conda", "env", "list", "--json"]))["envs"]
```
frostming, line 293: Can we achieve this by only looking at some files?

frostming (approved):

PR #1748, opened 2023-03-03
fix: sync failure when dynamic version is rewritten in pre_install hook

leventeliu:
## Pull Request Checklist

### The issue

I was using some other tool in hooks to create customized versioning during package publishing.  Then I noticed that when the `pre_install` hook changes the package version, the `sync` command reports an error.

Reproduce can be done in the new fixture project under `tests/fixtures/projects/test-pre-install-version-rewriting`:

```shell
$ cd tests/fixtures/projects/test-pre-install-version-rewriting
$ pdm sync -G:all --clean
All packages are synced to date, nothing to do.
Installing the project as an editable package...
  ✖ Install test-pre-install-version-rewriting 0.0.1 failed

See /tmp/pdm-install-u05bdw5w.log for detailed debug log.
[KeyError]: "There is no item named 'test_pre_install_version_rewriting-0.0.1.dist-info/RECORD' in the archive"
```

I will not bring the debug log here cause it's not quite helpful.

### Root cause

The root cause is simple: `make_self_candidate` in the synchronizer class constructor happens before `pre_install` emitting so that the candidate version and the built package version are mismatched. And the fix is simple too: consider the self-candidate resolving as part of the sync/install process and move it after `pre_install` emitting.

### Discussion

I am not very sure whether the hooks are designed to do this. And I am aware that the package version is mostly irrelevant for an editable package. But I instinctively think always trying to keep the version number up to date is some good practice when I add these hooks. Hence this fixing PR.


frostming:
There are some refactoring around the Synchronizer class in the `dev` branch, it should be fixed there, can you verify it?

leventeliu:
@frostming Sure. There are a lot of diffs but I can confirm that the dev version can pass my test case. I will close this and wait for your refactored release. Thanks :)

PR #1751, opened 2023-03-04
feat: Add `.pdm-python` file to project root `.gitignore`

### LINKED ISSUE #1749: Include .pdm.toml in root .gitignore when running `pdm init`

When running `pdm --init` add a root level `.gitignore` and write the `.pdm.toml` path to it. Alternatively, if a `.gitignore` already exists append the path to the file.

This would be similar to how `pdm --init` created `__pypackages__/.gitignore`, but for the `.pdm.toml` file. (#446)

This would save people a bit of extra time reading the docs (#1364) to know that you don't want to commit this file.

The `pdm --init` creates (or updates) the project root's `.gitignore` to include the `.pdm.toml` path.

---

If this is accepted I would be happy to open a PR to implement this.

Nadock:
## Pull Request Checklist

When running `pdm init` the `.gitignore` file at the root of the project will be updated to include the `.pdm-python` file which is not meant to be committed to git. If the project has no `.gitignore` one will be created, and then the ignore rule will be appended to the file.

Apologies for the lack of test case for the change, I couldn't see any prexisting tests for the file create logic in `pdm init`. 

Closes #1749

### DIFF
--- a/news/1749.feature.md
+++ b/news/1749.feature.md
@@ -1,0 +1,1 @@
+Include `.pdm-python` in project root `.gitignore` when running `pdm init`.
--- a/src/pdm/cli/actions.py
+++ b/src/pdm/cli/actions.py
@@ -514,6 +514,8 @@ def do_init(
             readme = project.root.joinpath("README.md")
             readme.write_text(f"# {name}\n\n{description}\n", encoding="utf-8")
         data["project"]["readme"] = readme.name
+    with project.root.joinpath(".gitignore").open("a", encoding="utf-8") as file:
+        file.write(".pdm-python")
     get_specifier(python_requires)
     project.pyproject._data.update(data)
     project.pyproject.write()


frostming:
It would be better to create a useful `.gitignore` if it doesn't exist or update the existing one. Minimal required entries would be:
```
__pycache__/
dist/
build/
*.egg-info/
.pdm-python
```

Nadock:
> It would be better to create a useful .gitignore if it doesn't exist or update the existing one.

Agreed, looking at the [GitHub default .gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore) and removing entries that are for other tools, I get a file that looks like this:

```.gitignore
# Byte-compiled / optimized / DLL files
__pycache__/
*.py[cod]
*$py.class

# C extensions
*.so

# Distribution / packaging
.Python
build/
develop-eggs/
dist/
downloads/
eggs/
.eggs/
lib/
lib64/
parts/
sdist/
var/
wheels/
share/python-wheels/
*.egg-info/
.installed.cfg
*.egg
MANIFEST

# PDM specific ignores, see https://pdm.fming.dev/#use-with-ide for more information.
.pdm-python

# PEP 582
__pypackages__/

# Environments
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/
```

Thoughts on if we should add anything back or remove something? Also, in the scenario where a `.gitignore` already exists do we want to try and merge the two files (avoiding duplicates) or just add the minimal PDM specific stuff?

Nadock:
Are we trying to create a useful `.gitignore` or just add some PDM specific bits? 

Personally, I'd prefer the latter but in that case `.pdm-python` is really the only file that is specific to PDM.

frostming:
> Are we trying to create a useful `.gitignore` or just add some PDM specific bits?

For initialization, it's better to add a useful `.gitignore`.

frostming:
@Nadock Hi, I've merged the dev branch into main, would you like to continue this PR? Or if you wouldn't it's okay I can do the rest.

Nadock:
> @Nadock Hi, I've merged the dev branch into main, would you like to continue this PR? Or if you wouldn't it's okay I can do the rest.

Yeah thanks @frostming, I've been unwell for a little while so if you want to finish this up go ahead. Otherwise I'll try to get back to this when I'm feeling better.

frostming:
Superseded by #1785, thanks

PR #1855, opened 2023-04-26
feat(scripts): hide tasks starting with an underscore from listing

noirbizarre:
## Pull Request Checklist

This PR is considering the tasks with a name starting with an underscore as internal tasks (helpers for composite...) and hide them from the listing.


frostming (approved):
Good

PR #1893, opened 2023-05-06
feat: install plugins from project config

### LINKED ISSUE #1461: Install plugins from pyproject.toml (+config)

Hello!

To solve some workflow issues with torch I've developed a plugin that I'm storing in the same repository as the main project. I want all developers to have access to this tool for ease of use. However; there isn't any (documented?) way of putting a plugin into pyproject.toml, including configuration. I tried putting it in dev-dependencies, but since `pdm` runs in another environment it doesn't find the entrypoint.

I'd to be able to specify plugins and their configuration in the project.toml. The config part is roughly how I've done it today; but I'm just digging out the key from the `Project.pyproject` data. The first part I don't think exists at all.

```toml
[tool.pdm]
plugins = [
    "pdm_plugin_torch @ file:///${PROJECT_ROOT}/plugins/pdm-plugin-torch",
]

[tool.pdm.plugin.pdm_plugin_torch]
torch-version = "torch==1.10.2"
lockfile = "torch.lock"
enable-cpu = true

enable-rocm = true
rocm-versions = ["4.2"]

enable-cuda = true
cuda-versions = ["cu111", "cu113"]
```

A potential format version that's a bit more structured:

```toml
[[tool.pdm.plugins]]
dependency = "pdm_plugin_torch @ file:///${PROJECT_ROOT}/plugins/pdm-plugin-torch"

[tool.pdm.plugins.configuration]
torch-version = "torch==1.10.2"
lockfile = "torch.lock"
enable-cpu = true

enable-rocm = true
rocm-versions = ["4.2"]

enable-cuda = true
cuda-versions = ["cu111", "cu113"]
```

frostming:

## Pull Request Checklist

Resolve #1461

This adds a config field `tool.pdm.plugins` as a list of plugin requirements. When user runs `pdm install --plugins`, the dependencies will be installed to a project plugin library under the project root, which will be gitignored by default.

Then, the following invocations of PDM will add the path to `sys.path` and load the plugins in it.


tgolsson:
I've personally never installed a cargo plugin as a dev dependency and I wouldn't expect that to work as you can normally only depend on libs. 

Thus, I do think more specific naming would be good. What about `tool.pdm.plugins` instead? Since it's already in a tool namespace it seems like treating it specially all the way makes more sense given it'd be a very special case of dev dependency. Same thing with install, I'd specialcase it there. `pdm install --plugins`? `pdm init --plugins`? 

In essence: does the specialness of the plugins warrant more special workflows, as opposed to overloading existing commands with "similar" functionality.

If keeping it as a dev-dep group I'd definitely call it `pdm-plugins` as it's very likely to collide with existing use otherwise.

pawamoy:
I agree with @tgolsson. I rely on a PDM plugin locally but not in CI, and therefore would prefer that it does not get installed by default.

tgolsson:
I don't particularlymind if they get automatically installed though. I was just positing that they are a conceptually third kind of dependency. :-)

frostming:
I put them in dev-dependencies because I also want the plugins to be locked while reusing the lock ability. However I agree that plugins should not be install by default but should be requested explicitly.

tgolsson:
That's an aspect I hadn't considered, but I think it could be argued both ways. Pants doesn't, and I haven't felt any major issues working with it so far - but I also always do strict pinning Though that only solves some issues, not all. Presumably it could be retrofitted with locking capabilities if it turns out to not work well enough?

PR #1915, opened 2023-05-11
Fix a number of `ResourceWarning`s when running the test suite

falbrechtskirchinger:
## Pull Request Checklist

Running the test suite with warnings enabled generates a large number of warnings because of open file descriptors. This PR closes the two offending file handles and also a few others found in the same code files. ~~Two~~ Four warnings remain, one of which is outside this project's control.


pawamoy (approved):
Nice!

PR #1924, opened 2023-05-14
Document ability to install deps from a plain URL

sophiabits:
## Pull Request Checklist

Small edit to the docs explicitly indicating that you can install deps straight from a URL. Reading the section on VCS deps it wasn't obvious to me that any HTTP(s) url works fine with pdm.


frostming:
Thanks 🎉

Issue #1968, opened 2023-06-01
Incorrect file path after migrating Poetry config 


dbalabka:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

## Steps to reproduce

```yaml
[tool.poetry]
name = "someproject"
version = "0.1.0"
description = ""
authors = ["Dmitry Balabka <dmitry.balabka@gmail.com>"]
license = "proprietary"
packages = [
    { include = "somefile.py", from = "src/" }
]
[tool.poetry.dependencies]
python = "^3.7"
```

## Actual behavior

`pdm init` will produce:
```
[tool.pdm.build]
includes = [
    "src//somefile.py"
]
```

## Expected behavior

I'm expecting to see:
```
[tool.pdm.build]
includes = [
    "src/somefile.py"
]
```

## Environment Information

```bash
❯ pdm info && pdm info --env
PDM version:
  2.7.0
Python Interpreter:
  /home/torinaki/src/project/.venv/bin/python (3.10)
Project Root:
  /home/torinaki/src/project
Local Packages:

{
  "implementation_name": "cpython",
  "implementation_version": "3.10.11",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.15.90.1-microsoft-standard-WSL2",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Fri Jan 27 02:56:13 UTC 2023",
  "python_full_version": "3.10.11",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "linux"
}

```

PR #1969, opened 2023-06-01
Fix path while import poetry deps (#1968)

dbalabka:
## Pull Request Checklist

### DIFF
--- a/src/pdm/formats/poetry.py
+++ b/src/pdm/formats/poetry.py
@@ -156,7 +156,7 @@ class PoetryMetaConverter(MetaConverter):
             assert isinstance(item, dict)
             include = item["include"]
             if item.get("from"):
-                include = f"{item.get('from')}/{include}"
+                include = str(Path(str(item.get("from"))).joinpath(include))
             includes.append(include)
         for item in source.pop("include", []):
             if not isinstance(item, dict):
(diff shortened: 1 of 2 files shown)


src/pdm/formats/poetry.py
```diff
@@ -156,7 +156,7 @@ def includes(self, source: dict[str, list[str] | str]) -> list[str]:
             assert isinstance(item, dict)
             include = item["include"]
             if item.get("from"):
-                include = f"{item.get('from')}/{include}"
+                include = str(Path(str(item.get("from"))).joinpath(include))
```
frostming, line 159:
```suggestion
                include = str(Path(item.get("from"), include))
```

frostming (approved):

src/pdm/formats/poetry.py
```diff
@@ -156,7 +156,7 @@ def includes(self, source: dict[str, list[str] | str]) -> list[str]:
             assert isinstance(item, dict)
             include = item["include"]
             if item.get("from"):
-                include = f"{item.get('from')}/{include}"
+                include = Path(item.get("from"), include).as_posix()
```
frostming, line 159:
```suggestion
                include = Path(str(item.get("from")), include).as_posix()
```

PR #2002, opened 2023-06-12
fix: install-pdm.py looking in wrong place for `yanked` flag

kwigley:
## Pull Request Checklist


frostming:
Thanks!

Issue #1988, opened 2023-06-08
Deprecation: nesting argument groups


pawamoy:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

Nesting argument groups seems to be deprecated, starting at Python 3.11.

## Steps to reproduce

Simply run PDM with deprecation warnings enabled.

## Actual behavior

From pytest output:

```
  ___________________________________________________________________________________________ test_multirun ____________________________________________________________________________________________
  [gw0] linux -- Python 3.11.0 /home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.11.0/bin/python3.11

      def test_multirun() -> None:
          """Basic run."""
  >       main(["multirun", "python", "-V"])

  tests/test_plugin.py:8: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  __pypackages__/3.11/lib/pdm/core.py:289: in main
      return Core().main(args)
  __pypackages__/3.11/lib/pdm/core.py:57: in __init__
      self.init_parser()
  __pypackages__/3.11/lib/pdm/core.py:93: in init_parser
      self.register_command(klass, klass.name or name)
  __pypackages__/3.11/lib/pdm/core.py:232: in register_command
      command.register_to(self.subparsers, name)
  __pypackages__/3.11/lib/pdm/cli/commands/base.py:44: in register_to
      command = cls(parser)
  __pypackages__/3.11/lib/pdm/cli/commands/base.py:25: in __init__
      self.add_arguments(parser)
  __pypackages__/3.11/lib/pdm/cli/commands/run.py:354: in add_arguments
      exec = action.add_argument_group("execution", "Execution parameters")
[...]

PR #2004, opened 2023-06-12
Use `PYTHONDEVMODE=1` for CI

sanmai-NL:
See https://docs.python.org/3/library/devmode.html. This helps to detect issues like #1988 earlier.

## Pull Request Checklist

The title and commit message say it.

I‘m using this setting in CI pipelines everywhere and I have only good experiences with it.

(Please explain whether the two checklist items need completion in this case.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,8 @@ concurrency:
 
 jobs:
   Testing:
+    env:
+      PYTHONDEVMODE: 1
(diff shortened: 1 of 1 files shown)


sanmai-NL:
@frostming Based on the CI output for the test, unless I missed something, `PYTHONDEVMODE=1` wouldn't have caught #1988. Only if you test PDM by invoking the CLI, would it have caught such a problem.

PR #2012, opened 2023-06-13
Refactor deprecated `warn` method and use appropriate warnings class

sanmai-NL:
## Pull Request Checklist

### DIFF
--- a/src/pdm/utils.py
+++ b/src/pdm/utils.py
@@ -413,8 +413,8 @@ def deprecation_warning(message: str, stacklevel: int = 1, raise_since: str | No
 
     if raise_since is not None:
         if Version(__version__) >= Version(raise_since):
-            raise DeprecationWarning(message)
-    warnings.warn(message, DeprecationWarning, stacklevel=stacklevel + 1)
+            raise FutureWarning(message)
+    warnings.warn(message, FutureWarning, stacklevel=stacklevel + 1)
 
 
 def is_pip_compatible_with_python(python_version: Version | str) -> bool:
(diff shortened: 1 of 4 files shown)


sanmai-NL:
@frostming May we simplify more of the logging and warnings use with https://docs.python.org/3/library/logging.html#logging.captureWarnings?

E.g., at [termui.logger.warning(message)](https://github.com/sanmai-NL/pdm/blob/e738c3769eb048257e881ba2b5337991b344ecb9/src/pdm/models/candidates.py#L550-L551)?

src/pdm/utils.py
```diff
@@ -413,8 +413,8 @@ def deprecation_warning(message: str, stacklevel: int = 1, raise_since: str | No
 
     if raise_since is not None:
         if Version(__version__) >= Version(raise_since):
-            raise DeprecationWarning(message)
-    warnings.warn(message, DeprecationWarning, stacklevel=stacklevel + 1)
+            raise FutureWarning(message)
+    warnings.warn(message, FutureWarning, stacklevel=stacklevel + 1)
```
pawamoy, line 417: Why future warnings instead of deprecation warnings? The function  is called `deprecation_warning`. The test function as well.

frostming: Found this: https://stackoverflow.com/questions/55377831/difference-between-deprecationwarning-pendingdeprecationwarning-and-futurewarni

TIL

sanmai-NL: @pawamoy This was motivated in the commit message.

pawamoy: @sanmai-NL I missed your commit message indeed, my bad. I agree with not changing the function name as to avoid a breaking change.

frostming:
> * A news fragment is added in `news/` describing what is new.

Can you please add a news fragment?

PR #2046, opened 2023-06-21
fix: Don't try to replace shebang in binary scripts

### LINKED ISSUE #2045: PDM corrupts binary executable (ruff)

## Steps to reproduce

```bash
# setup
mkdir pdm-ruff-repro
cd pdm-ruff-repro
cat >pyproject.toml <<EOF
[project]
name = ""
version = ""
description = ""
authors = [
    {name = "Hi PDM", email = "hi@pdm.com"},
]
dependencies = [
    "ruff>=0.0.274",
]
requires-python = ">=3.7"
license = {text = "MIT"}
EOF
touch test.py
pdm config --local python.use_venv false
pdm install

# assert that it initially works:
pdm run ruff check test.py  # OK
cp __pypackages__/3.7/bin/ruff ruff37  # backup executable

# now switch Python version and back:
pdm use -f python3.8
pdm use -f python3.7

# assert executable was changed
diff __pypackages__/3.7/bin/ruff ruff37
# Binary files __pypackages__/3.7/bin/ruff and ruff37 differ

# assert ruff does not work anymore
pdm run ruff check test.py  # no output, failure
__pypackages__/3.7/bin/ruff check test.py
# [1]    178079 segmentation fault (core dumped)  __pypackages__/3.7/bin/ruff check test.py
```

## Actual behavior

PDM corrupts binary executable scripts, probably when trying to replace the shebang.

## Expected behavior

Would it make sense to only try replacing the shebang on non-binary files?
Maybe by trying to open the file as text (utf8) and if it fails, do nothing? Otherwise continue with the regex replace?

## Environment Information

```bash
PDM version:
  2.7.4
Python Interpreter:
  /home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.7.15/bin/python3.7 (3.7)
Project Root:
  /home/pawamoy/pdm-ruff-repro
Local Packages:
  /home/pawamoy/pdm-ruff-repro/__pypackages__/3.7
{
  "implementation_version": "3.7.15",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.3.8-arch1-1",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT_DYNAMIC Wed, 14 Jun 2023 20:10:31 +0000",
  "python_full_version": "3.7.15",
  "python_version": "3.7",
  "sys_platform": "linux"
}
```

pawamoy:
## Pull Request Checklist

Closes #2045.

### DIFF
--- a/news/2045.bugfix.md
+++ b/news/2045.bugfix.md
@@ -1,0 +1,1 @@
+Don't try to replace shebang in binary scripts.
--- a/src/pdm/environments/local.py
+++ b/src/pdm/environments/local.py
@@ -14,7 +14,7 @@ if TYPE_CHECKING:
     pass
 
 
-def _get_shebang_path(executable: str, is_launcher: bool) -> bytes:
+def _get_shebang_path(executable: str, is_launcher: bool) -> str:
     """Get the interpreter path in the shebang line
 
     The launcher can just use the command as-is.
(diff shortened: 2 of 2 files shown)


pawamoy:
I tried this fix locally and it works, the Ruff binary doesn't get corrupted anymore.

frostming:
haha, you probably don't have windows in mind

frostming:
do you know why is ruff binary broken when we are only matching the shebang line `#!...`? we even match it on its own line only

pawamoy:
I don't. The binary is generated by Maturin I guess, maybe there are places in the binary that have such a string, but for other reasons.

Would it make sense to:

1. remove the multiline flag for the shebang regex?
2. search only on the first line(s) for both regex?

pawamoy:
```pycon
>>> import re
>>> r = rb"^#!(.+?)\s*$"
>>> from pathlib import Path
>>> ruff = Path("__pypackages__/3.8/bin/ruff").read_bytes()
>>> m = re.search(r, ruff, flags=re.M)
>>> m
<re.Match object; span=(10819556, 10819578), match=b'#!/usr/bin/env python3'>
```

First matching string in ruff appears at position 10819556 for example.

frostming:
> I know nothing about Python scripts (.exe) on Windows, but maybe if their shebang appears *first* in the file, we could somehow restrict the regex search, to avoid matching a string later in the file.

it doesn't, unfortunately.

frostming:
@pawamoy maybe we can check if the content is zip decompressed?

PR #2095, opened 2023-07-09
chore(ruff): use isort and extend-select

aarnphm:
## Pull Request Checklist


pyproject.toml
```diff
@@ -118,7 +118,8 @@ target-version = ["py37", "py38", "py39", "py310"]
 
 [tool.ruff]
 line-length = 120
-select = [
+extend-select = [
```
vvanglro, line 121: This already contains the default E and F rules, so extend-select should not be used.

frostming (approved):

PR #2103, opened 2023-07-14
doc: improve help info format for subcommands

leemars:
## Pull Request Checklist

Improve help info format for subcommands. 

* Set `metavar` to empty for subparsers
* Remove extra line breaks and indent block when action_help is empty
* Remove extra continuous spaces in usage info

By the way, it also fixed the help info difference for older Python version.

### DIFF
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -78,12 +78,11 @@ class Core:
             "--config",
             help="Specify another config file path [env var: PDM_CONFIG_FILE] ",
         )
-        self.parser._positionals.title = "Commands"
         verbose_option.add_to_parser(self.parser)
         ignore_python_option.add_to_parser(self.parser)
         pep582_option.add_to_parser(self.parser)
 
-        self.subparsers = self.parser.add_subparsers(parser_class=ArgumentParser, dest="fooo")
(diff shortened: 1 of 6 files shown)


src/pdm/core.py
```diff
@@ -78,12 +78,11 @@ def init_parser(self) -> None:
             "--config",
             help="Specify another config file path [env var: PDM_CONFIG_FILE] ",
         )
-        self.parser._positionals.title = "Commands"
         verbose_option.add_to_parser(self.parser)
         ignore_python_option.add_to_parser(self.parser)
         pep582_option.add_to_parser(self.parser)
 
-        self.subparsers = self.parser.add_subparsers(parser_class=ArgumentParser, dest="fooo")
+        self.subparsers = self.parser.add_subparsers(parser_class=ArgumentParser, title="commands", metavar="")
```
frostming, line 85: commands is the default title?

leemars: Yes.

In previous revision, the 81L `self.parser._positionals.title` was set to `Commands`, force argparse displays "Commands:" as title instead of "Positional Arguments:". But it can be done by setting subparsers title to "commands", as same as other venv/self/cache subparsers, without touching the internal members.

frostming (approved):

Issue #2126, opened 2023-07-23
Improve output when importing pyproject.toml from poetry


quag-cactus:
## Is your feature request related to a problem? Please describe.

I am attempting to convert the pyproject.toml file, which was originally written for Poetry.

Here is an example of the toml file used to manage a test project with Poetry:

<details>
<summary>original toml file</summary>

```toml
name = "test-migration-project"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"
packages = [{include = "test_migration_project"}]

[tool.poetry.dependencies]
python = "^3.8"
pendulum = "^2.1.2"
tqdm = "^4.65.0"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

</details>

following toml file is generated by executing `pdm import pyproject.toml`:

<details>
<summary>outputted toml file</summary>

```toml
[tool.poetry]
name = "test-migration-project"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"
packages = [{include = "test_migration_project"}]

[tool.poetry.dependencies]
python = "^3.8"
pendulum = "^2.1.2"
tqdm = "^4.65.0"

[tool.pdm.build]
includes = [
    "test_migration_project",
]
[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"

[project]
authors = [
    {name = "Your Name", email = "you@example.com"},
]
requires-python = ">=3.8,<4.0"
dependencies = [
    "pendulum<3.0.0,>=2.1.2",
    "tqdm<5.0.0,>=4.65.0",
]
name = "test-migration-project"
version = "0.1.0"
description = ""
readme = "README.md"
```

</details>

## Describe the solution you'd like

* Existing sections `[tool.poetry.*]`are retained. There does not seem to be much need for these to be do.
* The order of the sections inside pyproject.toml generated by the import command seem to be undefined. The order of the sections should be in the same order as when the new project is generated.
* The value of the `build-system` section is still specifying the poetry backend. It should be updated to a backend supported by PDM.

If this proposal is needed, I am prepared to work on this. 🙂

frostming:
Thanks for that. Welcome to contribute.

frostming:
> * The value of the `build-system` section is still specifying the poetry backend. It should be updated to a backend supported by PDM.

This item is already resolved, and there is no more item we could act on here. Closing now

PR #2172, opened 2023-08-06
feat: Remove unused section from converted pyptoject.toml

quag-cactus:
## Pull Request Checklist

This PR is part of the solution to problem #2126.
It is intended to solve the following problem.

 * Existing sections `[tool.poetry.*]` are retained. there does not seem to be much need for these to be done.

These codes don't allow sections other than "pdm" to exist at the level below `[tool.*]` in the converted pyproject.toml.

### DIFF
--- a/news/2126.feature.md
+++ b/news/2126.feature.md
@@ -1,0 +1,1 @@
+Add code for remove unused toml section to conversion process for import command.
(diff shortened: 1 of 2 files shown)


frostming:
This is not acceptable because:

1. There might be configurations of various tools, not only poetry, that users want to keep. You may think of them as unneeded but others may be surprised if you remove them.
2. The import process doesn't support all features of poetry, some configurations may be missing and some may be converted incorrectly. Leave the original section for manually check, and remove them later.

I will close this for now.

quag-cactus:
I undarstand. I'll delete this branch for now. Thank you for your checking!

PR #2200, opened 2023-08-23
fix: Switch to Truststore by default

### LINKED ISSUE #2195: Switch to Truststore by default
Hello! Wanted to thank you for your work on the PDM project. Recently PDM adopted truststore as an alternate root CA mechanism to certifi. I've [noticed this through downloads](https://twitter.com/sethmlarson/status/1687845307558600704) and dependency data and have been monitoring your issue tracker for a few weeks now waiting for potential issues to be found in truststore. So far I haven't seen any.

This is a good sign, I am so appreciative of taking this step and getting a large representative sample of users using Truststore. I would like to ask if you'd be interested in taking the next step which is using Truststore by default for all installations (Currently Truststore is being used by ~10% of your users, if we go by downloads alone).

This information is important because I would like to drive adoption of Truststore and mechanisms like it for other projects like pip, requests, and Python itself. Having this proof that Truststore works without issue for large populations and important projects is invaluable in convincing other projects to also adopt Truststore-like CA mechanisms.

- Add Truststore as a default dependency for Python 3.10+
- Continue using Truststore by default

Let me know if you have questions or concerns, and again thank you so much for everything you've already done. :raised_hands:

frostming:
Fixes #2195

## Pull Request Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "resolvelib>=1.0.1",
     "installer<0.8,>=0.7",
     "cachecontrol[filecache]>=0.13.0",
+    "truststore; python_version >= \"3.10\"",
     "tomli>=1.1.0; python_version < \"3.11\"",
     "importlib-resources>=5; python_version < \"3.9\"",
     "importlib-metadata>=3.6; python_version < \"3.10\"",
@@ -59,9 +60,8 @@ keyring = ["keyring"]
 template = [
     "pdm[copier,cookiecutter]",
 ]
-truststore = ["truststore; python_version >= \"3.10\""]
 all = [
-    "pdm[keyring,template,truststore]",
+    "pdm[keyring,template]",
 ]
 
 [project.scripts]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -59,9 +60,8 @@ keyring = ["keyring"]
 template = [
     "pdm[copier,cookiecutter]",
 ]
-truststore = ["truststore; python_version >= \"3.10\""]
```
sethmlarson, line 62: I don't think the `[truststore]` extra was used much so this shouldn't be a huge problem (likely many people were getting truststore via `pdm[all]`) so this should be fine? The `pdm[truststore]` extra isn't in the docs, but there is a separate section on installing `truststore` which may not be needed anymore now that it's a default?

sethmlarson (changes requested):

pyproject.toml
frostming: Requesting for non-existing extra names will result in a warning. So that is okay.

PR #2214, opened 2023-08-30
Make `Config` kwargs optional.

drunkwcodes:
Make Config kwargs optional. reference: project/core.py L135: Config(self.root / 'pdm.toml')

It is for the convenience to init a config object.

## Pull Request Checklist

### DIFF
--- a/src/pdm/project/config.py
+++ b/src/pdm/project/config.py
@@ -14,6 +14,7 @@ from pdm import termui
 from pdm._types import RepositoryConfig
 from pdm.compat import cached_property
 from pdm.exceptions import NoConfigError, PdmUsageError
+from pdm.utils import find_project_root
 
 REPOSITORY = "repository"
 SOURCE = "pypi"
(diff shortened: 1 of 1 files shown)


drunkwcodes:
The test failure is not related to this PR.

The return type of `def find_project_root(cwd: str = ".", max_depth: int = 5) -> str | None:` is bothersome.
It is not comply with `os.PathLike`. It can not be used with `Path(find_project_root())`, not convenient.

I will also make a PR for `find_project_root()`.

frostming:
This change makes the logic convoluted, config object shouldn't know how project is found, but project should know how to get a config.

drunkwcodes:
I think the project root is static as a file, and `find_project_root()` is in the `utils.py`, not a huge burden to Config class.
Is it ok?

Actually, I put `config.py` and `project_file.py` in the `models/` in mojo_muse now.
And I refactored those which depends `Project` class in the `project/`.

https://github.com/drunkwcodes/mojo_muse/tree/main/src/mojo_muse/project

frostming:
I still insists, although `find_project_root` is a static function, it is a project-level knowledge, moreover, `find_project_root` takes a `project_max_depth` config value.

drunkwcodes:
OK, if you insist. Nonetheless, thank you for your instant response.

I will make next PR right a way.
Please re-consider my refactoring effort again.

frostming:
Thanks.

drunkwcodes:
@frostming Please look at my WIP unit test: https://github.com/drunkwcodes/mojo_muse/blob/main/tests/unit/models/test_config.py

I can init a Config object without knowing exact config file path now. Because I don't need to write it.
It is more convenient for calling and external use, I think.

frostming:
You can start from a Project object and get config from it. Regard project as the entry point of the package manager

drunkwcodes:
Thanks for the guide.

But I'm refining every aspect of the code base now.
I will let you know if fruitful result is gained in the near future.

PR #2243, opened 2023-09-07
Change default value of `find_project_root()`

drunkwcodes:
## Pull Request Checklist

Currently, if `find_project_root()` fails to find a root, it returns `None`.
It is inconvenient for `Path(find_project_root())`, which needs `os.PathLike` str instead of Optional[str].

`None` is not welcome in here.

So I change the function to have a `default` kwarg that will be returned when the root was not found.


frostming:
This changes the behavior, see the CI failure. Sometimes we do want to know a project root can't be found. To make the type check, there are two options:

1. Raise an error instead of return None, so that we can know the project root can't be found. And in certain cases you can assume it always succeeds.
2. Make two signatures with `typing.overload`, when default isn't given it is the same as the current behavior, returning `str | None`, and if default is a string value, it will always return a `str`.

drunkwcodes:
@frostming Thanks for your quick reply.

I have a solution now.
It can tell us the root is not found by assuring return value is Path(".") and not absolute.

If it finds root, it will be absolute path. There is a subtle difference.

frostming:
> @frostming Thanks for your quick reply.
> 
> I have a solution now. It can tell us the root is not found by assuring return value is Path(".") and not absolute.
> 
> If it finds root, it will be absolute path. There is a subtle difference.

We need a more obvious difference to distinguish the two cases. `str | None`, or `success | exception`

drunkwcodes:
@frostming If so, I think `None` is better. Current function is good.

But we can gain the `Path()` compatibility by this PR.
I fixed the Project initialization.
It works perfect now.

Is the obvious difference really needed?

frostming:
Comparing to `Path(".")` to detect a failing path is not good to me.
Why not take either option I mentioned above, it should solve your problem.

drunkwcodes:
@frostming As I mentioned, `None` is not compatible with `Path()`.
And an exception is exaggerating. 

So I decide to leave the function as it is.
But I kept my branch in my fork.
If you change your mind in some day, I will re-open this PR.

frostming:
@drunkwcodes You may not understand my suggestion:

```python
def find_project_root(cwd: str = ".", max_depth: int = 5, default: str | None = None) -> str | None:
     ...
```

When `default` is not given, it is the same as the current behavior, but when it is given a string, `Path(find_project_root(default="."))` is just happy with it. You can even achieve this by `Path(find_project_root() or ".")`

I don't think there is any other problem with it.

drunkwcodes:
OK, it sounds promising.

I will take some time to make this change,
see you next time!

drunkwcodes:
But I think mypy will not happy with this. 
Fine. There is no `Path(find_project_root())` usage.

frostming:
> But I think mypy will not happy with this.

It will, use `@typing.overload`

drunkwcodes:
@frostming I'm not experienced with `overload`.

Please add that for me.

Do you mean:

```python
from typing import overload

@overload
def find_project_root(cwd: str = ".", max_depth: int = 5, default: str) -> str:
     ...
```

frostming:
```
from typing import overload

@overload
def find_project_root(cwd: str = ".", max_depth: int = 5, default: None = ...) -> str | None:
     ...

@overload
def find_project_root(cwd: str = ".", max_depth: int = 5, default: str = ...) -> str:
     ...

def find_project_root(cwd: str = ".", max_depth: int = 5, default: str | None = None) -> str | None:
    # the implementation goes here
```

But I wonder why you are not fine with `Path(find_project_root() or ".")` ?

frostming:
Okay, mypy isn't happy, can you answer this question?

> But I wonder why you are not fine with `Path(find_project_root() or ".")` ?

drunkwcodes:
@frostming I didn't think about that.

And mypy complains: `src/pdm/utils.py:79: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]
Found 1 error in 1 file (checked 108 source files)`.

So I think your workaround is the best answer. I'm going to close this PR.

PR #2286, opened 2023-09-28
refactor: Recursively look for the project root

xulongwu4:
## Pull Request Checklist

This PR refactors the `find_project_root()` and `find_pypackage()` functions so that they now recursively looks for the `pyproject.toml` file or the `__pypackages__/{version}/lib` directory in the current directory or its parent directory. This means that we no longer need to set an upper limit on the number of levels we tranverse in such a search, and the `PDM_PROJECT_MAX_DEPTH` environment variable and the `project_max_depth` setting is no longer useful.


frostming (approved):

frostming (changes requested):
Can you add a news fragment under `news/`?

xulongwu4:
Added a news fragment

drunkwcodes:
Well done, @xulongwu4 !
mojo-muse also appreciate your work.

PR #2291, opened 2023-10-02
Cookiecutter raising exception instead of returning errcode

KalleDK:
Seems like click of cookiecutter has changed to return None instead of 0

## Pull Request Checklist

Seems like click of cookiecutter has changed to return None instead of 0


frostming:
It looks like a mistake. In fact, the `main` itself is an entry point function that exits on error by its own.

Can you change it to not checking the retval at all?

KalleDK:
What should it raise instead of SystemExit ?

PR #2370, opened 2023-11-04
fix: import error when install pyqt6 and install.cache = True

### LINKED ISSUE #2303: import error when install pyqt6 and install.cache = True

## Steps to reproduce

1. mkdir test
2. cd test 
3. pdm init 
4. pdm add pyqt6 
5. open a pdm python and import PyQt6

## Actual behavior

```bash
D:\WalkingPad\test>pdm run
No command is given, default to the Python REPL.
Python 3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from PyQt6 import QtGui
  File "<stdin>", line 1, in <module>
ImportError: DLL load failed while importing QtGui: 找不到指定的模块。
>>> exit()
```

it is ok when install.cache = False.

## Environment Information

```bash
D:\WalkingPad\test>pdm info
PDM version:
  2.9.3
Python Interpreter:
  D:\Python\Python310\python.EXE (3.10)
Project Root:
  D:/WalkingPad/test
Local Packages:
  D:\WalkingPad\test\__pypackages__\3.10

D:\WalkingPad\test>pdm info --env
{
  "implementation_version": "3.10.9",
  "os_name": "nt",
  "platform_machine": "AMD64",
  "platform_release": "10",
  "platform_system": "Windows",
  "platform_version": "10.0.22621",
  "python_full_version": "3.10.9",
  "python_version": "3.10",
  "sys_platform": "win32"
}

```

xzmeng:
## Pull Request Checklist

Fixes #2303

### DIFF
--- a/src/pdm/installers/installers.py
+++ b/src/pdm/installers/installers.py
@@ -32,6 +32,8 @@ if TYPE_CHECKING:
 
     from pdm.environments import BaseEnvironment
 
+TEXT_FILE_SUFFIXES = (".py", ".pyi", ".ipynb", ".pth")
+
 
 @lru_cache()
 def _is_python_package(root: str | Path) -> bool:
@@ -120,7 +122,7 @@ def _create_symlinks_recursively(source: str, destination: str) -> Iterable[str]
             destination_path = os.path.join(destination_root, f)
             if os.path.exists(destination_path):
                 os.remove(destination_path)
-            if any(f.endswith(x) for x in [".py", ".pyi", ".pyx", ".ipynb", ".pth"]):
+            if f.endswith(TEXT_FILE_SUFFIXES):
                 os.symlink(source_path, destination_path, False)
             else:
                 # Some compiled files use relative paths to themeselves to reference other files,
--- a/src/pdm/installers/installers.py
+++ b/src/pdm/installers/installers.py
@@ -120,7 +120,7 @@ def _create_symlinks_recursively(source: str, destination: str) -> Iterable[str]
             destination_path = os.path.join(destination_root, f)
             if os.path.exists(destination_path):
                 os.remove(destination_path)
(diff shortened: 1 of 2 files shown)


src/pdm/installers/installers.py
```diff
@@ -113,7 +120,17 @@ def _create_symlinks_recursively(source: str, destination: str) -> Iterable[str]
             destination_path = os.path.join(destination_root, f)
             if os.path.exists(destination_path):
                 os.remove(destination_path)
-            os.symlink(source_path, destination_path, False)
+            if any(f.endswith(x) for x in [".py", ".pyi", ".pyx", ".ipynb", ".pth"]):
```
frostming, line 123:
```suggestion
            if any(f.endswith(TEXT_FILE_SUFFICES)):
```

And make `TEXT_FILE_SUCCFICES` a combination of these suffices(`endswith` can take a tuple as parameter)

frostming:
I am still concerned that this solution may be too "specialized". Specifically, the root cause is that some binaries find other files through its own relative path. However, in this PR, all _non-text_ files have been changed to hard links for this purpose, and you can notice the mismatch between the solution and the problem. As a general package manager, we cannot only focus on solving specific issues with a package, but hope that this solution is generally preferable in most packages.

With this said, I cannot propose a more appropriate solution. Perhaps we can reverse the filtering rules, from matching text files to not matching binary files.

xzmeng:
Completely agree with what you said.

This PR cannot resolve the problem fundamentally, because we cannot prevent a package from doing anything that breaks the symlink (like what PyQt6 does). And the usage of hard link makes `install.cache_method=symlink` inconsistent either from source code or from behavior.

>Or, we can add another new cache method named hardlink that creates hard links instead of symlinks. When user opts this in, all files are linked by hard links

This way may be better.

PR #2394, opened 2023-11-10
feat: add a new setting `package-type` to distinguish application and library

frostming:

## Pull Request Checklist


deronnax:
This option is still undocumented 😢

frostming:
> This option is still undocumented 😢 

yeah. But how would you like it to be documented and why not contribute?

Issue #2360, opened 2023-11-02
Need to prevent wrong project name (including space and ilegal characters)


metemaddar:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

**Make sure you run commands with `-v` flag before pasting the output.**

## Steps to reproduce

Init a project with a name including space. For example "New Project"

## Actual behavior

The PDM accepts the name and creates  the project. Now if you try to add a package you see the following error:
```bash
[BuildError]: Build backend raised error: Showing the last 10 lines of the build output:
    self.initialize(context)
  File "/tmp/pdm-build-env-6vjura12-shared/lib/python3.10/site-packages/pdm/backend/base.py", line 180, in initialize
    self.call_hook("pdm_build_initialize", context)
  File "/tmp/pdm-build-env-6vjura12-shared/lib/python3.10/site-packages/pdm/backend/base.py", line 138, in call_hook
    getattr(hook, hook_name)(context, *args, **kwargs)
  File "/tmp/pdm-build-env-6vjura12-shared/lib/python3.10/site-packages/pdm/backend/editable.py", line 28, in pdm_build_initialize
    editables = self._prepare_editable(context)
  File "/tmp/pdm-build-env-6vjura12-shared/lib/python3.10/site-packages/pdm/backend/editable.py", line 61, in _prepare_editable
    raise ConfigError(str(e)) from None
pdm.backend.exceptions.ConfigError: Project name New Project is not valid
```

## Expected behavior

Prompt to fix the project name including help text.

## Environment Information

```bash
PDM version:
  2.10.0
Python Interpreter:
  /home/mo/eawa-bot/.venv/bin/python (3.10)
Project Root:
  /home/mo/eawa-bot
Local Packages:
  
{
  "implementation_name": "cpython",
  "implementation_version": "3.10.12",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.2.0-34-generic",
  "platform_system": "Linux",
  "platform_version": "#34~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Sep  7 
13:12:03 UTC 2",
  "python_full_version": "3.10.12",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "linux"
}
```

PR #2402, opened 2023-11-14
Issues 2360 (#1)

LeetaoGoooo:
## Pull Request Checklist

Fix bugs mentioned in https://github.com/pdm-project/pdm/issues/2360, do some check before init a project, follow [PEP 426](https://peps.python.org/pep-0426/#name) and sanitize the default name by replacing the illegal characters

### DIFF
--- a/src/pdm/cli/commands/init.py
+++ b/src/pdm/cli/commands/init.py
@@ -14,7 +14,7 @@ from pdm.models.backends import _BACKENDS, DEFAULT_BACKEND, BuildBackend, get_ba
 from pdm.models.python import PythonInfo
 from pdm.models.specifiers import get_specifier
 from pdm.models.venv import get_venv_python
-from pdm.utils import get_user_email_from_git, package_installed, validate_project_name
+from pdm.utils import get_user_email_from_git, package_installed, sanitize_project_name, validate_project_name
 
 if TYPE_CHECKING:
     from pdm.project import Project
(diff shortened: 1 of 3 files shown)


src/pdm/cli/commands/init.py
```diff
@@ -90,6 +90,21 @@ def ask(self, question: str, default: str) -> str:
             return default
         return termui.ask(question, default=default)
 
+    def ask_project(self, project: Project) -> str:
+        name = self.ask("Project name", project.root.name)
+        if validate_project_name(name):
+            return name
+        project.core.ui.echo(
+            "Project name is not validate \n"
```
pawamoy, line 98:
```suggestion
            "Project name is not valid \n"
```

src/pdm/utils.py
```diff
@@ -453,3 +453,26 @@ def package_installed(package_name: str) -> bool:
         return False
     else:
         return True
+
+
+def validate_project_name(name: str) -> bool:
+    """Check if the project name is validate or not"""
```
pawamoy, line 459:
```suggestion
    """Check if the project name is valid or not."""
```

src/pdm/utils.py
```diff
@@ -453,3 +453,26 @@ def package_installed(package_name: str) -> bool:
...
+
+    allowed_chars = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-")
```
pawamoy, line 461: I suppose we could use `string.ascii_letters` and `string.digits` instead:

```suggestion
    allowed_chars = set(string.ascii_letters + string.digits + "_-")
```

LeetaoGoooo: yeah, you're right, it's much better ~~ anything else

src/pdm/utils.py
```diff
@@ -453,3 +453,27 @@ def package_installed(package_name: str) -> bool:
...
+
+    if keyword.iskeyword(name):
+        return False
+
+    import builtins
+
+    if name in dir(builtins):
+        return False
+
+    return True
```
frostming, line 479: We can only consider the standard of [PEP 426](https://peps.python.org/pep-0426/#name) which even gives a regex pattern that you can use. No need to check the keywords and builtin.

src/pdm/cli/commands/init.py
```diff
@@ -90,6 +90,21 @@ def ask(self, question: str, default: str) -> str:
             return default
         return termui.ask(question, default=default)
 
+    def ask_project(self, project: Project) -> str:
+        name = self.ask("Project name", project.root.name)
```
frostming, line 94: Also please don't validate the default value or user may feel confused.
Instead, we sanitize the default name by replacing the illegal characters.

frostming:
Fine, can you please add a news fragment under `news` and we are good to go.

LeetaoGoooo:
> Fine, can you please add a news fragment under `news` and we are good to go.

It is completed. Thank you for your patient guidance and assistance.～～

frostming (approved):

PR #2408, opened 2023-11-15
add support for `{pdm}` placeholder in scripts

tgolsson:
## Pull Request Checklist

    - Only found a single meaningful test-case -- I tried to do some more testing for the interpolation but it got far too complex with paths, dev versions etc.  

As discussed on Discord, this adds support for using a `{pdm}` placeholder in scripts to refer to the current interpreter. I'm not sure if this should affect calls in any way, but I don't think so since they'd (presumably) use the same python interpreter/venv.

### DIFF
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -45,9 +45,10 @@ def exec_opts(*options: TaskOptions | None) -> dict[str, Any]:
 
 
 RE_ARGS_PLACEHOLDER = re.compile(r"{args(?::(?P<default>[^}]*))?}")
+RE_PDM_PLACEHOLDER = re.compile(r"{pdm}")
 
 
-def interpolate(script: str, args: Sequence[str]) -> tuple[str, bool]:
+def _interpolate_args(script: str, args: Sequence[str]) -> tuple[str, bool]:
     """Interpolate the `{args:[defaults]} placeholder in a string"""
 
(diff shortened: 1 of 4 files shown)


tgolsson:
Hmm, will have to look at Windows failures tomorrow. I'm going to guess related to path normalization in some way...

src/pdm/cli/commands/run.py
```diff
@@ -58,6 +59,24 @@ def replace(m: re.Match[str]) -> str:
...
+    interpolated, count = RE_PDM_PLACEHOLDER.subn(replace, script)
+    return interpolated, count > 0
+
+
+def interpolate(script: str, args: Sequence[str]) -> tuple[str, bool]:
+    """Interpolate the `{args:[defaults]} placeholder in a string"""
+
+    script, args_interpolated = _interpolate_args(script, args)
+    script, pdm_interpolated = _interpolate_pdm(script)
+    return script, args_interpolated or pdm_interpolated
```
frostming, line 77: You can't mix args_interpolated with pdm_interpolated, since the former is needed by the runner to know whether it needs to append other arguments.

frostming, line 77:
```suggestion
    script, _ = _interpolate_pdm(script)
    return script, args_interpolated
```

frostming:
Hi, I changed the base to dev, can you resolve the conflicts? Thanks

PR #2418, opened 2023-11-22
feat: export cross-platform requirements

frostming:

## Pull Request Checklist


deronnax:
Hey, we actually appreciated (and used) the previous requirement (snyk does not understands platform markers). Would it be doable to add an option to export following the previous behavior?
Thank you.

frostming:
> (snyk does not understands platform markers)

That is quite surprising, does it mean it only understands the output of `pip freeze`?

deronnax:
I think so. I re-did the test various time and I confirm snyk does not understand platform markers. I can provide use case if you want.

PR #2421, opened 2023-11-22
feat: New lock strategy: inherit_metadata

frostming:

## Pull Request Checklist


sigma67:
@frostming I'm curious, are there any disadvantages to using the `inherit_metadata` strategy? I'm wondering why it is opt-in.

If so it should probably be mentioned in doc/changelog.

frostming:
> are there any disadvantages to using the new, faster `inherit_metadata` strategy?

Since it is a significant change in how installation finds packages, I am not sure it is 100% consistent with old behavior. But I also like this strategy and would like to use it as default in my packages.

PR #2460, opened 2023-12-02
fix default verbosity level in UI.echo docstring

deronnax:
## Pull Request Checklist

`UI.echo` docstring default verobsity level was wrong. it was changed in [#2304](https://github.com/pdm-project/pdm/pull/2304/files#diff-731843a62f3036f02e987fb915be5c721aaad5010b252aa88383dd7184dafb11R180)
However, I am bit skeptical: shouldn't the default level be `NORMAL`?

### DIFF
--- a/src/pdm/termui.py
+++ b/src/pdm/termui.py
@@ -184,7 +184,7 @@ class UI:
 
         :param message: message with rich markup, defaults to "".
         :param err: if true print to stderr, defaults to False.
-        :param verbosity: verbosity level, defaults to NORMAL.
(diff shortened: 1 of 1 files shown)


frostming:
> However, I am bit skeptical: shouldn't the default level be `NORMAL`?

Default to `NORMAL` means it can be suppressed by `--quiet`, but there are many information-only commands whose only effect is to show something. If the default value is `QUIET` we don't need to bother passing the level explicitly.

deronnax:
OK, legit. Thank you very much.

PR #2480, opened 2023-12-08
feat: Show subcommand's help info when passing unrecognized arguments

leemars:
## Pull Request Checklist

When passing unrecognized arguments to a pdm subcommand, pdm displays its own help info instead of the subcommand's one. It would be more helpful for users if pdm could display the help info of the subcommand.

This issue is caused by a known design of Python's argparse module, and the argparse module is not planning to improve it. (Refs: https://github.com/python/cpython/issues/83287 )

I found a modification on Stackoverflow, which I tried and found to be effective. (Refs: https://stackoverflow.com/a/61724607 )

### DIFF
--- a/news/2480.feature.md
+++ b/news/2480.feature.md
@@ -1,0 +1,1 @@
+Show subcommand's help info when passing unrecognized arguments.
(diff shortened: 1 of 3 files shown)


frostming (approved):
No other side effects were found, accepted.

PR #2482, opened 2023-12-10
doc: update Dockerfile snippet

sanmai-NL:
## Pull Request Checklist

### DIFF
--- a/docs/docs/usage/advanced.md
+++ b/docs/docs/usage/advanced.md
@@ -148,38 +148,219 @@ Testing:
 
 ## Use PDM in a multi-stage Dockerfile
 
-It is possible to use PDM in a multi-stage Dockerfile to first install the project and dependencies into `__pypackages__`
-and then copy this folder into the final stage, adding it to `PYTHONPATH`.
+The following Containerfile/Dockerfile template shows:
+
+- Using multiple stages to separate building from production and development container images.
+- Caching of PDM data. The caching approach is generic and doesn’t show how CI platform builders that distribute builds
+across ephemeral hosts could be adopted efficiently.
+- Copying the minimal amount of data.
+
+`.containerignore`/`.dockerignore`:
+
+```dockerignore
+*
+.*
+!/pdm.lock
+!/pdm.toml
+!/pyproject.toml
+!/README.md
+!/src/
+```
+
+`Containerfile`/`Dockerfile`:
 
 ```dockerfile
-# build stage
-FROM python:3.8 AS builder
-
-# install PDM
-RUN pip install -U pip setuptools wheel
-RUN pip install pdm
-
-# copy files
-COPY pyproject.toml pdm.lock README.md /project/
-COPY src/ /project/src
-
-# install dependencies and project into the local packages directory
-WORKDIR /project
-RUN mkdir __pypackages__ && pdm sync --prod --no-editable
-
-
-# run stage
-FROM python:3.8
-
-# retrieve packages from build stage
-ENV PYTHONPATH=/project/pkgs
-COPY --from=builder /project/__pypackages__/3.8/lib /project/pkgs
-
-# retrieve executables
-COPY --from=builder /project/__pypackages__/3.8/bin/* /bin/
-
-# set command/entrypoint, adapt to fit your needs
-CMD ["python", "-m", "project"]
+# syntax=docker/dockerfile:1
+ARG IMAGE_PREFIX=
+FROM ${IMAGE_PREFIX}alpine:3.18 AS builder
+ARG NAME_PRODUCT
+ARG REVISION
+# ,z for Podman, empty otherwise
+ARG SELINUXRELABEL
+ENV \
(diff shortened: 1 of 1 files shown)


sanmai-NL:
This was written up hastily based on one existing Containerfile setup of ours, that's still a work in progress. Please review/contribute on:

- Using settings and parameters where the default values would be equal.
- Improvements.
- Adaptations and extra context for various CI environments (GitHub Actions, GitLab).

docs/docs/usage/advanced.md
```diff
@@ -148,38 +148,230 @@ Testing:
...
-RUN pip install pdm
+- Using multiple stages to separate building from production and development container images.
+- Caching of PDM data. The caching approach is generic and doesn’t show how CI platform builders that distribute builds
+across ephemeral hosts could be adopted efficiently.
+- Copying the minimal amount of data.
 
-# copy files
-COPY pyproject.toml pdm.lock README.md /project/
-COPY src/ /project/src
+`pdm.toml`:
```
frostming, line 158: Is it part of the "minimal amount of data"?

The following example seems too complex to me.

sanmai-NL: Unless caching is on by default. For a time, it wasn't.

sanmai-NL: It depends on the requirements of the users. If they build small apps, it could be shorter (not so much simpler). If they build large data science apps and don't want to wait hours and/or pay large infrastructure costs, I don't see much opportunity for simplifying.

frostming: Well, the purpose for this part in docs is to showcase how users can integrate PDM in the dockerfile, not how to do it correctly or most efficiently. People should adjust the dockerfile to what they need in the production.

While your example looks like a spaceship, it fits more in a dedicated repo named `pdm-docker-example` than in a documentation section.

sanmai-NL: I might add that the commands to use PDM are comprehensively documented already and someone who knows container basics and PDM should be ready to go without an example.

frostming: If the document gives you that impression, I'm fine with removing it from the document altogether (or leaving some links)

A well-explained blog post might also be nice, and I'd be happy to put it in `awesome-pdm`

deronnax: Wow, sorry but the I feel Dockerfile example is way, way too complex.

sanmai-NL: The interesting part would, how did you achieve the desired performance with your previous tool? Then we can compare and determine the required complexity level to achieve your objectives with PDM.

By the way, I find GitHub Actions and GitLab CI/CD to be very complex. Don't confuse verbosity with complexity.

noirbizarre: Great ! I ~can~ will submit a PR in this way 👍🏼

sanmai-NL: @noirbizarre My Dockerfile template isn't ready yet. The current caching tactics (symlink, hardlink) don't work with Buildkit mount caching, like I use. Moreover there's opportunity for some repetition reduction in my sample.

On a general level, I think that PDM suggesting a plugin/tool for containerizing a PDM-managed product is a bit roundabout.

See also https://github.com/orgs/python-poetry/discussions/1879 for prior art.

Presently, in the Poetry docs I could only find one mention, a tip on a specific containerization best practice. So not a complete Containerfile template.

I think it's best if @noirbizarre and me update his template and perhaps PDM promote his package, and remove the full Containerfile template from the docs.

sanmai-NL: Let's keep this PR open to capture more input, improvements, before @noirbizarre  and I move over to his repo.

noirbizarre: TBH, I think Docker images for applications are very specific to their target deployment and I think this is not the role of `pdm` to handle that (I am a regular contributor, I could have submitted all this to the core project but I chose to craft a plugin as the mechanism is meant just for that).
For documentation purpose, example here should be the minimum viable (even if it's not deployment ready, this is not the goal of this snippet I think). Past that, it's your specific way of doing it (I understand the snippet you provided, but it is both overkill and not portable to showcase as example to me).
However, the plugin is also a place where we can share this kind of examples that fit some different needs and processes.

Here's my proposal @sanmai-NL:
- I submit a PR amending the doc to say "hey if you need to produce a docker image, there's a plugin for that"
- we work on `pdm-dockerize` for a place to showcase way of crafting docker images (sample dir, full mkdocs site... whatever)
- we can have continuous improvement on the docker part without pressure on or from the core project (and there are a lot of stuff to do, especially with docker frontend engine 1+)

This way:
- we keep `pdm` doc as minimalist as possible so can people find quickly doc or solution for basic recurrent need
- we have a plugin which is able to do the common dirty work of dockerizing
- we have a place to share best practices or even ask questions specific to `pdm`+`docker`
- improving or fixing Docker-related stuff is not tied to `pdm` release, which can focus on slow pace but stable releases.

WDYT ?

PR #2494, opened 2023-12-14
drop support for python 3.7

danieleades:
support for python 3.7 has already been dropped, but some traces remain in a couple of places

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set Python 3.7
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.7
-          architecture: ${{ matrix.arch }}
       - name: Set Python 3.8
         uses: actions/setup-python@v4
         if: matrix.python-version != '3.8'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,7 +139,7 @@ extend-select = [
 extend-ignore = ["B018", "B019"]
 src = ["src"]
 exclude = ["tests/fixtures"]
-target-version = "py37"
+target-version = "py38"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -52,11 +52,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set Python 3.7
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.7
-          architecture: ${{ matrix.arch }}
```
frostming, line 59: Thanks, but please keep it for verifying PDM can still **manage** Python 3.7. We just drop Python 3.7 as a working version that PDM can be **installed** with.

danieleades: What is the python support policy for this project? Python 3.7 is end of life

frostming: We tend to extend that lifetime for management. But PDM itself no longer works on 3.7 right?

danieleades: i'm not saying it doesn't make sense to support older versions, i'm just wondering what the policy is?

- why support 3.7 and not 3.6?
- if 3.7 support won't be dropped now, then when?

Would be good to have a policy documented

frostming: It is difficult to have a clear boundary for the minimum supported version, as it depends on several factors:

1. The syntax used by the **in-process** scripts. This is maintained by PDM, but it only needs little effort to keep syntax compatibility.
2. The toolchain support, such as build backends. This is not controlled by PDM, some tools are still supporting up to Python 3.6, in this case users are able to run their projects on older Python versions. However some other tools may have an aggressive policy so users working on this tools are not able to use old Python versions.

In the CI matrix we are justing running a simple integration against Python 3.7, it doesn't mean we  only supports up to Python 3.7, or not supporting, for the above reasons.

PR #2515, opened 2023-12-20
fix(scripts): expose `$PYTHONPATH` in scripts if needed

noirbizarre:
## Pull Request Checklist

This PR fixes the missing `PYTHONPATH` environment variable missing for script execution.
Before 2.11, it was not an issue because project were self-installing on `pdm install`.
Now, in `pdm>=2.11`, project are either `package-type=library` or `package-type=application`. Applications aren't self-installed anymore (and can't) so sources not at the project root are not visible anymore in scripts.
It is documented that `pdm` automatically handle non-root packages in 2 ways:
- explicit `tool.pdm:package-dir`
- implicit if there is a `src` directory
But this mechanism only exists in `pdm-backend`: https://github.com/pdm-project/pdm-backend/blob/main/src/pdm/backend/config.py#L250-L254

This PR:
- expose the resolved `package_dir` on the `Project` using the same algorithm than `pdm-backend`
- ensure the script `PYTHONPATH` is properly computed for `package-type=application` when `package-dir` is not the project root
- ensure the `PYTHONPATH` is properly injected (var was existing but not injected in the environment)

### DIFF
--- a/src/pdm/project/core.py
+++ b/src/pdm/project/core.py
@@ -681,3 +681,9 @@ class Project:
     @property
     def is_library(self) -> bool:
         return bool(self.name) and self.pyproject.settings.get("package-type", "library") == "library"
+
+    @property
+    def package_dir(self) -> str | None:
+        """An optional directory containing the project sources"""
+        default = "src" if self.root.joinpath("src").exists() else None
+        return self.pyproject.settings.get("build", {}).get("package-dir", default)
--- a/news/2515.bugfix.md
+++ b/news/2515.bugfix.md
@@ -1,0 +1,1 @@
+Add `package-dir` to scripts `$PYTHONPATH` if project is not a library
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -189,6 +189,10 @@ class TaskRunner:
         project_env = project.environment
         this_path = project_env.get_paths()["scripts"]
         process_env.update(project_env.process_env)
+        if not project.is_library and project.package_dir:
+            pythonpath.insert(0, str(project.root / project.package_dir))
+        if pythonpath:
+            process_env["PYTHONPATH"] = os.pathsep.join(pythonpath)
         if env:
             process_env.update(env)
         if shell:
--- a/tests/cli/test_run.py
+++ b/tests/cli/test_run.py
@@ -855,3 +855,33 @@ def test_empty_positional_args_display_help(project, pdm):
     assert "Usage:" in result.output
     assert "Commands:" in result.output
     assert "Options:" in result.output
+
+
+@pytest.mark.parametrize("pkg_type", ["library", "application", pytest.param(None, id="not-set")])
+@pytest.mark.parametrize("pkg_prefix", [pytest.param("", id="root"), "src", "other", pytest.param(None, id="implicit")])
+def test_pythonpath(project, pdm, pkg_type, pkg_prefix, capfd):
+    pkg_root = project.root / ("src" if pkg_prefix is None else pkg_prefix) / "pkg"
+    pkg_root.mkdir(parents=True, exist_ok=True)
+    (pkg_root / "__init__.py").write_text(
+        textwrap.dedent(
+            """
+            def main():
+                print("OK")
+            """
+        )
+    )
+    project.pyproject.settings["scripts"] = {
+        "test": "python -c 'from pkg import main; main()'",
+    }
+    if pkg_type is not None:
+        project.pyproject.settings["package-type"] = pkg_type
+    if pkg_prefix:
+        project.pyproject.settings["build"] = {"package-dir": pkg_prefix}
+    project.pyproject.write()
+    pdm("install", obj=project, strict=True)
+    capfd.readouterr()
+
+    result = pdm("run test", obj=project)
+    out, err = capfd.readouterr()
+    assert result.exit_code == 0, err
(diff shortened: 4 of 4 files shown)


src/pdm/project/core.py
```diff
@@ -681,3 +681,9 @@ def tool_settings(self) -> dict[str, Any]:
     @property
     def is_library(self) -> bool:
         return bool(self.name) and self.pyproject.settings.get("package-type", "library") == "library"
+
+    @property
+    def package_dir(self) -> str | None:
+        """An optional directory containing the project sources"""
+        default = "src" if self.root.joinpath("src").exists() else None
+        return self.pyproject.settings.get("build", {}).get("package-dir", default)
```
frostming, line 689: This is backend-specific and PDM shouldn't rely on this.

frostming:
I am -1 on this PR.

> Now, in pdm>=2.11, project are either package-type=library or package-type=application

I don't think there is anything changed, before 2.11 if you have name and version property that project will become a library, no matter what you think it is. 2.11 just makes it more explicit. You can still achieve the same by setting` package-type` to `library`

For an application it won't be self-installed and it's expected to not having the project root in python path. Think of it in the old days how you install a web project with `requirements.txt`. You must cd to the correct location to make the application work.

noirbizarre:
Oh OK it works this way 😅 Interesting because I did not understand it this way.

So maybe naming is a bit misleading given `library` and `application` have meaning which does not imply anything on self-installation and so it's not explicit. [The documentation](https://pdm-project.org/latest/usage/project/#library-or-application) don't says applications can't be self-installed, it says libraries should be packageable and publisheable (aka. it should be installable by tools other than `pdm`) and that it will be self-installed automatically, which is not exactly what it is doing (because `application` explicitly forbid self-install which [now require an explicit library](https://github.com/pdm-project/pdm/blob/main/src/pdm/cli/actions.py#L231)).

Because naively, I set `package-type=library` on all libraries and `package-type=application` on all applications to be explicit, but it just broke all my application type projects 😅 (because src layout don't work with `package-type=application`)

Given `pdm init` still create an `src`-layout project for applications too, this is totally misleading because the resulting generate project won't be able to use commands with packages references and packages in src are not importable meaning that `pdm run` (interactive shell) and `pdm run python` seems broken for them. So I think the result of `pdm init` for `package-type=application` is currently broken.

Maybe self-installation shouldn't be triggered by that because in the `setup.py` or requirements days, it was possible to `pip install -e .` and I was doing it on all projects, was it a library or an application (I just put `-e .` in the `requirements.pip` and it was working just fine, especially for src-layout).

Intuitively, I would have:

- something like an explicit `tool.pdm.self-install: bool` telling whether it should self-install during dev using `pdm sync|install` (and to be consistent with the cli flag `--no-self`, which is also misleading because it implies self-install is the default which is not the case anymore, maybe there should be a `--self` flag for app). `package-type` should just change this default behavior but shouldn't prevent it if needed.
- the `tool.pdm.package-type: library|application` which is just telling whether it will be packaged/published so whether it should comply to PEP517 and indicate it has extra metadata requirements like `project.name` and `project.version` (I would enforce `description` or `readme` for libs) AND a PEP517 backend should be present (and maybe there should be `pdm check` command should validate the `pyproject.toml` and fail if those are missing)

Because in my opinion, most project will just set `package-type=library` without being a library and then it will just remove any sense to this statement

WDYT ?

frostming:
> So I think the result of `pdm init` for `package-type=application` is currently broken.

I think that makes sense to me, we can improve it following this path.

noirbizarre:
Agreed !

I think we need to ensure first that the `pdm init` don't generate a src-layout for an application, and amend the doc to add to the `Library or Application` section that:
- applications can't self install
- if not specified, `package-type` is considered a library

(I can submit a PR going this way)

Then what would be the path ?

I see 2 possibilities

### Proposal 1 - Explicit self-installation

This is mostly what I told above (and my preference):
- add an explicit `tool.pdm.self-install` which default to `True` for libs and `False` for applications. In case of applications, if `True`, enforce `project.name` and `project.version` being set and `pdm-backend` is implicitly used on `pdm install|sync` for dev if none is specified
- keep the `tool.pdm.package-type=library|application` and enforce `project.name`, `project.version` and `build-backend` for `tool.pdm.package-type=library` (do we enforce one of `description` or `readme` too ?)
- `pdm publish` should fail and explain why if `package-type=application` (It might already be the case, I just haven't checked yet). Installation with other tools from sources will just fail as usual if there is not PEP517 backend specified
- add a `pdm check` command which is validating the `pyproject.toml` consistency according to those rule (and why not also trigger a `pdm lock --check`)

### Proposal 2 - Be explicit on side effects but no changes

In addition to the introduction fixes (details in Library or Application, `pdm init` default template change to drop src-layout in case of application) we make sure that it is clear for everyone what are the side effects, and document/warn when necessary:
- the `--no-self` flag and `PDM_NO_SELF` env var is only relevant to libraries
- the `build` command makes no sense for `applications`
- non-root packages are not supported for `package-type=application` in scripts meaning that:
  - only tools having a config or parameter for paths will work (like `pytest` or linters)
  - `call` scripts won't be working with non-root relative scripts
  - any Python `import` won't be working with non-root package
- that the docker snippet only work for root-package based applications unless you also add the `src` or any other prefix to the `$PYTHONPATH` (which can be handled by `pdm-dockerize`)

WDYT ?

PR #2523, opened 2023-12-27
chore(lib): enhance warning message

### LINKED ISSUE #2522: Misleading `PackageWarning` when adding a package (`myst-nb==1.0.0`) with a version free dependency (`ipython`) whose `requires-python` range is not satisfied on the latest version

## Steps to reproduce

Just add `myst-nb==1.0.0` to any project that has a minimal Python version below or equal to `3.9`, like in the screenshot below:

image

## Actual behavior

A warning is thrown, when a valid version could be found anyway.

## Expected behavior

Just don't raise a warning because that scared me a bit like: do I have dependencies conflicts? Did I do something?

There is no issue that the latest version available may not be compatible with the Python versions you claim to support in the `requires-python` range. But what's the problem with that?

## Environment Information

```bash
PDM version:
  2.11.1
Python Interpreter:
  /home/jeertmans/Desktop/install-warning/.venv/bin/python (3.10)
Project Root:
  /home/jeertmans/Desktop/install-warning
Local Packages:
  
{
  "implementation_version": "3.10.12",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.2.0-39-generic",
  "platform_system": "Linux",
  "platform_version": "#40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 
UTC 2",
  "python_full_version": "3.10.12",
  "python_version": "3.10",
  "sys_platform": "linux"
}

```

jeertmans:
Closes #2522.

## Pull Request Checklist

- ~~[ ] A news fragment is added in `news/` describing what is new.~~ N/A
- ~~[ ] Test cases added for changed code.~~ N/A (?)

As discussed in #2522, the warning message is a bit misleading because the installation proceeds correctly, so this PR modifies the message to reflect that.

### DIFF
--- a/src/pdm/models/repositories.py
+++ b/src/pdm/models/repositories.py
@@ -196,12 +196,11 @@ class BaseRepository:
                     if working_requires_python.is_impossible:  # pragma: no cover
                         continue
                     warnings.warn(
-                        f"Could not install the latest available version for {candidate.name} "
-                        f"(i.e., {candidate.name}@{candidate.version}) because it requires "
+                        f"Skipping {candidate.name}@{candidate.version} because it requires "
                         f"{python_specifier(candidate.requires_python)} but the project claims to work with "
-                        f"{python_specifier(project_requires_python)}. Instead, the latest version of "
+                        f"{python_specifier(project_requires_python)}. Instead, another version of "
                         f"{candidate.name} that supports {python_specifier(project_requires_python)} will "
-                        f"be installed.\nIf you want to install {candidate.name}@{candidate.version}, "
+                        f"be used.\nIf you want to install {candidate.name}@{candidate.version}, "
                         "narrow down the `requires-python` range to "
                         f'include this version. For example, "{working_requires_python}" should work.',
                         PackageWarning,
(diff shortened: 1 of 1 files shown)


src/pdm/models/repositories.py
```diff
@@ -196,9 +196,13 @@ def python_specifier(spec: str | PySpecSet) -> str:
                     if working_requires_python.is_impossible:  # pragma: no cover
                         continue
                     warnings.warn(
-                        f"Skipping {candidate.name}@{candidate.version} because it requires "
+                        f"Could not install the latest available version for {candidate.name} "
```
frostming, line 199: But it may not be the latest version, this is not accurate.

frostming (commented):
I suspect that you may have made this information overly complicated in order to alleviate some possible panic. And that is not correct either.

We just iterate through each version of the package and see if the python range is compatible, but you call them all as "latest available version" , this is not the case, apparently.

jeertmans:
> I suspect that you may have made this information overly complicated in order to alleviate some possible panic. And that is not correct either.
> 
> We just iterate through each version of the package and see if the python range is compatible, but you call them all as "latest available version" , this is not the case, apparently.

This is what I was skeptical about, thanks for clarifying. I updated the message.

frostming (approved):

PR #2564, opened 2024-01-17
feat: rename the package-type config to distribution

frostming:

## Pull Request Checklist

### DIFF
--- a/news/2564.feature.md
+++ b/news/2564.feature.md
@@ -1,0 +1,1 @@
+Rename the `package-type` field under `tool.pdm` settings table to `distribution` to make it more clear.
--- a/src/pdm/cli/actions.py
+++ b/src/pdm/cli/actions.py
@@ -227,7 +227,7 @@ def do_sync(
         clean=clean,
         dry_run=dry_run,
         no_editable=no_editable,
-        install_self=not no_self and project.is_library,
+        install_self=not no_self and project.is_distribution,
         reinstall=reinstall,
         only_keep=only_keep,
         fail_fast=fail_fast,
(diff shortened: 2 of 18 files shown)


lmmx:
@frostming Cool, should this be updated in the docs section "Usage > Project > Library or Application"?

electrocnic:
For some reason, I now get a weird error when I try to run `pdm install` in my existing project.
It worked fine with the old `package-type = "application"` and it would still work with that same line after my update from 3.11.1 to 3.12.1, but now when I use `distribution = "false"` I get:

image

with the log folder content:
```
pdm.termui: error: Multiple top-level packages discovered in a flat-layout: ['api', 'pages', 'stats', 'vuedj', 'contact', 'pricing', 'accounts', 'test_util'].
pdm.termui:
pdm.termui: To avoid accidental inclusion of unwanted files or directories,
pdm.termui: setuptools will not proceed with this build.
pdm.termui:
pdm.termui: If you are trying to create a single distribution with multiple packages
pdm.termui: on purpose, you should not rely on automatic discovery.
pdm.termui: Instead, consider the following options:
pdm.termui:
pdm.termui: 1. set up custom discovery (`find` directive with `include` or `exclude`)
pdm.termui: 2. use a `src-layout`
pdm.termui: 3. explicitly set `py_modules` or `packages` with a list of names
pdm.termui:
pdm.termui: To find more information, look for "package discovery" on setuptools docs.
```
My folder layout and packages and other stuff did not change.
What am I supposed to do to fix this?

For now, I left the old "application" string in my pyproject.toml.

frostming:
@electrocnic distribution = false

without the quotes

frostming:
Thanks, the auto fixer isn't behaving correctly.

Andreas Wiesinger ***@***.***> 于2024年1月21日周日 01:07写道：

steakhutzeee:
I had the same issue and solved with that, thought the problem was with the fact i'm running an application and not a library and i had to remove the project name from pyproject.toml.

Is this still the case? Should project name be omitted? It actually works with the name, no version set, and  `distribution = false`

deronnax:
to link things together: package-type was introduced in https://github.com/pdm-project/pdm/pull/2394.

PR #2582, opened 2024-01-23
An option to run full composite despite to errors

mganisin:
New option introduced that changes the behavior of composite script to executes all the scripts despite to present errors. Exit code is sum of all the codes.  The name of the option 'keep_going' is inspired/reused from 'GNU make' with similar/same purpose.

## Pull Request Checklist

### DIFF
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -34,6 +34,7 @@ if TYPE_CHECKING:
         env: Mapping[str, str]
         env_file: EnvFileOptions | str | None
         help: str
+        keep_going: bool
         site_packages: bool
 
 
@@ -103,7 +104,7 @@ class TaskRunner:
     """The task runner for pdm project"""
 
     TYPES = ("cmd", "shell", "call", "composite")
-    OPTIONS = ("env", "env_file", "help", "site_packages")
+    OPTIONS = ("env", "env_file", "help", "keep_going", "site_packages")
 
     def __init__(self, project: Project, hooks: HookManager) -> None:
         self.project = project
@@ -270,7 +271,8 @@ class TaskRunner:
             args = list(args)
             should_interpolate = any(RE_ARGS_PLACEHOLDER.search(script) for script in value)
             should_interpolate = should_interpolate or any(RE_PDM_PLACEHOLDER.search(script) for script in value)
-            code = 0
+            composite_code = 0
+            keep_going = options.pop("keep_going", False) if options else False
             for script in value:
                 if should_interpolate:
                     script, _ = interpolate(script, args)
@@ -279,8 +281,10 @@ class TaskRunner:
                 subargs = split[1:] + ([] if should_interpolate else args)
                 code = self.run(cmd, subargs, options, chdir=True)
                 if code != 0:
-                    return code
-            return code
+                    if not keep_going:
+                        return code
+                    composite_code += code
+            return composite_code
         return self._run_process(
             args,
             chdir=True,
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -283,7 +283,7 @@ class TaskRunner:
                 if code != 0:
                     if not keep_going:
                         return code
-                    composite_code += code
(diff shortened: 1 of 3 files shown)


src/pdm/cli/commands/run.py
```diff
@@ -279,8 +281,10 @@ def run_task(self, task: Task, args: Sequence[str] = (), opts: TaskOptions | Non
                 subargs = split[1:] + ([] if should_interpolate else args)
                 code = self.run(cmd, subargs, options, chdir=True)
                 if code != 0:
-                    return code
-            return code
+                    if not keep_going:
+                        return code
+                    composite_code += code
```
frostming, line 286: What is the meaning of sum of the return code, if more than one processes return non-zero?

mganisin: Well, the expectation (mine expectation) is that composite will fail in case of a failure. Only change is that it will run all the "sobcommands", so non-zero code is needed. And so far I believe the exact value of the code does not matter that much, so I chose to do a sum. Anyway any value, even fixed should be appropriate, just the indication that something has failed (non-zero exit code) should remain.

In theory, if all the "subcommands" will fail with code 1 (or any other known value), then yes, I was also thinking that it could be useful information to preserve it somehow in the composite_code, therefore sum. Honestly I do not assume that this can have real use, but still there.

frostming:
I am not going to accept this patch because I don't think this requirement has been carefully considered.
There may be many different situations here, each corresponding to a user case.

1. People want to ignore errors for specific steps, but fail fast for others
2. People want to continue on errors, but the return code should be equal to what the second step returns. As long as the second step succeeds, the whole task should succeed.

While your patch only cover a small part of all use cases, yes, your use case, it will bring more confusion compared to not supporting at all. The scope of a feature may be larger than you'd think, large enough that I think it goes beyond the responsibilities of PDM. As a workaround, you can either:

1. Make a script wrapper that ignores errors for specific commands
2. Seek for help in a dedicated task orchestrator

PDM is never built to be an all-in-one tool, we choose to not support some features, leaving the flexibility to users.

mganisin:
Correct, this covers only an existing use case that I am aware of. Actually it might be rather bad idea if I was implementing a solution for use cases that I am not aware of as the outcome wouldn't be optimal.

This proposal doesn't seem to be in conflict with use cases you outlined.

For example anyone interested in case #1 could suggest/implement another option like 'ignore_errors' (again inspiration for name from 'make') for individual scripts and that would lead to pretty versatile solution yet included in pdm out of the box. (if ever needed, maybe in such case it would make sense to implement ignoring errors directly in the script; I can't say this is not my use case)

Case #2 really sounds to me like dependency execution (like 'make' or other pipeline managers do) and I understand, that is not in scope of pdm. Neither this my PR suggests anything similar, I don't consider my proposal to be a scope changer.

I understand that there can be many other 'unspoken' and uncovered use cases, however the development should not be stopped by possible conflicts with unknowns, should be?

Of course, plenty of solutions outside of pdm already exist, actually neither 'composite' is needed. But it is there luckily and it is very helpful. It can be even better with this small change.

I can think of returned code, current sum doesn't seem to be optimal actually. I think fixed value or code of last failure would be much better.

I'd like to ask you kindly to reconsider your decision again. I agree this does solve only my requirement, on the other hand I believe that it is fair. The 'composite' feature when implemented also solved only someone's requirement, at the end is useful for everyone. Furthermore as mentioned I don't think this neither conflicts with anything nor introduces any sign of complexity.

Thank You for your time spent on my PR in any case.

pawamoy:
I'd suggest following what Bash's `pipefail` option does: returning 0 if all commands returned 0, or the exit code of the last command (rightmost) that failed.

> If set, the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands in the pipeline exit successfully. This option is disabled by default.

IMO summing exit codes is not only confusing, it can even be wrong: if all exit codes reach 256 together, then your exit code will be modded to 0!

mganisin:
I updated this PR to return rather the code of last failed job in case of failure.

PR #2611, opened 2024-02-03
fix: properly handle interpreter resolution while importing

### LINKED ISSUE #2608: `pdm install` with implicit import from setup.py (non-PEP621) cannot resolve any dependencies

## Steps to reproduce

1. Check out any project that does not use PEP 621-compliant metadata, for example, [liquidctl/liquidctl](https://github.com/liquidctl/liquidctl/)
2. Attempt to run `pdm install`
3. Agree to import from setup.cfg using "setuppy"

## Actual behavior

pdm generates a massive amount of dependency resolution failure messages reporting absence of `required-python` configuration, ending up in a failure to import the project:

```
<...>
/home/intelfx/devel/ext/pdm/src/pdm/resolver/providers.py:202: PackageWarning: Skipping colorlog@6.8.2 because it requires Python>=3.6 but the project claims to work with all Python versions. Instead, another
version of colorlog that supports all Python versions will be used.
If you want to install colorlog@6.8.2, narrow down the `requires-python` range to include this version. For example, ">=3.6" should work.
  return self.repository.find_candidates(
<...>
```

[Full `pdm install -vv` log](https://github.com/pdm-project/pdm/files/14149362/pdm-install.log)

## Expected behavior

Project successfully imported, taking into account the `python_requires` option set in setup.cfg.

## Environment Information

```bash
$ pdm info && pdm info --env
PDM version:
  2.12.3+editable
Python Interpreter:
  /home/intelfx/devel/ext/liquidctl/.venv/bin/python (3.11)
Project Root:
  /home/intelfx/devel/ext/liquidctl
Local Packages:

{
  "implementation_version": "3.11.6",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.7.3-arch1pf5-2",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT_DYNAMIC Thu, 01 Jan 1970 00:00:00 +0000",
  "python_full_version": "3.11.6",
  "python_version": "3.11",
  "sys_platform": "linux"
}
```

intelfx:
## Pull Request Checklist

Add and use `Project.resolve_interpreter(in_import=True)` which does not memoize the result of resolution that uses a potentially incorrect `requires-python` information and does not create a virtualenv from this (potentially wrong) python version.

Additionally, the quality of informational output from `Project.resolve_interpreter()` is improved, hiding some misleading messages during the first resolution.

Fixes #2608.

### DIFF
--- a/src/pdm/cli/commands/import_cmd.py
+++ b/src/pdm/cli/commands/import_cmd.py
@@ -88,7 +88,8 @@ class Command(BaseCommand):
             pyproject["build-system"] = DEFAULT_BACKEND.build_system()
 
         if "requires-python" not in pyproject["project"]:
-            python_version = f"{project.python.major}.{project.python.minor}"
+            python = project.resolve_interpreter(in_import=True)
+            python_version = f"{python.major}.{python.minor}"
             pyproject["project"]["requires-python"] = f">={python_version}"
             project.core.ui.echo(
                 "The project's [primary]requires-python[/] has been set to [primary]>="
--- a/src/pdm/formats/setup_py.py
+++ b/src/pdm/formats/setup_py.py
@@ -17,7 +17,8 @@ def check_fingerprint(project: Project, filename: Path) -> bool:
 def convert(project: Project, filename: Path, options: Any | None) -> tuple[Mapping[str, Any], Mapping[str, Any]]:
     from pdm.models.in_process import parse_setup_py
 
-    parsed = parse_setup_py(str(project.environment.interpreter.executable), os.path.dirname(filename))
+    python = project.resolve_interpreter(in_import=True)
+    parsed = parse_setup_py(str(python.executable), os.path.dirname(filename))
     metadata: dict[str, Any] = {}
     settings: dict[str, Any] = {}
     for name in [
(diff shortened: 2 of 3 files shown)


frostming:
This is too complex IMO. What you need is to revert all side-effect caused by `Project.resolve_interpreter()`, this is enough:

```python
project._saved_python = None
project._python = None
```

intelfx:
@frostming Consider a project that has an upper-bound `requires-python` dependency, something like `requires-python < 3.10` (together with a virtualenv provider that can satisfy such a dependency).

In this case, the virtualenv will be created with a wrong python version, because it is created on the first resolution pass, where the bound is not yet known.

frostming:
I agree that `resolve_interpreter` has too much side-effects and it's bad.

We need a side-effect-free version of it, no writing files, not saving states, no virtualenv creation. but `in_import` isn't a good name for it.

And I wonder if the side-effect version can call the side-effect-free version inside.

intelfx:
Fair enough, I'll think about it!

frostming:
I am closing this PR, if you have further fixes feel free to open a new one.

PR #2655, opened 2024-02-28
Option to exclude optional groups from install/sync -G:all 

### LINKED ISSUE #2258: Allow opting-out from optional groups

I often want to install all optional groups except one, which contains e.g. expensive ML libraries I don't require. With poetry I could just `install --without ml`, but pdm doesn't seem to have any negative resolving logic.  As a workaround I can generate all the groups by parsing the pyproject.toml myself, but this feels like a hack.

It should be possible to install with all optional groups except some.  Something like:

```bash
pdm {install sync} -G:all -G "not expensive-group"
```

Alternatively, pdm could behave like poetry:

```bash
pdm {install sync} --with group --without group --without group
```

Apologies if this is already possible and I've just missed it.

elramen:
## Pull Request Checklist

Added a flag, ```--without```, that allows users to exclude specific optional groups when running ```pdm install/sync -G:all```, e.g. ```pdm install -G:all --without myGroup1,myGroup2```. 

Added alias ```--with``` for ```-G/--group```.

Closes #2258

### DIFF
--- a/news/2258.feature.md
+++ b/news/2258.feature.md
@@ -1,1 +1,1 @@
-Add option to exclude group(s) when running ```pdm sync/install -G:all``` by adding flag ```--exclude group1,group2,...```
+Add option to exclude group(s) when running ```pdm sync/install -G:all``` by adding flag ```--without group1,group2,...```
--- a/src/pdm/cli/options.py
+++ b/src/pdm/cli/options.py
@@ -220,6 +220,7 @@ groups_group = ArgumentGroup("Dependencies Selection")
 groups_group.add_argument(
     "-G",
     "--group",
+    "--with",
     dest="groups",
     metavar="GROUP",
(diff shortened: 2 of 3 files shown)


frostming:
Hi, I prefer `--without` as the option name, and we may add `--with` as an alias of `--group/-G` as well.

elramen:
> Hi, I prefer `--without` as the option name, and we may add `--with` as an alias of `--group/-G` as well.

On it! @frostming

elramen:
> Hi, I prefer `--without` as the option name, and we may add `--with` as an alias of `--group/-G` as well.

Done! :) @frostming

frostming:
🎉 Cheers. PR merged

2e0byo:
@elramen amazing, thank you very much.

PR #2659, opened 2024-02-29
Fix #2183 no timeout for PyPI calls

### LINKED ISSUE #2183: Add a timeout to Pypi calls

## Steps to reproduce
1. Create a vm with all outgoing packets droped by a firewall
  eg with `sudo iptables -A OUTPUT -j DROP`
2. Try to `pdm sync`, or `pdm self update`

## Actual behavior
PDM hangs indefinitely

## Expected behavior
A timout with a clear message should be shown. Maybe after 10 or 60 seconds ?

In my case the dev vm is in corporate network, without access to the internet, all packets are dropped. I forgot to update the `pdm config pypi.url` to our internal mirror thinking it would use the /etc/pip.conf config file. Having a timeout with an error message would have helped finding my mistake.

## Environment Information

```bash
pdm info
PDM version:
  2.8.2
Python Interpreter:
  /home/.../dev/misc/.venv/bin/python (3.9)
Project Root:
  /home/.../dev/misc
Local Packages:

pdm info --env
{
  "implementation_version": "3.9.16",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "4.18.0-477.15.1.el8_8.x86_64",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Fri Jun 2 08:27:19 EDT 2023",
  "python_full_version": "3.9.16",
  "python_version": "3.9",
  "sys_platform": "linux"
}
```

elramen:

Updated unearth dependency to version 0.14.0 to prevent pdm from hanging indefinitely when pypi.org is resolved but all other packets (i.e. non-DNS packets) are dropped.  

After investigation of issue #2183, we found that the problem was caused by the unearth package. However, it was solved in v. 0.14.0, specifically in commit [1d9a2dc](https://github.com/frostming/unearth/commit/1d9a2dc5279df60a9d3abda81af9a83590a80096). Therefore, this PR simply updates the unearth dependency to version 0.14.0. 

Closes #2183

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "virtualenv>=20",
     "pyproject-hooks",
     "requests-toolbelt",
-    "unearth==0.14.0",
+    "unearth>=0.14.0",
     "dep-logic>=0.2.0,<1.0",
     "findpython>=0.4.0,<1.0.0a0",
     "tomlkit>=0.11.1,<1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "virtualenv>=20",
     "pyproject-hooks",
     "requests-toolbelt",
-    "unearth>=0.12.1",
+    "unearth==0.14.0",
     "dep-logic>=0.2.0,<1.0",
     "findpython>=0.4.0,<1.0.0a0",
     "tomlkit>=0.11.1,<1",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = [
     "virtualenv>=20",
     "pyproject-hooks",
     "requests-toolbelt",
-    "unearth>=0.12.1",
+    "unearth==0.14.0",
```
frostming, line 21: please change it to a minimum constraint instead of an exact match. or is there any reason behind this?

frostming (changes requested):

pyproject.toml
elramen: Fixed :)

frostming (approved):

PR #2670, opened 2024-03-07
fix/group exclusion

2e0byo:
## Pull Request Checklist

Fixes unintuitive behaviour by making `--without` imply `--with :all` if no explicit groups given.

What should we do in the case of `--with foo --without foo`?  Currently if I read the logic right we silently drop it.

### DIFF
--- a/news/2670.bugfix.md
+++ b/news/2670.bugfix.md
@@ -1,0 +1,1 @@
+Made --without imply --with :all.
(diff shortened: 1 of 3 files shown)


frostming:
What groups will be installed if I run `pdm install —prod —without abc`?

2e0byo:
As I understand it, only default.  I'll add a test though.

2e0byo:
good catch.  Fixed.

frostming (approved):

PR #2689, opened 2024-03-13
feat: allow updating specific sub-deps with `--allow-transitive` flag

### LINKED ISSUE #2628: Option to update specific sub-dependency in lock file
## Is your feature/enhancement proposal related to a problem? Please describe.

At work, we use scanning tools to check if there are any known vulnerabilities in our images before we deploy. This is straightforward if the package is defined in the `pyproject.tom` using `pdm update`, but often the offending package is a sub-dependency that's only defined in the lock file.

For example, if a package like FastAPI has a vulnerability, I can do `pdm update fastapi`. If Starlette, a dependency of FastAPI, has a vulnerability, this does not work.

```
$ pdm update starlette                                                   
[ProjectError]: starlette does not exist in default dependencies.
 ```

I know I can do do eager updating on the entire file, but (to my knowledge) I can't limit this to a specific sub-dependency.

I'd love a feature similar to what `pdm update` currently does, but which only updates the versions in the lock file.

```
$ pdm <update-lock-command> starlette
```

dantevi:
## Pull Request Checklist

*copied from commit message*:

Add the `--allow-transitive` argument handler to the class `Command` and its `do_update` method in `update.py`.

Add logic for updating transitive dependencies in `update.py`. If the specified package is not in `pyproject.toml`, and the '--allow-transitive' flag is used, then try to match the package name with packages in the lock file instead. Ensures transitive dependencies are not written to `pyproject.toml`.

Add three tests for `--allow-transitive` flag.

Add shell completions for bash, fish, powershell and zshell.

Resolves #2628

### DIFF
--- a/news/2628.feature.md
+++ b/news/2628.feature.md
@@ -1,0 +1,1 @@
+Add the boolean flag `--allow-transitive` to the `update` command for allowing updating specific sub-dependencies (i.e., transitive dependencies) in the lock file.
--- a/src/pdm/cli/commands/update.py
+++ b/src/pdm/cli/commands/update.py
@@ -66,6 +66,13 @@ class Command(BaseCommand):
             action="store_false",
             help="Only update lock file but do not sync packages",
         )
+        parser.add_argument(
+            "--allow-transitive",
+            dest="allow_transitives",
+            default=False,
+            action="store_true",
+            help="Allow updating of transitive dependencies",
+        )
         parser.add_argument("packages", nargs="*", help="If packages are given, only update them")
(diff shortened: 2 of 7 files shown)


dantevi:
The commit is rebased to `main` and ready to be fast-forward merged as is.

frostming (commented):
I don't think we need an extra option for this.
Just fallback to update transitive dependencies and it should be harmless.

frostming:
Hi @dantevi can you make modifications to this PR?

PR #2711, opened 2024-03-21
chore: delete `setup.cfg`, move tool configurations under it to `pyproject.toml`

### LINKED ISSUE #2703: Move tool configurations under `setup.cfg` to `pyproject.toml`

This repo's [`setup.cfg`](https://github.com/pdm-project/pdm/blob/main/setup.cfg) file contains configurations about three tools: `codespell`, `coverage` and `mypy`.
According to their (newest) documentation, all three tools support to be configured via `pyproject.toml`:
* codespell: https://github.com/codespell-project/codespell?tab=readme-ov-file#using-a-config-file
* coverage: https://coverage.readthedocs.io/en/7.4.3/config.html#sample-file
* mypy: https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml

By moving configurations to `pyproject.toml`, we can **get rid of `setup.cfg`**, that's what I want, but I'm not sure if it's properly to do.

boholder:
Close #2703 

## Pull Request Checklist

~~- [ ] Test cases added for changed code.~~ N/A

As said in #2703, the goal is to remove`setup.cfg` file (and makes the tool configuration a little bit cleaner).

Result of `pdm lint` only shows two Windows special falsely positives (I'm working on a Windows PC) (because `os` module of cpython on windows [do not have `uname` API](https://docs.python.org/3/library/os.html#os.uname)):

```log
***\pdm>pdm lint
pyupgrade................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
codespell................................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src\pdm\environments\base.py:173: error: Module has no attribute "uname"; maybe "name"?  [attr-defined]
src\pdm\environments\base.py:182: error: Module has no attribute "uname"; maybe "name"?  [attr-defined]
Found 2 errors in 1 file (checked 111 source files)
```

I'm not sure how to test coverage config, let's see the bot's result.


frostming:
Pre-commit linters are happy. But please fix the reference of `setup.cfg` in `ci.yml`

boholder:
> But please fix the reference of `setup.cfg` in `ci.yml`
> 
> Also please add a news fragment under `news/`

Yep, now I see where is wrong, will working on it.

PR #2766, opened 2024-04-03
fix: refuse to run recursive composite scripts

j178:
## Pull Request Checklist

If user defined a recursive composite script like this:

```toml
[tool.pdm.scripts]
first = { "composite" = ["first"] }
```

Then pdm will get into a resurive loop, resulting in a `RecursionError`:

```sh
> pdm run first
[RecursionError]: maximum recursion depth exceeded
WARNING: Add '-v' to see the detailed traceback
```

This PR adds a check for composite script, if one is already seen, then raise an error.


frostming (approved):
Thank you, good catch.

PR #2786, opened 2024-04-09
fix: add all sources to trusted_hosts if verify_ssl is False

### LINKED ISSUE #2784: Skipping <url> for not being trusted, please add it to `trusted_hosts` list starting from PDM 2.13.3

## Steps to reproduce

I'm using pdm 2.13.3, and I have a source configured with the name "pypi" and verify_ssl = false. The URL is replaced with a fake generic one, but it's also using http instead of https.
```
[[tool.pdm.source]]
name = "pypi"
url = "http://some.url/simple"
verify_ssl = false
```

There are no other sources configured.

## Actual behavior

When I run `pdm sync -v --prod`, I see this warning message and stack trace
```
unearth.collector: Skipping http://some.url/simple/setuptools/ for not being trusted, please add it to `trusted_hosts` list
  ✖ Install setuptools 69.2.0 failed

  File "/usr/local/bin/pdm", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/pdm/core.py", line 341, in main
    return core.main(args or sys.argv[1:])
  File "/usr/local/lib/python3.9/dist-packages/pdm/core.py", line 259, in main
    raise cast(Exception, err).with_traceback(traceback) from None
  File "/usr/local/lib/python3.9/dist-packages/pdm/core.py", line 254, in main
    self.handle(project, options)
  File "/usr/local/lib/python3.9/dist-packages/pdm/core.py", line 209, in handle
  File "/usr/local/lib/python3.9/dist-packages/pdm/cli/commands/sync.py", line 44, in handle
    actions.do_sync(
  File "/usr/local/lib/python3.9/dist-packages/pdm/cli/actions.py", line 238, in do_sync
    synchronizer.synchronize()
  File "/usr/local/lib/python3.9/dist-packages/pdm/installers/synchronizers.py", line 445, in synchronize
    handlers[kind](key, progress)
  File "/usr/local/lib/python3.9/dist-packages/pdm/installers/synchronizers.py", line 282, in install_candidate
    self.manager.install(can)
  File "/usr/local/lib/python3.9/dist-packages/pdm/installers/manager.py", line 33, in install
    prepared.get_cached_package(),
  File "/usr/local/lib/python3.9/dist-packages/pdm/models/candidates.py", line 404, in get_cached_package
    self._obtain(allow_all=False)
  File "/usr/local/lib/python3.9/dist-packages/pdm/models/candidates.py", line 466, in _obtain
    raise CandidateNotFound(
pdm.exceptions.CandidateNotFound: No candidate is found for `setuptools` that matches the environment or hashes
```

It seems like my custom pypi source stopped being recognized as a trusted host.

## Expected behavior

pdm sync worked when I was using 2.13.2, but it doesn't in 2.13.3

## Environment Information

```bash
PDM version:
  2.13.3
Python Interpreter:
  /.venv/bin/python (3.9)
Project Root:
  /
Local Packages:

{
  "implementation_version": "3.9.18",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.10.186-179.751.amzn2.x86_64",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Tue Aug 1 20:51:38 UTC 2023",
  "python_full_version": "3.9.18",
  "python_version": "3.9",
  "sys_platform": "linux"
}
```

rjchee:
## Pull Request Checklist

Fix #2784.
- Move the logic to update `session._trusted_host_ports` before checking the `pypi` source so the pypi source can also be trusted if necessary
- Check that `verify_ssl is False` before adding to the trusted host ports which seems in line with how things worked before pdm 2.13.3

I'm not familiar enough with this codebase to know how to properly add unit tests, but I did verify that `pdm sync` works with my custom pypi source after this change.

### DIFF
--- a/src/pdm/models/session.py
+++ b/src/pdm/models/session.py
@@ -134,15 +134,18 @@ class PDMPyPIClient(PyPIClient):
         self._trusted_host_ports: set[tuple[str, int | None]] = set()
         transport: httpx.BaseTransport | None = None
         for s in sources:
+            url = None
+            if s.url is not None:
+                url = httpx.URL(s.url)
+                if s.verify_ssl is False:
+                    self._trusted_host_ports.add((url.host, url.port))
             if s.name == "pypi":
                 transport = self._transport_for(s)
                 continue
-            assert s.url is not None
-            url = httpx.URL(s.url)
+            assert url is not None
             mounts[f"{url.scheme}://{url.netloc.decode('ascii')}/"] = hishel.CacheTransport(
                 self._transport_for(s), storage, controller
             )
-            self._trusted_host_ports.add((url.host, url.port))
         mounts.update(kwargs.pop("mounts", None) or {})
 
         httpx.Client.__init__(self, mounts=mounts, follow_redirects=True, transport=transport, **kwargs)
--- a/src/pdm/models/session.py
+++ b/src/pdm/models/session.py
@@ -134,15 +134,13 @@ class PDMPyPIClient(PyPIClient):
         self._trusted_host_ports: set[tuple[str, int | None]] = set()
         transport: httpx.BaseTransport | None = None
         for s in sources:
-            url = None
-            if s.url is not None:
-                url = httpx.URL(s.url)
-                if s.verify_ssl is False:
-                    self._trusted_host_ports.add((url.host, url.port))
+            assert s.url is not None
+            url = httpx.URL(s.url)
+            if s.verify_ssl is False:
+                self._trusted_host_ports.add((url.host, url.port))
             if s.name == "pypi":
                 transport = self._transport_for(s)
                 continue
-            assert url is not None
             mounts[f"{url.scheme}://{url.netloc.decode('ascii')}/"] = hishel.CacheTransport(
                 self._transport_for(s), storage, controller
             )
--- a/news/2784.bugfix.md
+++ b/news/2784.bugfix.md
@@ -1,0 +1,1 @@
+Check verify_ssl when trusting each source.


src/pdm/models/session.py
```diff
@@ -134,15 +134,18 @@ def __init__(self, *, sources: list[RepositoryConfig], cache_dir: Path, **kwargs
         self._trusted_host_ports: set[tuple[str, int | None]] = set()
         transport: httpx.BaseTransport | None = None
         for s in sources:
+            url = None
+            if s.url is not None:
+                url = httpx.URL(s.url)
+                if s.verify_ssl is False:
+                    self._trusted_host_ports.add((url.host, url.port))
```
frostming, line 141: I have a trouble understanding this fix. This moves the unconditionally adding trust host to a conditional one, then how can I fix the problem?

rjchee: Sorry for making this unclear. I ended up making 2 slightly unrelated changes here.

The first is to move the logic to add a source's URL to `self._trusted_host_ports` to **before** we check for the `pypi` name. This is the part that fixes my issue.

The second is that we add the check for `verify_ssl is False` before we actually add to `self._trusted_host_ports`, which seems to be the behavior for `pdm.utils.get_trusted_hosts` which this code was using prior to `2.13.3`. This check doesn't affect my use case since I explicitly set `verify_ssl = false` in my `pyproject.toml` config for my source. I added it because I thought it was a bit strange to trust sources unconditionally. However, I'm definitely not an expert in this codebase or the best practices around this sort of thing so I'm definitely open to removing this check if you think it's not necessary.

frostming: thanks, that explains

src/pdm/models/session.py
```diff
@@ -134,15 +134,18 @@ def __init__(self, *, sources: list[RepositoryConfig], cache_dir: Path, **kwargs
         self._trusted_host_ports: set[tuple[str, int | None]] = set()
         transport: httpx.BaseTransport | None = None
         for s in sources:
+            url = None
+            if s.url is not None:
```
frostming, line 138: this check is unnecessary. we just assert source.URL is non-null

rjchee: Is there a case where a `pypi` source doesn't have a URL? If so I'm not sure if it would make sense to add it to the `_trusted_host_ports` since there's no URL to add

frostming: yes. but not in this specific function. it always have a URL for a source

frostming:
Thanks for following up this

rjchee:
Thanks for your quick feedback on this!

PR #2787, opened 2024-04-10
fix: name check for project itself in `pdm outdated`

### LINKED ISSUE #2785: pdm outdated shows self package when project.name contains underscores

## Steps to reproduce

In a project whose name contains underscores, run the command `pdm outdated`.

pyproject.toml
```toml
[project]
name = "package_name"
```

## Actual behavior

A table is printed which includes the self package.

```
╭──────────────┬────────────────┬────────┬────────╮
│ Package      │ Installed      │ Pinned │ Latest │
├──────────────┼────────────────┼────────┼────────┤
│ package-name │ 0.8.1+editable │        │ 0.8.1  │
╰──────────────┴────────────────┴────────┴────────╯
```

## Expected behavior

The self package should not be included in the output. It works properly after replacing underscores with dashes in `project.name`.

## Environment Information

```bash
❯ pdm info && pdm info --env
PDM version:
  2.13.3
Python Interpreter:
  /home/messingm/git/cmdb-client/.venv/bin/python (3.9)
Project Root:
  /home/messingm/git/cmdb-client
Local Packages:
  
{
  "implementation_version": "3.9.18",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "5.15.146.1-microsoft-standard-WSL2",
  "platform_system": "Linux",
  "platform_version": "#1 SMP Thu Jan 11 04:09:03 UTC 2024",
  "python_full_version": "3.9.18",
  "python_version": "3.9",
  "sys_platform": "linux"
}
```

huxuan:
## Pull Request Checklist

This should fix #2785

1. Wrap `normalize_name` for `project.name` before name checking.
2. Add `lru_cache` for `normalize_name` function. 

<del>but I am not sure whether this is the best solution.</del>

<del>The cached property `normalized_name` for `Project` is added as the `normalize_name()` need to be triggered for multiple times in the for-loop and there are several other existence of `normalize_name(project.name)` in the repo. If this is OK, a follow up would be change those `normalize_name(project.name)` to `project.normalized_name`.</del>

<del>Other possible options:</del>
<del>1. Use a local variable, e.g., `normalized_project_name`, before the for-loop and use it inside.</del>
<del>3. Add `@lru_cache` for `normalize_name()`. </del>

BTW, I wonder the reason to use the normalized name rather than `distribution.name` directly as the key in `WorkingSet`.

### DIFF
--- a/news/2785.bugfix.md
+++ b/news/2785.bugfix.md
@@ -1,0 +1,1 @@
+Fix name check for project itself in `pdm outdated`
--- a/src/pdm/cli/commands/outdated.py
+++ b/src/pdm/cli/commands/outdated.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 from pdm.cli.commands.base import BaseCommand
 from pdm.cli.utils import normalize_pattern
 from pdm.models.requirements import strip_extras
+from pdm.utils import normalize_name
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace
@@ -95,7 +96,7 @@ class Command(BaseCommand):
         for name, distribution in installed.items():
             if not self._match_pattern(name, options.patterns):
                 continue
-            if name == project.name:
+            if name == normalize_name(project.name):
                 continue
             constrained_version = resolved.pop(name).version or "" if name in resolved else ""
             collected.append(ListPackage(name, distribution.version or "", constrained_version))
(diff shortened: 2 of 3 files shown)


src/pdm/project/core.py
```diff
@@ -148,6 +149,10 @@ def project_config(self) -> Config:
     def name(self) -> str:
         return self.pyproject.metadata.get("name")
 
+    @cached_property
```
frostming, line 152: `self.pyproject` is subject to change, don't cache this. Instead, you can add cache to the `normalize_name` function

huxuan: Done.

PR #2840, opened 2024-04-19
fix(cache): ensure installation and cache works with any environment

### LINKED ISSUE #2838: Command completion scripts produces output when a new version is available (at least on oh-my-zsh)
Command completion scripts should never produce any output; it interferes with command completion. Command completion on my machine with pdm 2.13.3 on macos, with oh-my-zsh however produced an update notice.

## Steps to reproduce
 - Install an outdated version of pdm with homebrew
 - Enable pdm command completion in oh-my-zsh
 - Make sure `check_update` is enabled (default)
 - Use command completion
 
## Actual behavior

Pdm's command completion for oh-my-zsh prints the following, interrupting command completion:
```
INFO: PDM 2.13.3 is installed, while 2.15.0 is available.
Please run `brew upgrade pdm` to upgrade.
Run `pdm config check_update false` to disable the check.
```

## Expected behavior

Normal command completion happens, no output from pdm.

## Environment Information

Unfortunately, I already updated pdm, so this is post-update, not when I observed the behaviour.
```
PDM version:
  2.15.0
Python Interpreter:
  /Users/yves/epymetrics/graphical-interfaces/emda2-streamlit/.venv/bin/python (3.12)
Project Root:
  /Users/yves/epymetrics/graphical-interfaces/emda2-streamlit
Local Packages:
  
{
  "implementation_version": "3.12.2",
  "os_name": "posix",
  "platform_machine": "arm64",
  "platform_release": "23.4.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000",
  "python_full_version": "3.12.2",
  "python_version": "3.12",
  "sys_platform": "darwin"
}
```

noirbizarre:
## Pull Request Checklist

This PR disable check update on commands with an expected output format to avoid breaking it (`export`, `info` with a flag, `show` with a flag, `completions`)

Fixes #2838

### DIFF
--- a/news/2838.bugfix.md
+++ b/news/2838.bugfix.md
@@ -1,0 +1,1 @@
+Disable check update on commands with an expected output format.
--- a/src/pdm/cli/commands/completion.py
+++ b/src/pdm/cli/commands/completion.py
@@ -33,3 +33,6 @@ class Command(BaseCommand):
         completion = resources_read_text("pdm.cli.completions", f"pdm.{suffix}")
         # Can't use rich print or otherwise the rich markups will be interpreted
         print(completion.replace("%{python_executable}", sys.executable))
+
+        # Disable check update to avoid broken completions
+        project.can_check_update = False
--- a/src/pdm/cli/commands/export.py
+++ b/src/pdm/cli/commands/export.py
@@ -83,3 +83,4 @@ class Command(BaseCommand):
         else:
             # Use a regular print to avoid any formatting / wrapping.
             print(content)
+        project.can_check_update = False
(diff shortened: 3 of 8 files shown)


frostming:
I think the original issue was about running the `pdm` command in zsh completion script, which would print an update prompt and cause completion to fail.

You should add `PDM_CHECK_UPDATE=no` to `pdm.zsh`.

Issue #2760, opened 2024-04-02
Fail to authenticate for Azure Artifacts


westr029:
## Steps to reproduce
To add a private pypi from Azure Artifacts, I added a section in the `pyproject.toml` as follows:

```
[[tool.pdm.source]]
name = "datahub"
url = "https://pkgs.dev.azure.com/xxx/Datahub/_packaging/Datahub/pypi/simple/"
verify_ssl = true
```

I try to add a package to the pyproject.toml that is available in the private repo by:
`pdm add datahub-cli==0.2.0a6 -v`

## Actual behavior
When I run the above command, I get the message below:
```
Adding packages to default dependencies: datahub-cli==0.2.0a6
STATUS: Resolving dependencies
pdm.termui: ======== Start resolving requirements ========
pdm.termui:   datahub-cli==0.2.0a6
pdm.termui:   python==3.12.*
pdm.termui:   Adding requirement datahub-cli==0.2.0a6
unearth.collector: Failed to collect links from https://pypi.org/simple/datahub-cli/: Client Error(404): Not Found

Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.IdentityModel.Clients.ActiveDirectory, Version=5.2.9.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.
   at NuGetCredentialProvider.Program.<Main>d__3.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at NuGetCredentialProvider.Program.Main(String[] args)
   at NuGetCredentialProvider.Program.<Main>(String[] args)
unearth.auth: Keyring is skipped due to an exception: Failed to get credentials: process with PID 24148 exited with code 3762504530; additional error message: 
User for pkgs.dev.azure.com:
```
However, when I pass personal login information via the `pyproject.toml` in the `tool.pdm.source` section (which I don't want), it does work. 

The same error also appears, when I want to install a package from a public pypi, which is very inconvenient.

## Expected behavior
Package from private pypi is added to `pyproject.toml` and installed.

## Environment Information
`pdm info; pdm info --env `:

```bash
PDM version:
  2.13.2
Python Interpreter:
  C:\Users\avanwestreenen\OneDrive - Waterschap Vallei Veluwe\Projects\PlayGround2\.venv\Scripts\python.exe (3.12)
[...]

PR #2854, opened 2024-04-26
Fix activate for windows and updated docs for Azure

Pluttodk:
## Pull Request Checklist

Altered the `pdm venv activate` to not include `"."` on windows, since this causes an error. This also ensures, that for non Powershell users on windows, the outputted command can be directly pasted into shell.

Additionally added documentation as to how to add authentication to Azure Artifacts, since this seems to be a common question asked. Such as: #2760


src/pdm/cli/commands/venv/activate.py
```diff
@@ -55,6 +56,8 @@ def get_activate_command(self, venv: VirtualEnv) -> str:  # pragma: no cover
             command, filename = "source", "activate"
         activate_script = venv.interpreter.with_name(filename)
         if activate_script.exists():
+            if platform.system() == "Windows":
+                return f"{shlex.quote(str(activate_script))}"
```
frostming, line 60: look above several lines, we have a case for powershell. this will completely bypass it.

Pluttodk: This only works for powershell 7 and not 5.1. Neither does the old solution work for Batch.

I did not want to remove the previous check with powershell, since you can install powershell on linux (not sure why you would want that but you can) and that supports the dot command.

Output from powershell 5.1
```ps1
> Invoke-Expression (pdm venv activate)
C:\Code\data_indhentning\.venv\Scripts\Activate.ps1 : Cannot dot-source this command because it was defined in a differ
ent language mode. To invoke this command without importing its contents, omit the '.' operator.
At line:1 char:1
+ . 'C:\Code\.venv\Scripts\Activate.ps1'
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Activate.ps1], NotSupportedException
    + FullyQualifiedErrorId : DotSourceNotSupported,Activate.ps1
```

PR #2954, opened 2024-06-19
Use project name for venv prompt

gsiddhant159:


src/pdm/cli/commands/venv/backends.py
```diff
@@ -124,7 +124,7 @@ def create(
         args = (*self.pip_args(with_pip), *args)
         if prompt is not None:
             prompt = prompt.format(
-                project_name=self.project.root.name.lower() or "virtualenv",
+                project_name=self.project.name.lower() or "virtualenv",
```
frostming, line 127: Please guarantee `project.name` is not null and fallback to `project.root.name`

PR #2959, opened 2024-06-20
docs(publish): add note on PyPI API tokens in docs/usage/publish.md

vvanglro:
## Pull Request Checklist

### DIFF
--- a/docs/usage/publish.md
+++ b/docs/usage/publish.md
@@ -8,6 +8,8 @@ pdm publish
 
 This will automatically build a wheel and a source distribution(sdist), and upload them to the PyPI index.
 
+PyPI requires API tokens to publish packages, you can use `__token__` as the username and API token as the password.
+
 To specify another repository other than PyPI, use the `--repository` option, the parameter can be either the upload URL or the name of the repository stored in the config file.
 
 ```bash


vvanglro:
```sh
pdm publish --help  
  -u USERNAME, --username USERNAME
                        The username to access the repository [env var: PDM_PUBLISH_USERNAME]
  -P PASSWORD, --password PASSWORD
                        The password to access the repository [env var: PDM_PUBLISH_PASSWORD]
```
The tips here should perhaps be changed as well?

frostming:
Oh, so the username and password are no longer allowed to publish package? Then it needs to be updated.

vvanglro:
I'm still looking for news that pypi is officially not supporting username and password release packages, but haven't found it yet. Instead, I'm running into a lot of the same problems with what appears to be pypi's unpublished instructions?

frostming:
> I'm still looking for news that pypi is officially not supporting username and password release packages, but haven't found it yet. Instead, I'm running into a lot of the same problems with what appears to be pypi's unpublished instructions?

That's fine, but some private pypi instances may still allow username and password

vvanglro:
It seems that after turning on 2FA, you must use a token to publish.

https://discuss.python.org/t/announcement-2fa-now-required-for-pypi/42251/4

vvanglro:
> That's fine, but some private pypi instances may still allow username and password. I think it's fine to make it default to `__token__` for now.

We pointed out in the docs that if you want to publish the package to the official pypi, you must use the token method.

vvanglro:
Or cli adds an api token option to differentiate from username/password.🤔

Issue #2943, opened 2024-06-14
[Feature Request] Take "requires-python" into account for installing and using Python interpreter


o-moe:
## Motivation
- Streamline / automate the set-up process for new local virtual environments.
- Use a different Python interpreter for managing the project with pdm and developing the project.

## Use case
An existing project using pdm for dependency management is cloned/cleaned by a development team member (no .venv and .pdm-python available). The pyproject.toml already defines a (minimum Python version for the project) by the `requires-python`attribute.

## Feature request
Let `pdm python install` and `pdm use` respect the (minimum) Python version given in the pyproject.toml file (if existing) as default. Prefer cPython interpreters for downloading and using.
Additionally, add a `--automatic-install` option to `pdm use` that automatically installs the default version (from pyproject.toml) or from the user provided version.

## Expected behavior
- The user can work with `pdm python install` and `pdm use` as before if the Python version is provided.
- If a pyproject.toml is already existing that defines a (minimum) Python version, the user can automate installation and usage of this version by `pdm use --automatic-install` without any further interaction.
  - I don't mind having a `-y`option for `--automatic-install` 

### Discussed in https://github.com/pdm-project/pdm/discussions/2938

<div type='discussions-op-text'>

<sup>Originally posted by **o-moe** June 12, 2024</sup>
Hi everyone,

sorry if this is obvious but I am new to pdm and my knowledge about Python in general is limited.

Recently, I started looking into pdm to manage my project - and it look very promising. Thanks to you @frostming.
The project sources are under version control. When I follow pdm's documentation recommendation, I will commit pyproject.toml, pdm.toml and pdm.lock to have the environment reproducible across members working with my project.

But what makes me wonder is, how do I make sure that everybody uses the same Python interpreter that the project is based on (and has been initialized)? If the members just use `pdm use` after their initial checkout, they can use whatever version is available on their system. If the members use `pdm python install` prior to `pdm use`, they can choose whatever version they like.
I trust the members of my project but I would like to have the Python interpreter version controlled by the same principles like I control the dependencies and configuration with the pyproject.toml and the pdm.lock file, i.e. reproducible and failure-safe.

Am I missing something here? Is pdm supporting this use case and I just didn't realize it? Is this use case uncommon?

BTW: I also want the Python interpreter of the project being explicitly different  from the Python interpreter that pdm is using.

Thanks in advance for your thoughts.</div>

frostming:
Sorry, can you explain how `requires-python` is not being respected?

[...]

PR #2969, opened 2024-06-24
feat: add unattended installation/using of cPython defined by `requires-python` for `pdm use` (#2943)

o-moe:
## Pull Request Checklist

While maintaining backwards compatibility to the best I could, `pdm use` and `pdm python install` will now automatically install cPython based on a "best match" strategy which takes requires-python from pyproject.toml into account (if available). The behavior from `pdm use` up to this change was to take latest/greatest available for installation that met `requires-python` - I modified this to also take the platform/arch into account to apply the same "best match" strategy everywhere.

Additionally, I added some options for unattended installation which should streamline CI/CD or fresh checkout use cases:

`pdm use --auto-install-min` will then install and use the minimum `requires-python` version available for the current platform/arch without any user-interaction (just like `-f` did for the first found interpreter)

`pdm use --auto-install-max` will then install and use the maximum `requires-python` version available for the current platform/arch without any user-interaction (just like `-f` did for the first found interpreter)

Since `-f` and the new unattended options for `pdm use` do not go well together, I also put them into a mutually exclusive options group for better CLI experience.


o-moe:
@frostming I wanted to reference this PR to the issue (#2943), but couldn't see how (it's my first PR in Github)

docs/usage/project.md
```diff
@@ -57,6 +57,25 @@ pdm python remove 3.9.8
 
     Afterwards you can manage the installations using either `rye toolchain` or `pdm python`.
 
+### Installation strategy based on `requires-python`
+
++++ 2.15.5
```
frostming, line 62:
```suggestion
+++ 2.16.0
```

This is what the next version planned to be

src/pdm/cli/commands/python.py
```diff
@@ -97,7 +108,20 @@ def handle(self, project: Project, options: Namespace) -> None:
             for version in PYTHON_VERSIONS:
                 project.core.ui.echo(str(version))
             return
-        self.install_python(project, options.version)
+        version = options.version
+        if version is None:
+            match = (
+                project.get_min_matching_cpython_version()
+                if options.min
+                else project.get_best_matching_cpython_version()
```
frostming, line 116: It's not worth for two separate methods, just merge them into one and switch via an argument.

src/pdm/cli/commands/use.py
```diff
@@ -70,6 +84,25 @@ def version_matcher(py_version: PythonInfo) -> bool:
                 project.core.ui.info("Using the last selection, add '-i' to ignore it.")
                 return cached_python
 
+        if not python and not first and (auto_install_min or auto_install_max):
```
frostming, line 87: Why should it be preferred than existing interpreters? Wouldn't it cause unnecessary installations when I run it more than once?

o-moe: I agree that this could be optimized to figure out, if the existing ones already meet "auto install" requirements (pdm installed cPython version). But then again, the installation function is already taking care of this (if already present then it is re-used to my understanding).
I explicitly wanted to use only pdm installed cPython interpreter for the "auto install" functionality to have an independent interpreter be used to avoid side-effects (conda environments etc.). This is also a good approach for CI/CD pipelines.
Since the original behavior is still present, a dedicated Python interpreter may be chosen or selected by user interaction if the "auto install" is not used.

src/pdm/cli/commands/python.py
o-moe: done

src/pdm/project/core.py
```diff
@@ -12,6 +12,7 @@
 from typing import TYPE_CHECKING, Any, Callable, Iterable, Mapping, cast
 
 import tomlkit
+from pbs_installer import PythonVersion
```
frostming, line 15: Move this inside, too.

o-moe:
Anything on my side regarding the failing [Test / Testing (3.13, windows-latest, pip)](https://github.com/pdm-project/pdm/actions/runs/9643787786/job/26595023948?pr=2969) job?

The original error is apparently an installation error from this dependency:

```
pdm.termui: building 'msgpack._cmsgpack' extension
pdm.termui: error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
  x Install msgpack 1.0.8 failed
```

frostming:
> Anything on my side regarding the failing [Test / Testing (3.13, windows-latest, pip)](https://github.com/pdm-project/pdm/actions/runs/9643787786/job/26595023948?pr=2969) job?

Nope, LGTM

PR #2973, opened 2024-06-25
fix: pbs_installer interface (build_dir) for auto-install best match strategy

o-moe:
## Pull Request Checklist

Fix for new pbs_installer interface regarding `build_dir` and best match auto-install strategy for `pdm-use`. Basically the same fix as for `pdm python install --list` recently.

### DIFF
--- a/news/2943.bugfix.md
+++ b/news/2943.bugfix.md
@@ -1,0 +1,2 @@
+Fix new interface from pbs_installer regarding `build_dir` and best match auto-install strategy for `pdm use`
+(same as for `pdm python install --list`)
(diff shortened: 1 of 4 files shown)


o-moe:
@frostming shall new `news` entry be changed to reflect this PR #? Currently I used my original feature request issue #

o-moe:
This "quick fix" makes me wonder if `pdm python install --list` should better share the same implementation to get the candidates as the best match auto-install functionality (which would filter additionally). So that in the future - if pbs_installer changes again, all CLI commands benefit from a fix

frostming:
> This "quick fix" makes me wonder if `pdm python install --list` should better share the same implementation to get the candidates as the best match auto-install functionality (which would filter additionally). So that in the future - if pbs_installer changes again, all CLI commands benefit from a fix

Sure, please move it to `utils` module as it does not depend on `self`.

Also please update `pdm.lock` and set the minimum version of `pbs_installer` to `2024.4.18`

PR #2975, opened 2024-06-26
Fix incompatibility with importlib-metadata 8

### LINKED ISSUE #2974: 2.16.0 regression - `KeyError: 'Requires-Python'` crash when running `pdm add` in github actions on python 3.8
- [X] I have searched the issue tracker and believe that this is not a duplicate.

when running the `pdm add` command on python 3.8 in github actions, it crashes with the following error:

```
KeyError: 'Requires-Python'
```

i can't reproduce this locally, but it happens in github actions (windows-latest, ubuntu-latest and macos-latest) on python 3.8 but not 3.12. this bug seems to have been introduced in pdm 2.16.0

## Steps to reproduce

1. clone https://github.com/DetachHead/test/tree/0b2d5a592e30f2bfc76d8fddedf73b0ef17d6a4b
1. run the following commands in github actions on python 3.8:
   ```
   ./pw pdm install -v
   ./pw pdm add pytest==3.8 -v
   ```
## Actual behavior
see github actions for the full output: https://github.com/DetachHead/test/actions/runs/9671949594/job/26683445102
```
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/bin/pdm", line 8, in <module>
    sys.exit(main())
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/core.py", line 363, in main
    return core.main(args or sys.argv[1:])
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/core.py", line 281, in main
    raise cast(Exception, err).with_traceback(traceback) from None
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/core.py", line 276, in main
    self.handle(project, options)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/core.py", line 212, in handle
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/cli/commands/add.py", line 69, in handle
    self.do_add(
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/cli/commands/add.py", line 162, in do_add
    resolved = do_lock(
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/cli/actions.py", line 101, in do_lock
    mapping, dependencies = resolve(
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/resolver/core.py", line 39, in resolve
    result = resolver.resolve(requirements, max_rounds)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/resolvelib/resolvers.py", line 546, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/resolvelib/resolvers.py", line 426, in resolve
    name = min(unsatisfied_names, key=self._get_preference)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/resolvelib/resolvers.py", line 203, in _get_preference
    return self._p.get_preference(
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/resolver/providers.py", line 135, in get_preference
    deps = self.get_dependencies(candidate)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/resolver/providers.py", line 272, in get_dependencies
    deps, requires_python, _ = self.repository.get_dependencies(candidate)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/models/repositories.py", line 90, in get_dependencies
    requirements, requires_python, summary = getter(candidate)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/models/repositories.py", line 48, in wrapper
    result = func(self, candidate)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/models/repositories.py", line 270, in _get_dependencies_from_metadata
    deps = prepared.get_dependencies_from_metadata()
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/models/candidates.py", line 625, in get_dependencies_from_metadata
    return filter_requirements_with_extras(self.metadata.requires or [], extras)
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/pdm/models/candidates.py", line 618, in metadata
    self.candidate.requires_python = cast(str, result.metadata["Requires-Python"] or "")
  File "/home/runner/work/test/test/.pyprojectx/venvs/pdm-2ad7fbee887cffa1265ac100532e45d2-py3.8/lib/python3.8/site-packages/importlib_metadata/_adapters.py", line 54, in __getitem__
    raise KeyError(item)
KeyError: 'Requires-Python'
```

## Expected behavior

no error

## Environment Information

```bash
{
  "implementation_version": "3.8.18",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.5.0-1022-azure",
  "platform_system": "Linux",
  "platform_version": "#23~22.04.1-Ubuntu SMP Thu May  9 17:59:24 UTC 2024",
  "python_full_version": "3.8.18",
  "python_version": "3.8",
  "sys_platform": "linux"
}
```

JelleZijlstra:
Fixes #2974

## Pull Request Checklist

Still running the test suite locally to verify whether existing tests catch this.

### DIFF
--- a/src/pdm/cli/commands/list.py
+++ b/src/pdm/cli/commands/list.py
@@ -309,6 +309,19 @@ def parse_comma_separated_string(
     return items
 
 
+def _get_metadata_key(dist: im.Distribution, key: str) -> str | None:
+    """Get a key from metadata, or return None if it is not found or 'UNKNOWN'."""
+    # Starting with importlib-metadata 8.0, KeyError is thrown for missing keys.
+    try:
+        value = dist.metadata[key]
+    except KeyError:
+        return None
+    else:
+        if not value or value == "UNKNOWN":
+            return None
+        return value
+
+
 class Listable:
     """Wrapper makes sorting and exporting information about a Distribution
     easier.  It also retrieves license information from dist-info metadata.
@@ -322,20 +335,17 @@ class Listable:
     def __init__(self, dist: im.Distribution, groups: set[str]):
         self.dist = dist
 
-        self.name: str | None = dist.metadata["Name"]
+        self.name = _get_metadata_key(dist, "Name")
         self.groups = "|".join(groups)
 
-        self.version: str | None = dist.metadata["Version"]
-        self.version = None if self.version == "UNKNOWN" else self.version
+        self.version = _get_metadata_key(dist, "Version")
 
-        self.homepage: str | None = dist.metadata["Home-Page"]
-        self.homepage = None if self.homepage == "UNKNOWN" else self.homepage
+        self.homepage = _get_metadata_key(dist, "Home-Page")
 
         # If the License metadata field is empty or UNKNOWN then try to
         # find the license in the Trove classifiers.  There may be more than one
         # so generate a pipe separated list (to avoid complexity with CSV export).
-        self.licenses: str | None = dist.metadata["License"]
-        self.licenses = None if self.licenses == "UNKNOWN" else self.licenses
+        self.licenses = _get_metadata_key(dist, "License")
 
         # Sometimes package metadata contains the full license text.
         # e.g. license = { file="LICENSE" } in pyproject.toml
--- a/src/pdm/cli/commands/list.py
+++ b/src/pdm/cli/commands/list.py
@@ -309,19 +309,6 @@ def parse_comma_separated_string(
     return items
 
 
-def _get_metadata_key(dist: im.Distribution, key: str) -> str | None:
-    """Get a key from metadata, or return None if it is not found or 'UNKNOWN'."""
-    # Starting with importlib-metadata 8.0, KeyError is thrown for missing keys.
-    try:
-        value = dist.metadata[key]
-    except KeyError:
-        return None
-    else:
-        if not value or value == "UNKNOWN":
-            return None
-        return value
-
-
 class Listable:
     """Wrapper makes sorting and exporting information about a Distribution
     easier.  It also retrieves license information from dist-info metadata.
@@ -335,17 +322,20 @@ class Listable:
     def __init__(self, dist: im.Distribution, groups: set[str]):
         self.dist = dist
 
-        self.name = _get_metadata_key(dist, "Name")
+        self.name = dist.metadata.get("Name")
         self.groups = "|".join(groups)
 
-        self.version = _get_metadata_key(dist, "Version")
+        self.version = dist.metadata.get("Version")
+        self.version = None if self.version == "UNKNOWN" else self.version
 
-        self.homepage = _get_metadata_key(dist, "Home-Page")
+        self.homepage = dist.metadata.get("Home-Page")
+        self.homepage = None if self.homepage == "UNKNOWN" else self.homepage
 
         # If the License metadata field is empty or UNKNOWN then try to
         # find the license in the Trove classifiers.  There may be more than one
         # so generate a pipe separated list (to avoid complexity with CSV export).
-        self.licenses = _get_metadata_key(dist, "License")
+        self.licenses = dist.metadata.get("License")
+        self.licenses = None if self.licenses == "UNKNOWN" else self.licenses
 
         # Sometimes package metadata contains the full license text.
         # e.g. license = { file="LICENSE" } in pyproject.toml
--- a/src/pdm/models/candidates.py
+++ b/src/pdm/models/candidates.py
@@ -615,7 +615,12 @@ class PreparedCandidate:
             if not self.candidate.version:
                 self.candidate.version = result.version
             if not self.candidate.requires_python:
-                self.candidate.requires_python = cast(str, result.metadata["Requires-Python"] or "")
+                # Starting in importlib_metadata 8.0, KeyError is thrown if the key is missing
+                try:
+                    requires_python = result.metadata["Requires-Python"] or ""
+                except KeyError:
+                    requires_python = ""
+                self.candidate.requires_python = requires_python
             self._metadata = result
         return self._metadata
 
--- a/src/pdm/models/candidates.py
+++ b/src/pdm/models/candidates.py
@@ -611,16 +611,11 @@ class PreparedCandidate:
         if self._metadata is None:
             result = self.prepare_metadata()
             if not self.candidate.name:
-                self.req.name = self.candidate.name = cast(str, result.metadata["Name"])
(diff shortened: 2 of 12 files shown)


sydney-runkle:
Thanks for this fix!

I think this might also be related: https://github.com/pydantic/pydantic/actions/runs/9670225358/job/26678546379 (an issue with `Home-Page` from `dist.metadata`). Thoughts?

frostming:
Please fix other occurrences as well, search for (regex) `metadata\["[A-Z]` to find them.

You will also need a compatibility helper function for this to not repeat the try-catch everywhere.

sydney-runkle:
Thanks @JelleZijlstra, quite speedy 😎!

JelleZijlstra:
Thanks, do you know if it is safe to assume that the "Name" key is always present? That's the one PDM looks at most frequently.

src/pdm/cli/commands/list.py
```diff
@@ -309,6 +309,19 @@ def parse_comma_separated_string(
...
 
+def _get_metadata_key(dist: im.Distribution, key: str) -> str | None:
+    """Get a key from metadata, or return None if it is not found or 'UNKNOWN'."""
+    # Starting with importlib-metadata 8.0, KeyError is thrown for missing keys.
+    try:
+        value = dist.metadata[key]
+    except KeyError:
+        return None
+    else:
+        if not value or value == "UNKNOWN":
```
frostming, line 320: I would prefer to just leave `UNKNOWN` as-is? or why not?

JelleZijlstra: All the calling code in this file was replacing UNKNOWN with None, see further down in the diff.

src/pdm/models/candidates.py
```diff
@@ -615,7 +615,12 @@ def metadata(self) -> im.Distribution:
             if not self.candidate.version:
                 self.candidate.version = result.version
             if not self.candidate.requires_python:
-                self.candidate.requires_python = cast(str, result.metadata["Requires-Python"] or "")
+                # Starting in importlib_metadata 8.0, KeyError is thrown if the key is missing
+                try:
+                    requires_python = result.metadata["Requires-Python"] or ""
+                except KeyError:
+                    requires_python = ""
```
frostming, line 622: replace it using the helper function?

src/pdm/cli/commands/list.py
JelleZijlstra: But agree that if we want a more general helper function, it's better to keep the "UNKNOWN" check out. I think I'll add a helper to `utils.py` and use that in the places this PR already changes plus a few more.

src/pdm/models/candidates.py
JelleZijlstra: I realized I can use the `.get()` method since the metadata dictionary is a subclass of `email.message.Message`: https://docs.python.org/3.10/library/email.message.html#email.message.EmailMessage.get.

With that, I don't think the helper function is necessary, so I removed it.

I went over all the places I could find that did an unguarded key access to the metadata mapping and replaced most of them with `.get()`. I found some other code in PDM dealing with cases where the Name field is missing, so I changed even that.

src/pdm/cli/commands/list.py
```diff
@@ -335,17 +322,20 @@ class Listable:
     def __init__(self, dist: im.Distribution, groups: set[str]):
         self.dist = dist
 
-        self.name = _get_metadata_key(dist, "Name")
+        self.name = dist.metadata.get("Name")
```
frostming, line 325: I object to the use of `.get`, which is not exposed in the [public protocol of importlib-metadata](https://github.com/python/importlib_metadata/blob/f3901686abc47853523f3b211873fc2b9e0c5ab5/importlib_metadata/_meta.py#L11) but is only an implementation detail which we shouldn't rely on.

src/pdm/models/candidates.py
frostming: Oops, you will need to fix the CI failures.

JelleZijlstra: Thanks, just fixed the syntax issue on pre-3.12. The 3.12 tests did pass so hopefully the other Python versions will pass too now.

JelleZijlstra:
Not sure what's causing the failure in 3.13 on Windows, but the same failure appeared on #2973, so it's probably not related to my changes.

frostming (approved):

frostming:
Thanks

PR #2977, opened 2024-06-26
refactor: `pdm python install --list` re-use the same implementation …

o-moe:
…as other cli commands that work with Python interpreters from pbs_installer

## Pull Request Checklist

Refactored using the same Python interpreter list across commands.
Made `build_dir=False` more verbose everywhere it is used

### DIFF
--- a/src/pdm/cli/commands/python.py
+++ b/src/pdm/cli/commands/python.py
@@ -11,6 +11,7 @@ from pdm.cli.options import verbose_option
 from pdm.environments import BareEnvironment
 from pdm.exceptions import InstallationError, PdmArgumentError
 from pdm.models.python import PythonInfo
+from pdm.utils import get_all_installable_python_versions
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace, _SubParsersAction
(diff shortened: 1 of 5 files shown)


src/pdm/cli/commands/python.py
```diff
@@ -136,7 +133,7 @@ def install_python(project: Project, request: str) -> PythonInfo:
         version, _, arch = version.partition("-")
         arch = "x86" if arch == "32" else (arch or THIS_ARCH)
 
-        ver, python_file = get_download_link(version, implementation=implementation, arch=arch)
```
o-moe, line 139: made it more obvious that build_dir is used and is False, like with the new util method `get_all_installable_python_versions`

src/pdm/project/core.py
```diff
@@ -779,35 +780,25 @@ def env_or_setting(self, var: str, key: str) -> Any:
...
-        arch = "x86" if THIS_ARCH == "32" else THIS_ARCH
-        matches = [
-            v
-            for v, u in PYTHON_VERSIONS.items()
-            if get_version(v) in self.python_requires
-            and v.implementation.lower() == "cpython"
-            and u.get((THIS_PLATFORM, arch, True))
-        ]
-        return matches
-
```
o-moe, line 799: moved to util without the filtering for best match strategy

frostming (approved):

PR #2978, opened 2024-06-26
feat: add --license and --project-version option to `pdm init`

o-moe:
## Pull Request Checklist

Since license and project's version are both not taken from a given template (pdm's built in template mechanism), I propose to add them as CLI options to control and streamline them during `init` - especially in automated scenarios with `--non-interactive`

I did not adjust the documentation since the documentation does not mention the other options very detailed other than referring to the CLI reference.


frostming:
LGTM

PR #2995, opened 2024-07-04
feat: Lock for specific (python, platform, impl) targets

frostming:

## Pull Request Checklist


frostming:
Hi plugin authors:

In this PR, which I would merge shortly, I've made some changes to the public(I wish I have made some of them private though) APIs. While I am trying hard to make the change compatible, I am not sure if it would break downstream projects.

If you have time, could you use the changes in this PR to verify your plugin functionality, ensure there are no disruptions, and that you can see deprecation warnings?

/cc some of the plugin authors： @macro128 @carstencodes @noirbizarre @sigma67

sigma67:
@frostming I am unable today but I will have a look on Sunday

macro128:
Hi! Sure, I'll test my plugin asap and make the changes needed to work with the new lock mechanism

carstencodes:
I am busy this weekend with some weiqi, but I'll check ASAP. 

My current feeling is, that

- pdm-audit plugin won't break: It uses a post install hook (seems unaffected) and calls pdm export under the hood
- pdm-sbom plugin won't break: It parses the lock file directly and the version does not seem to change
- pdm-bump plugin won't break: It is not interested in the dependency tree.

frostming:
@carstencodes Thanks for the feedback. I've listed some breaking changes in the news and other removal, they are basically related to dependencies manipulation and resolution results.

I am asking to see if there is any I missed and whether we can make a compatibility layer for them.

PR #3001, opened 2024-07-08
use pdm-build-locked

sigma67:
## Pull Request Checklist


sigma67:
> prefer the backend-only approach. is it possible?

done

.github/workflows/release.yml
```diff
@@ -32,7 +32,9 @@ jobs:
 
       - name: Build artifacts
         run: |
-          pipx run build
+          pip install .
+          pdm install --plugins
```
frostming, line 36: we don't need to change this for backend approach

sigma67: undone just now

.github/workflows/test-release.yml
```diff
@@ -0,0 +1,43 @@
+### DO NOT MERGE
```
frostming, line 1: Should this file be removed before merging?

sigma67: Up to you. It could be used to test the release in merge requests. If desired this line should be removed, else the entire file

frostming: Let's remove the entire file

frostming (approved):

PR #3026, opened 2024-07-16
docs: make chatbot always display at the right height and set as topmost

ryomahan:
## Pull Request Checklist

fix chatbot style, make chatbot always display at the right height and set as topmost


frostming:
Looks good, thanks

PR #3056, opened 2024-07-30
feat: replace literal types with type aliases

vvanglro:
## Pull Request Checklist


src/pdm/_types.py
```diff
@@ -84,8 +84,8 @@ def __rich__(self) -> str:
         return "\n".join(lines)
 
 
-RequirementDict = Union[str, Dict[str, Union[str, bool]]]
-CandidateInfo = Tuple[List[str], str, str]
+RequirementDict = Union[str, dict[str, Union[str, bool]]]
```
vvanglro, line 87: python3.8 signature behavior of variables and parameters is inconsistent.
```py
from __future__ import annotations

from typing import Union, Dict, Tuple, List

RequirementDict = Union[str, Dict[str, Union[str, bool]]]
CandidateInfo = Tuple[List[str], str, str]


def a(b: Union[str, dict[str, Union[str, bool]]], c: tuple[list[str], str, str]):
    print(b)
    print(c)


a(b={"a": "b"}, c=(["a", "b"], "c", "d"))

```

vvanglro:
CI unexpected error?

Issue #2993, opened 2024-07-03
dinamically set env variable


ZeeD:
## Is your feature/enhancement proposal related to a problem? Please describe.

I am starting an airflow project.
I successfully used pdm to install airflow, my dependencies, etc...

my goal is to set `AIRFLOW_HOME` to a folder of my project (basically I want a layout like the following one)

```
.../myproject
├── airflow
│   └── dags
...
├── pdm.lock
└── pyproject.toml
```

and also, I am trying to leverage `pdm` scripts support.

basically I want to set

```
[tool.pdm.scripts]
_.env = { AIRFLOW_HOME = ".../myproject/airflow" }
start = { cmd = "airflow standalone" }
```

but airflow expect an absolute path for `AIRFLOW_HOME ` and I don't want to hardcode that in the `pyproject.toml` file

## Describe the solution you'd like

There are many tricks that could help me, but at the end I want to be able to set an environmental variable dinamically, and not with a literal.

* maybe `pdm` could expose `${PROJECT_ROOT}` as an env variable AND let me set `AIRFLOW_HOME = "${PROJECT_ROOT}/airflow"`
* or `pdm` let me invoke a command, so I could write something like `AIRFLOW_HOME = "$({pdm} info --where)/airflow"`

but at the moment the values are treated as literals

o-moe:
Have you seen the [pdm-dotenv](https://github.com/znd4/pdm-dotenv) plugin from the [awesome-pdm list](https://github.com/pdm-project/awesome-pdm)? Maybe you can utilize it to define your `AIRFLOW_HOME` and use it with pdm.

ZeeD:
hi @o-moe , thanks for your suggestion.
I was checking the plugin, but from a cursory look I might have some doubts about its benefits.

Moreso that my understanding is that `pdm`'s `[tool.pdm.scripts]` section already support the load of a shared custom env file, via `_.env_file = ".env"`

In any case your suggestion bring me to do some additional tests, and it seems that I can recycle env variables in the `.env` file. Witch is reasonable, as from my understanding `pdm` is using `python-dotenv` under the hood.

But it seems to me that `PROJECT_ROOT` is not available during the `.env` analysis, so I'm stuck again, because I cannot retrieve the project full path.
And from what I know `python-dotenv` doesn't support any kind of process execution in the `.env` file boy, but maybe I'm wrong.

Do you think there are some paths I've missed?

ZeeD:
Hi, it's me again.

[...]

Issue #1324, opened 2022-08-17
env file could not use PDM_PROJECT_ROOT


zhuoqiang:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

## Steps to reproduce

`.env` like following should put the project root into ptyhonpath variable 

```shell
PYTHONPATH=${PYTHONPATH}:${PDM_PROJECT_ROOT}
```

when we use it in `pyproject.toml`

```toml
[tool.pdm.scripts]
_.env_file.override = ".env"
```

## Actual behavior

the `PDM_PROJECT_ROOT` is empty

## Expected behavior

`PDM_PROJECT_ROOT` should be the absolute project root path when reading `.env` file

## Environment Information

```bash
PDM version:
  2.1.2
Python Interpreter:
  /opt/homebrew/Caskroom/mambaforge/base/envs/lc/bin/python (3.10)
Project Root:
  /project/root
Project Packages:
  None
Inside an active virtualenv /opt/homebrew/Caskroom/mambaforge/base/envs/lc, reuse it.
{
  "implementation_name": "cpython",
  "implementation_version": "3.10.5",
  "os_name": "posix",
  "platform_machine": "arm64",
  "platform_release": "21.6.0",
  "platform_system": "Darwin",
  "platform_version": "Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:22 PDT 2022; root:xnu-8020.140.41~1/RELEASE_ARM64_T6000",
  "python_full_version": "3.10.5",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "darwin"
}
```

frostming:
You need to first state the necessity of this and why you think it should work in this way and not others.

zhuoqiang:
I would like to add the project root into PYTHONPATH when execute command.
in the the following trick works in `.env` 

```shell
SCRIPT=$(readlink -f "$0")
THIS_DIRECTORY=$(dirname "$SCRIPT")
PYTHONPATH=${PYTHONPATH}:${THIS_DIRECTORY}
```

however, it did not work in pdm 2.1, when loading `.env` file, it assigns the literal instead the exapnding value.

anyway, is there any way that I could add the project root to PYTHONPATH automatically when using pdm?

frostming:
Python will add the current working directory to the front of `sys.path`, why would you need that?

zhuoqiang:
because '.' is always refering to current directory as the directory changes. 

[...]

PR #3072, opened 2024-08-02
feat: set project root env

ZeeD:
## Pull Request Checklist

Expose `PROJECT_ROOT` as an env variable during script runs

fixes https://github.com/pdm-project/pdm/issues/2993

### DIFF
--- a/news/2993.feature.md
+++ b/news/2993.feature.md
@@ -1,0 +1,1 @@
+set PROJECT_ROOT env variable during script run
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -145,9 +145,9 @@ class TaskRunner:
             "TaskOptions",
             self.project.scripts.get("_", {}) if self.project.scripts else {},
         )
-        if 'env' not in global_options:
-            global_options['env'] = {}
-        global_options['env']['PROJECT_ROOT'] = str(project.root)
+        if "env" not in global_options:
+            global_options["env"] = {}
+        global_options["env"]["PROJECT_ROOT"] = str(project.root)
         self.global_options = global_options.copy()
         self.recreate_env = False
         self.hooks = hooks
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -147,7 +147,7 @@ class TaskRunner:
         )
         if "env" not in global_options:
             global_options["env"] = {}
-        global_options["env"]["PROJECT_ROOT"] = str(project.root)
+        cast(dict[str, str], global_options["env"])["PROJECT_ROOT"] = str(project.root)
         self.global_options = global_options.copy()
         self.recreate_env = False
         self.hooks = hooks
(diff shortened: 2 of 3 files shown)


frostming:
Why not just inject the env into the `process_env` when running processes?

https://github.com/pdm-project/pdm/blob/61657a4b97b5225b5c7d3373df868ea19e371dd9/src/pdm/cli/commands/run.py#L258

You won't contaminate the global_options in this way.

frostming:
Oh, the project root is already exposed via `PDM_PROJECT_ROOT` env.

https://github.com/pdm-project/pdm/blob/c190e1439cf88cfa06bbf20ee01b1f0d6dfeabbb/src/pdm/environments/base.py#L80

If no objection, i will close this PR.

ZeeD:
> Why not just inject the env into the `process_env` when running processes?
> 
> https://github.com/pdm-project/pdm/blob/61657a4b97b5225b5c7d3373df868ea19e371dd9/src/pdm/cli/commands/run.py#L258
> 
> You won't contaminate the global_options in this way.

Yeah, I may try this approach instead


> Oh, the project root is already exposed via `PDM_PROJECT_ROOT` env.
> 
> https://github.com/pdm-project/pdm/blob/c190e1439cf88cfa06bbf20ee01b1f0d6dfeabbb/src/pdm/environments/base.py#L80
> 
> If no objection, i will close this PR.

I didn't knew about `PDM_PROJECT_ROOT`, I may have missed it in the docs, sorry.

But, while invoking directly a script I see that `PDM_PROJECT_ROOT` is available, it seems to me that in the `.env` variable expansion it doesn't work: let's say that I have

`pyproject.toml`:
```toml
[tool.pdm.scripts]
_.env_file = ".env"
var1 = { shell = "echo $VAR1" }
var2 = { shell = "echo $VAR2" }
var3 = { shell = "echo $PDM_PROJECT_ROOT" }
```

`.env`
```properties
VAR1 = "VAL1"

VAR2 = "VAR1=${VAR1}; USER=${USER}; PDM_PROJECT_ROOT=${PDM_PROJECT_ROOT};"
````

`pdm run` have some unexpected behaviour:


```bash
$ pdm run var1
VAL1
$ pdm run var2
VAR1=VAL1; USER=<my username>; PDM_PROJECT_ROOT=;
$ pdm run var3
<current directory
$ 
```

it seems that the definition of the `PDM_PROJECT_ROOT` env variable occurs "too late" to let `dotenv` enrich the `.env` file parsing.

Do you think it would be possible to move the `PDM_PROJECT_ROOT` before the evaluation of the `_.env` file during the execution of the run?

frostming:
> it seems that the definition of the `PDM_PROJECT_ROOT` env variable occurs "too late" to let `dotenv` enrich the `.env` file parsing.

Not only that, all additional env vars are not inject into `os.environ` to make `dotenv` be aware of.

ZeeD:
searching for `PDM_PROJECT_ROOT` I've found #1324 
Do you think that my use case - as you can see in #2993 - could be solved changing the behavior of `pdm`?

frostming:
This is superseded by #3087 thanks anyway

ZeeD:
thanks to you for the update!

Issue #2895, opened 2024-05-21
In PowerShell `iex (pdm venv activate)` fails


yodaldevoid:
- [x] I have searched the issue tracker and believe that this is not a duplicate.

**Make sure you run commands with `-v` flag before pasting the output.**

## Steps to reproduce

1. Install PDM 2.15.3
2. In PowerShell (any version) clone a project using PDM
```
git clone https://github.com/pdm-project/pdm.git
cd pdm
```
3. Create a venv by using `pdm sync`
4. Try to activate that venv using `iex (pdm venv activate)`

## Actual behavior

The path to the activation script is printed and the venv is not activated.

## Expected behavior

The venv is activated for the current PowerShell instance.

## Environment Information

```bash
> pdm info && pdm info --env
PDM version:
  2.15.3
Python Interpreter:
  $PROJECT_DIR\.venv\Scripts\python.exe (3.10)
Project Root:
  $PROJECT_DIR
Local Packages:

{
  "implementation_name": "cpython",
  "implementation_version": "3.10.5",
  "os_name": "nt",
  "platform_machine": "AMD64",
  "platform_release": "10",
  "platform_system": "Windows",
  "platform_version": "10.0.19044",
  "python_full_version": "3.10.5",
  "platform_python_implementation": "CPython",
  "python_version": "3.10",
  "sys_platform": "win32"
}
```

## Further Information

Before PDM 2.15.3
```
> pdm venv activate
. '$PROJECT_DIR\.venv\Scripts\Activate.ps1'
```

PDM 2.15.3
```
> pdm venv activate
'$PROJECT_DIR\.venv\Scripts\Activate.ps1'
```

I believe the missing period before the script path is causing the failure.

frostming:
see the context in #2854

you may need to discuss with @Pluttodk on how to solve this cleany and make both happy

vix597:
My current workaround

```
iex (iex (pdm venv activate))
```

:upside_down_face:

PR #3080, opened 2024-08-06
Fix activate for windows iex

Pluttodk:
## Pull Request Checklist

Fixed the issue #2895 that caused `iex (pdm venv activate)` to fail. Ensured that it instead returns the output without quotes


frostming (changes requested):
You don't need to update the CHANGLOG.md, but should add a news fragment into `news/` directory.

Pluttodk:
> You don't need to update the CHANGLOG.md, but should add a news fragment into `news/` directory.

It has been updated accordingly now :)

frostming (approved):

PR #3197, opened 2024-10-01
Replace bitwise AND with logical AND (which is breaking my use of UV)

theNewFlesh:
## Pull Request Checklist

`uv.py` uses bitwise and (&) for logic in which logical and (and) is required. When using UV with PDM, PDM throws a TypeError on these lines. This change resolves that.

### DIFF
--- a/src/pdm/formats/uv.py
+++ b/src/pdm/formats/uv.py
@@ -189,11 +189,11 @@ class _UvFileBuilder:
 
     def _make_dependency(self, candidate: Candidate, req: Requirement) -> dict[str, Any] | None:
         locked_repo = self.locked_repository
-        parent_marker = (req.marker or get_marker("")) & (candidate.req.marker or get_marker(""))
+        parent_marker = (req.marker or get_marker("")) and (candidate.req.marker or get_marker(""))
         matching_entries = [e for k, e in locked_repo.packages.items() if k[0] == req.key]
 
         def marker_match(marker: Marker | None) -> bool:
-            return not (parent_marker & (marker or get_marker("")).is_empty())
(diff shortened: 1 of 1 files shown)


frostming (commented):
because that is not a logical and, that is an intersection operation. can you show how would it raise typeerror when both sides are not None?

theNewFlesh:
I am looking for the error in my logs.

It was something like:
`TypeError: unsupported operand type(s) for &: Marker and bool`

I ran some test code to see if bitwise AND gives the same error and it does.

I'll post the error if I can find it.

theNewFlesh:
Well I'm not finding it in the logs, but I found an old search I performed on the exact error:

`unsupported operand type(s) for &: 'Marker' and 'bool'`

On line 192 I believe.

I am switching over some existing projects from vanilla PDM to PDM with UV. Most have worked fine, but one often throws this error when I `pdm sync`.

clement-tourriere:
Hello, I have the exact same error. Sadly I can't really provide an easy reproductible scenario, but from the code I guess there is parenthesis missing:

```
parent_marker & (marker or get_marker("")).is_empty()
```
raises the error: `*** TypeError: unsupported operand type(s) for &: 'Marker' and 'bool'`

but 
```
(parent_marker & (marker or get_marker(""))).is_empty()
```
works well.

I guess the fix could be something like:

```
        def marker_match(marker: Marker | None) -> bool:
            return not ((parent_marker & (marker or get_marker(""))).is_empty())
```

frostming:
> Hello, I have the exact same error. Sadly I can't really provide an easy reproductible scenario, but from the code I guess there is parenthesis missing:

Thank you, you found the real bug! Too many parentheses made me overlook it.

Issue #3270, opened 2024-11-10
Proposal: output similar command when the input command is not correct


Zheaoli:
## Is your feature/enhancement proposal related to a problem? Please describe.

For now, some other package manager will output a similar command when the input command is not correct.

For example

```text
╰─ poetry en

The command "en" does not exist.

Did you mean one of these?
    env use
    env info
    env list
    env remove
```

I think we can add this feature to help people use PDM more friendly

The basic code will be simple, like following below:

```python
distance = SequenceMatcher(None, actual_name, name).ratio()
```

We can calculate the distance by using `SequenceMatcher` in stdlib, and output the ratio >= 0.5

If this proposal has been accepted, I would like to send a PR for this

frostming:
SGTM

PR #3274, opened 2024-11-11
feat: Output similar commands or script command when the input command is not correct

Zheaoli:
Signed-off-by: Manjusaka <me@manjusaka.me>

## Pull Request Checklist

Support #3270

### DIFF
--- a/src/pdm/cli/utils.py
+++ b/src/pdm/cli/utils.py
@@ -157,25 +157,25 @@ def find_similar_text(origin_name: str, target_names: list[str], ratio: float = 
     return [name for name, _ in sorted(results, key=lambda x: x[1], reverse=True)]
 
 
-class ArgumentParserSimilarComandUtil:
-    color: ClassVar[dict[str, str]] = {
-        "red": "\033[91m",
-        "green": "\033[92m",
-        "yellow": "\033[93m",
-    }
-
-    @classmethod
-    def format_similar_command(cls, root_command: str, commands: list[str], script_commands: list[str]) -> str:
-        similar_commands = find_similar_text(root_command, commands)
-        similar_script_commands = find_similar_text(root_command, script_commands)
-        message = f"""{cls.color['red']}Command not found: {root_command}
-    {cls.color['green']}Did you mean one of these command?
-        {cls.color['green']}{similar_commands}
-
-    {cls.color['yellow']}Or one of these script command?
-        {cls.color['yellow']}{similar_script_commands}
-        """
-        return message
+def format_similar_command(root_command: str, commands: list[str], script_commands: list[str]) -> str:
+    similar_commands = find_similar_text(root_command, commands)
+    similar_script_commands = find_similar_text(root_command, script_commands)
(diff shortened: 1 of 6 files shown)


Zheaoli:
image

src/pdm/cli/utils.py
```diff
@@ -156,6 +157,36 @@ def parse_known_args(self, args: Any = None, namespace: Any = None) -> Any:
...
+            results.append((name, ratio))
+    return [name for name, _ in sorted(results, key=lambda x: x[1], reverse=True)]
+
+
+class ArgumentParserSimilarComandUtil:
+    color = {
+        "red": "\033[91m",
+        "green": "\033[92m",
+        "yellow": "\033[93m",
+    }
```
frostming: You can use this utility to get colorized message with rich:
https://github.com/pdm-project/pdm/blob/1fd94bd5044ea10bbc814d92144a2d1a1bd663fd/src/pdm/termui.py#L58-L70

For example

```python
message = style("[success]This text is green[/]")
```

src/pdm/cli/utils.py
```diff
@@ -156,6 +157,36 @@ def parse_known_args(self, args: Any = None, namespace: Any = None) -> Any:
...
+
+    @classmethod
+    def format_similar_command(cls, root_command: str, commands: list[str], script_commands: list[str]) -> str:
+        similar_commands = find_similar_text(root_command, commands)
+        similar_script_commands = find_similar_text(root_command, script_commands)
+        message = f"""{cls.color['red']}Command not found: {root_command}
+    {cls.color['green']}Did you mean one of these command?
+        {cls.color['green']}{similar_commands}
```
frostming, line 173: Should better render the choices as seprate lines, like:

```
Did you mean one of these command?
  - foo
  - bar
```

src/pdm/cli/utils.py
```diff
@@ -156,6 +157,36 @@ def parse_known_args(self, args: Any = None, namespace: Any = None) -> Any:
...
+def find_similar_text(origin_name: str, target_names: list[str], ratio: float = 0.5) -> list[str]:
+    results = []
+    for name in target_names:
+        ratio = SequenceMatcher(None, origin_name, name).ratio()
+        if ratio > 0.4:
+            results.append((name, ratio))
+    return [name for name, _ in sorted(results, key=lambda x: x[1], reverse=True)]
+
+
+class ArgumentParserSimilarComandUtil:
```
frostming, line 160: Just a personal flavor thing, I don't like static-method-only classes. Especially when the color attribute is removed, we can just make `format_similar_command` into a plain function.

frostming (commented):
Thank you very much, it is a good start point and we can investigate how to implement the same for subcommands.

src/pdm/cli/utils.py
Zheaoli: LGTM

src/pdm/cli/utils.py
Zheaoli: nice ,I will dupdate it

frostming (approved):
The UI looks great

PR #3281, opened 2024-11-13
Make `OrderedSet.__contains__` run in O(1)

eltoder:
## Pull Request Checklist

Use a dict to store set elements. Also add `__slots__` for speed.


frostming (approved):
Nice fix

PR #3293, opened 2024-11-15
fix: Fix the bug that pdm plugins are invalid on ubuntu system python(#3289)

### LINKED ISSUE #3289: [bug] ubuntu plugins not valid after installation
### Describe the bug

plugins is invalid after installation on Ubuntu

### To reproduce

In one pdm project, create plugin/my_plugin.py
```python
# plugin/my_plugin.py`
def plugin(core):
    print("test")
```
and plugin/pyproject.toml
```toml
# plugin/pyproject.toml
[project]
name = "my_plugin"
version = "0.0.0"

[project.entry-points.pdm]
test_plug = "my_plugin:plugin"

[build-system]
requires = ["pdm-backend"]
build-backend = "pdm.backend"
```
In pyproject.toml, add
```toml
[tool.pdm]
plugins = [
    "-e file:///${PROJECT_ROOT}/plugin#egg=my-plugin"
]
```
run `pdm install --plugins`, make sure the plugins are installed successfully by checking `.pdm-plugins` which should be
```
.pdm-plugins
└── local
    └── lib
        └── python3.10
            └── dist-packages
                ├── my_plugin-0.0.0.dist-info
                │   ├── direct_url.json
                │   ├── entry_points.txt
                │   ├── METADATA
                │   ├── RECORD
                │   └── WHEEL
                └── my_plugin.pth
```
Then, try to type something starts with `pdm` like `pdm lock`

### Expected Behavior

If the plugin is valid, whatever you input, "test" would be output as long as you are running a `pdm` command

### Environment Information

pdm 2.20.1
ubuntu 22.04

### pdm -v output

_No response_

## Reason Analysis
On windows system, plugins' installation route and loading route would be the same, both `${PROJECT}/.pdm-plugins/Lib/site-packages`, but on ubuntu they are different. 
On ubuntu, as `site.py` said
```
For Debian and derivatives, this sys.path is augmented with directories
for packages distributed within the distribution. Local addons go
into /usr/local/lib/python<version>/dist-packages, Debian addons
install into /usr/lib/python3/dist-packages.
/usr/lib/python<version>/site-packages is not used.
```
So the installation route would be `/${PROJECT}/.pdm-plugins/local/lib/python{ver}/dist-packages`, while the loading route got by 
```python
# pdm/core.py
    def _add_project_plugins_library(self) -> None:
        ...
        purelib = sysconfig.get_path("purelib", scheme, replace_vars)
```
would be `${PROJECT}/.pdm-plugins/lib/python{ver}/site-pacakges`, so pdm would not load any plugins on ubuntu.

In fact, when I manually change the `local/lib/python{ver}/dist-packages` to `lib/python{ver}/site-pacakges`, plugins are enabled.

So, the two routes on ubuntu must somehow be consistent.

- [x] Yes, I would like to submit a PR.

monchin:
## Pull Request Checklist

fix #3289

### DIFF
--- a/news/3289.bugfix.md
+++ b/news/3289.bugfix.md
@@ -1,1 +1,1 @@
-Fix the bug that pdm plugins are invalid after installation on ubuntu.
+Fix the bug that pdm plugins are invalid after installation on ubuntu system python.
--- a/news/3289.bugfix.md
+++ b/news/3289.bugfix.md
@@ -1,0 +1,1 @@
+Fix the bug that pdm plugins are invalid after installation on ubuntu.
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -316,7 +316,12 @@ class Core:
 
         base = str(project.root / ".pdm-plugins")
         replace_vars = {"base": base, "platbase": base}
-        scheme = "nt" if os.name == "nt" else "posix_prefix"
+        if os.name == "nt":
+            scheme = "nt"
+        elif sys.platform == "darwin" and "osx_framework_library" in scheme_names:
+            scheme = "posix_prefix"
+        else:
+            scheme = sysconfig.get_default_scheme()
         purelib = sysconfig.get_path("purelib", scheme, replace_vars)
         scripts = sysconfig.get_path("scripts", scheme, replace_vars)
         site.addsitedir(purelib)
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -316,6 +316,8 @@ class Core:
 
         base = str(project.root / ".pdm-plugins")
         replace_vars = {"base": base, "platbase": base}
+
+        scheme_names = sysconfig.get_scheme_names()
         if os.name == "nt":
             scheme = "nt"
         elif sys.platform == "darwin" and "osx_framework_library" in scheme_names:
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -320,7 +320,7 @@ class Core:
         scheme_names = sysconfig.get_scheme_names()
         if os.name == "nt":
             scheme = "nt"
-        elif sys.platform == "darwin" and "osx_framework_library" in scheme_names:
+        elif (sys.version_info < (3, 10)) or (sys.platform == "darwin" and "osx_framework_library" in scheme_names):
             scheme = "posix_prefix"
         else:
             scheme = sysconfig.get_default_scheme()
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -318,10 +318,11 @@ class Core:
         replace_vars = {"base": base, "platbase": base}
 
         scheme_names = sysconfig.get_scheme_names()
-        if os.name == "nt":
-            scheme = "nt"
-        elif (sys.version_info < (3, 10)) or (sys.platform == "darwin" and "osx_framework_library" in scheme_names):
+        if (sys.platform == "darwin" and "osx_framework_library" in scheme_names) or sys.platform == "linux":
             scheme = "posix_prefix"
+        # sysconfig._get_default_scheme is a private function in 3.8 & 3.9
+        elif sys.version_info < (3, 10):
+            scheme = "nt" if os.name == "nt" else "posix_prefix"
         else:
             scheme = sysconfig.get_default_scheme()
         purelib = sysconfig.get_path("purelib", scheme, replace_vars)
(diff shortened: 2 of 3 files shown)


monchin:
fix #3289

monchin:
I need to know why the original code is 
```python
if sys.platform == "darwin" and "osx_framework_library" in scheme_names:
```
Are there macos computers without "osx_framework_library" in scheme_names? Is my code correct on all mainstream systems?

o-moe:
> Are there macos computers without "osx_framework_library" in scheme_names? Is my code correct on all mainstream systems?

Yes, the built-in Python that comes with recent macOS, report

`'nt', 'nt_user', 'osx_framework_user', 'posix_home', 'posix_prefix', 'posix_user'`

Python from brew also reports `osx_framework_library`

If Python reports `osx_framework_library` on macOS is depending on how it is built. This will dictate where it is installed to and what osx functionalities it can access.

o-moe:
If you want to be 100% backwards compatible regarding `darwin` and just add linux to the mix, this could be an alternative:

```python
if ((sys.platform == "darwin" and "osx_framework_library" in scheme_names) or sys.platform == "linux") and kind == "prefix":
```

Basically that's similar to one of the statements from your issue

monchin:
> Basically that's similar to one of the statements from your issue

That's the very first code I thought which should be. But consider [here](https://github.com/pdm-project/pdm/blob/2.20.1/src/pdm/core.py#L314)
```python
# tag 2.20.1 src/pdm/core.py #L314
    def _add_project_plugins_library(self) -> None:
        ...
        scheme = "nt" if os.name == "nt" else "posix_prefix"
        purelib = sysconfig.get_path("purelib", scheme, replace_vars)
        scripts = sysconfig.get_path("scripts", scheme, replace_vars)
```
If `os.name != nt`, then all scheme would be "posix_prefix". That's the reason why this bug occurs on ubuntu. I'm trying to ensure the consistency here and the installation route.

Or maybe there is another way better, 
```python
    def _add_project_plugins_library(self) -> None:
        ...
        if os.name == "nt":
            scheme = "nt"
        elif sys.platform == "darwin" and "osx_framework_library" in scheme_names:
            scheme = "posix_prefix"
        else:
            scheme = sysconfig.get_default_scheme()
        purelib = sysconfig.get_path("purelib", scheme, replace_vars)
        scripts = sysconfig.get_path("scripts", scheme, replace_vars)
```
That would be exactly 100% backwards compatible, and could ensure the consistency of loading path and installation path

o-moe:
I don't think that this difference is a bug. Even though Ubuntu's (Debian's) Python reports as cPython, Debian patches it to better fit their distribution. There's a whole repository to track such changes: https://salsa.debian.org/python-team

Here's a thread regading changing the local installation path starting with Python 3.10 (which is part of the sysconfig patch): https://lists.debian.org/debian-python/2022/03/msg00039.html

frostming:
That's why using system python is a pain. So I can conclude that the issue occurs because packages are installed with `posix_local` scheme, but looked for with `posix_prefix` scheme, right. Is there anyway to use `posix_prefix` on both sides?

monchin:
> Is there anyway to use `posix_prefix` on both sides?

Consider
```python
# pdm/models/in_process/sysconfig_get_paths.py
def get_paths(kind="default", vars=None):
     ...
        if sys.platform == "darwin" and "osx_framework_library" in scheme_names and kind == "prefix":
            return sysconfig.get_paths("posix_prefix", vars=vars)
```
and 
```python
# pdm/core.py
    def _add_project_plugins_library(self) -> None:
        ...
        scheme = "nt" if os.name == "nt" else "posix_prefix"
```
when loading plugins, pdm uses "posix_prefix" for all situations except for windows; but when installing plugins, pdm uses "posix_prefix" only for macos with "osx_framework_library". If we use the first PR I made, then the 2 sides are consistent, but maybe it's not 100% backward compatible; if we use the code @o-moe wrote, I'm afraid that the loading side is still not consistent with the installing side as it does not show special handling for macos with "osx_framework_library".

If we need "posix_prefix" on both sides, how about this
```python
# pdm/models/in_process/sysconfig_get_paths.py
def get_paths(kind="default", vars=None):
     ...
        # as @o-moe  wrote
        if ((sys.platform == "darwin" and "osx_framework_library" in scheme_names) or sys.platform == "linux") and kind == "prefix":
            return sysconfig.get_paths("posix_prefix", vars=vars)
```
```python
# pdm/core.py
    def _add_project_plugins_library(self) -> None:
        ...
        # consisetnt with the installing side
        if ((sys.platform == "darwin" and "osx_framework_library" in scheme_names) or sys.platform == "linux"):
            scheme = "posix_prefix"
        elif sys.version_info < (3, 10):
            scheme = "nt" if os.name == "nt" else "posix_prefix"
        else:
            scheme = sysconfig.get_default_scheme()
```

frostming:
> If we need "posix_prefix" on both sides, how about this

This makes more sense to me. Installing and loading plugins are a prefix scenario

frostming (approved):

PR #3315, opened 2024-11-28
feat(run): allow customizing scripts display from settings (fix #3313)

### LINKED ISSUE #3313: Task verbosity change introduce in 2.21 breaks some scripts
### Describe the bug

In 2.21, in https://github.com/pdm-project/pdm/pull/3278 more precisely, default scripts verbosity has been changed, and it breaks some existing scripts relying on output:
- all scripts that are output-dependants now needs to run with `-q` (can represent some volume to migrate)
- script called using `pdm <command>` have to migrate to `pdm run -q <command>` to preserve output, `-q` not being available on the shortcut form
- script full command is outputted in addition to the task title, resulting in lots of line lost and broken terminal output when relying on some multiline scripts
- it makes the `-v` flag do nothing more
- It kind of break verbosity consistency as everywhere else command echo/feedback is only available with detailed verbosity (ex: https://github.com/pdm-project/pdm/blob/main/src/pdm/cli/commands/venv/backends.py#L72-L75)

### To reproduce

- create a script outputting some formatted data (JSON, CSV, ...) > not parsable anymore without `-q`
- create a script spanning on multiple lines > same amount of lines is prepended in addition to the title
- create a composite script, with optionally pre or post hooks > output is now having 2 extra lines by task hook (assuming they are single line scripts)

### Expected Behavior

One of:
1. default script behavior just run the script without decoration, echoing script name and command line being by definition verbosity
2. this behavior is opt-in, aka. no output as default, a user setting allows enabling this when not verbose, but remain whatever the setting value in verbose mode
3. this behavior is opt-in and progressive, aka. no output as default, a user setting allows disabling the task name only in normal, and the full output remain in verbose mode whatever the setting value

(The third one is having my preference)

### Environment Information

PDM version:
  2.21.0
Python Interpreter:
  <redacted workdir>/.venv/bin/python (3.11)
Project Root:
  <redacted workdir>
Local Packages:

{
  "implementation_version": "3.12.7",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.11.9-arch1-1",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT_DYNAMIC Sun, 17 Nov 2024 16:06:17 +0000",
  "python_full_version": "3.11.9",
  "python_version": "3.11",
  "sys_platform": "linux"
}

### Additional Context

Given I also understand some people like having tasks feedback by default, this PR is to start discussing what should be the solution, choose a setting name (`scripts.echo` ? `scripts.display_name` ? `scripts.always_show_name` ?) if a setting-based solution is accepted.

- [X] Yes, I would like to submit a PR.

noirbizarre:
## Pull Request Checklist

This PR is a proposal implementation for #3313 introduces a `scripts.show_header` setting allowing to customize scripts display on run.
If set to `True`, running a script will display the help before running, fallback to the command args when no help has been set.

This setting do not have impact on verbose mode, which is still displaying script name and resolved args (and not help as help can be pretty but args are useful, and `verbose` mode is often meant for debug/diagnostic).

## Sample

With the following `tool.pdm.scripts` section:

```toml
[tool.pdm.script]
hello = {cmd = "echo 'hello'", help = "Say hello"}
```

### Default verbosity, default settings

image

### Default verbosity, `scripts.show_header = true`

image

#### Without `help` defined

image

### Verbose

image

## Composite sample

With the following `tool.pdm.scripts` section:

```toml
[tool.pdm.script]
hello = {cmd = "echo 'hello'", help = "Say hello"}
composite = { composite = ["hello", "hello"], help = "Do many things" }
```

### Default verbosity, default settings

image

### Default verbosity, `scripts.show_header = true`

image

#### Without `help` defined

image

### Verbose

image

### DIFF
--- a/src/pdm/cli/commands/run.py
+++ b/src/pdm/cli/commands/run.py
@@ -341,11 +341,8 @@ class TaskRunner:
         elif kind == "composite":
             assert isinstance(value, list)
 
-        self.project.core.ui.echo(
-            f"Running {task}: [success]{args}[/]\n",
-            err=True,
-            verbosity=termui.Verbosity.NORMAL,
-        )
+        self.display_task(task, args)
+
         if kind == "composite":
             args = list(args)
             should_interpolate = any(RE_ARGS_PLACEHOLDER.search(script) for script in value)
@@ -365,6 +362,19 @@ class TaskRunner:
                     composite_code = code
             return composite_code
         return self._run_process(args, chdir=True, shell=shell, **merge_options(self.global_options, options, opts))  # type: ignore[misc]
+
+    def display_task(self, task: Task, args: Sequence[str]) -> None:
+        """Display a task given current verbosity and settings"""
+        is_verbose = self.project.core.ui.verbosity >= termui.Verbosity.DETAIL
+        if not (is_verbose or self.project.config.get("scripts.show_header")):
+            return
+        args = task.args if task.kind == "composite" else args
+        content = args if is_verbose else task.short_description
+        self.project.core.ui.echo(
+            f"Running {task}: [success]{content}[/]",
+            err=True,
+            verbosity=termui.Verbosity.NORMAL,
+        )
 
     def run(
         self,
--- a/news/3313.feature.md
+++ b/news/3313.feature.md
@@ -1,0 +1,1 @@
+Allow customizing scripts display with `scripts.show_header` settings.
--- a/src/pdm/project/config.py
+++ b/src/pdm/project/config.py
@@ -217,6 +217,12 @@ class Config(MutableMapping[str, str]):
             env_var="PDM_PYPI_JSON_API",
             coerce=ensure_boolean,
         ),
+        "scripts.show_header": ConfigItem(
+            "Display script name and help before running",
+            default=False,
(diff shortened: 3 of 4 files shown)


frostming:
Adding two configurations for this feature seems a bit finicky to me. People may be overwhelmed by this large number of configs

src/pdm/cli/commands/run.py
```diff
@@ -341,11 +341,8 @@ def run_task(
         elif kind == "composite":
             assert isinstance(value, list)
 
-        self.project.core.ui.echo(
-            f"Running {task}: [success]{args}[/]\n",
-            err=True,
-            verbosity=termui.Verbosity.NORMAL,
-        )
+        self.display_task(task, args)
```
frostming, line 344: There is one minor thing, when task type is `composite`, the `args` list is just empty, can you change it to show something meaningful like sub commands?

noirbizarre: PR updated with additional test cases for composite 👍🏼 
I reused the `Task.short_description` property already used in `pdm run -l` so it uses the same rules and formatting.

frostming (approved):
Thanks for that

PR #3325, opened 2024-12-02
fix: uninstallation error when pdm is not installed before

Georgehu716:
## Pull Request Checklist


frostming:
Please also update the .sha256 content

PR #3326, opened 2024-12-02
feat: Add resolution.overrides support for use_uv

clement-tourriere:
## Pull Request Checklist

Hello this is only a POC just to be sure what I'm doing is not wrong. It seems to work fine on my environment, but I'm not sure about it. @frostming if you think it works, I will be more than happy to add fragment in `news` and also `tests`, if not, don't hesitate to tell me so :)


frostming:
LGTM. Can you add a news entry?

frostming (approved):

PR #3358, opened 2024-12-25
feat: Add new command option to set env file for pdm script 

### LINKED ISSUE #3355: Allow inject enviroment variables via .env or other env file
### Feature Description

For now, we may run some code via PDM script like

```text
run_server = { shell = "uvicorn --host 0.0.0.0 --port $SERVER_PORT exporter.app:app --reload"}
```

Or I may some other environment variables in my program. I may want to write the environment variables to a file like .env when I deploy my program.

### Problem and Solution

I propose add new global flag `--env-file`, the default value is `.env`. The PDM main process will inject the environment variables into subprocess if the file is exist

### Additional Context

_No response_

- [X] Yes, I am willing to contribute to the development of this feature.

Zheaoli:
## Pull Request Checklist

Fix #3355

### DIFF
--- a/docs/usage/scripts.md
+++ b/docs/usage/scripts.md
@@ -232,7 +232,9 @@ Note how we use [TOML's syntax](https://github.com/toml-lang/toml) to define a c
 !!! note
     Environment variables specified on a composite task level will override those defined by called tasks.
 
-### `env_file`
+### Setup env vars from a file
+
+Note: **The env_file option will be deprecated in the future.**
 
 You can also store all environment variables in a dotenv file and let PDM read it:
 
@@ -251,12 +253,19 @@ start.cmd = "flask run -p 54321"
 start.env_file.override = ".env"
 ```
 
+After v2.23.0, we introduced a new way to specify the env file:
+
+1. We will default load the env file in the project under the name convention `.env.${environment}`. The `${environment}` is the value of the `PDM_ENVIRONMENT` environment variable(or You can set it by using `-e/--environment` command option). If the variable is not set, it will default to `local`.
+2. You can also specify the env file by using the `--env-file` command option.
+3. We will not override the other env vars loaded from the sources loaded before.
+
 !!! note "Environment variable loading order"
     Env vars loaded from different sources are loaded in the following order:
 
     1. OS environment variables
     2. Project environments such as `PDM_PROJECT_ROOT`, `PATH`, `VIRTUAL_ENV`, etc
     3. Dotenv file specified by `env_file`
+    4. Env file specified by `--env-file` CLI option or `.env.${environment}` file
     4. Env vars mapping specified by `env`
 
     Env vars from the latter sources will override those from the former sources.
--- a/news/3355.feat.md
+++ b/news/3355.feat.md
@@ -1,0 +1,1 @@
+Add new command option to set env file for pdm script.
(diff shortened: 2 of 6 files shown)


frostming (commented):
I would like to suggest some improvements

The env files can be **cascaded** in a way that more specific env file can override less specific ones. For example, if `PDM_ENVIRONMENT=prod`, we look for the following files and load them if exists:

1. `.env`  (effective for all environments)
3. `.env.prod`  (effective for `prod` only)
4. `.env.local`  (effective for all environments)
5. `.env.prod.local`  (effective for `prod` only)

The variables are **merged** in the above order, with the latter taking precedence.

The improvement is to make `.local` a special modifier rather than a regular environment name. It contains the env vars for the current machine only, and `.env*.local` should be added to the `.gitignore`, while other `.env` files can be shared safely.

Also happy to hear @noirbizarre 's thoughts since he is the one involved in PDM scripts the most.

Zheaoli:
> 1. `.env`  (effective for all environments)
> 2. `.env.prod`  (effective for `prod` only)
> 3. `.env.local`  (effective for all environments)
> 4. `.env.prod.local`  (effective for `prod` only)
> 
> The variables are **merged** in the above order, with the latter taking precedence.
> 
> The improvement is to make `.local` a special modifier rather than a regular environment name. It contains the env vars for the current machine only, and `.env*.local` should be added to the `.gitignore`, while other `.env` files can be shared safely.

Nice idea! I will update it ASAP

noirbizarre:
Sorry, I thought I answered this long ago, but it wasn't the case.

So here's my position:

I see value in the `environment` concept, but I consider that removing the `env_file` option from scripts is not a good idea (in my case I have dozen of corporate projects relying on this, with different values per script, the environment approach would not fit/scale those use cases). I also consider having an automatic env file loading a source of problems (you have no idea how much time I spent debugging just because an env file was silently loaded by a tool, and we were not aware of). So I would be against in the current form.

However, I have a counterproposal:

I see 2 main use cases for dotenv files:
- user dotenv (the classical `.env`): not committed, used by devs to export their personal settings, credentials... Standard and often shared by multiple apps (like docker-compose) by default.
- project wide dotenvs, committed, tests and tooling may rely on it, this is most likely the kind of dotenv we will have in `env_file` in the scripts.

I also like the idea of cascading env files.

So, I would:
- support a opt-in user dotenv autoloading, with a [setting](https://pdm-project.org/en/latest/reference/configuration/), either a boolean to force the standard `.env`, either a configurable string (but it is important for people having their own dotenv loader like `direnv` or `mise` to avoid forcing them into a double loading)
- support the environment concept with the `PDM_ENVIRONMENT` env var and the `-e/--environment` with default on `local` like done in this PR (think that devs may want to set `PDM_ENVIRONMENT` in their `.env`, it should be loaded first if the option is enabled)
- make the `env_file` support a `{environment}` placeholder so it is possible to write `task.env_file = ".env.{environment}"` (or `_.env_file` to make it global) while keeping backward compatibility
- make `env_file` also support a list of `env_file` so it is possible to have cascading

Then this should work:
```toml
_.env_file = [
  ".env.{environment}",
  ".env.{environment}.local",
]
```
But also this:
```toml
my-task.env_file = [
  "dotenvs/my-task",
  "dotenvs/my-task.{environment}",
  "dotenvs/my-task.{environment}.local",
]
```

Bonus: the environment could also be exposed as [script placeholder](https://pdm-project.org/en/latest/usage/scripts/#arguments-placeholder)

What do you think ?

PR #3366, opened 2024-12-31
Also use csh venv activation for tcsh

vedranmiletic:
## Pull Request Checklist

`shellingham.detect_shell()` returns `('tcsh', '/bin/tcsh')` for tcsh on FreeBSD, so the current code tries to use the Bash venv activation script and fails due to syntax error. This change fixes the issue.


frostming (approved):

vedranmiletic:
Thanks!

PR #3415, opened 2025-03-08
fix: pass dev selection to sync on pdm add and remove

huxuan:
## Pull Request Checklist

A trial fix to #3404. 

Actually, I am not sure whether this is the desired approach to fix. Any comments are welcome.

### DIFF
--- a/src/pdm/cli/commands/add.py
+++ b/src/pdm/cli/commands/add.py
@@ -174,7 +174,7 @@ class Command(BaseCommand):
         if sync:
             do_sync(
                 project,
-                selection=GroupSelection(project, groups=[group], default=False),
+                selection=GroupSelection(project, groups=[group], dev=selection.dev, default=False),
                 no_editable=no_editable and tracked_names,
                 no_self=no_self or group != "default",
                 requirements=group_deps,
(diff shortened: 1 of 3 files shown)


frostming:
LGTM, pdm add seems also suffer from [the same issue](https://github.com/pdm-project/pdm/blob/21375cd65753f03f4d8ba18d7ddde88f3e77873b/src/pdm/cli/commands/add.py#L177), can you change it too?

frostming (approved):

src/pdm/cli/commands/add.py
```diff
@@ -174,7 +174,7 @@ def do_add(
         if sync:
             do_sync(
                 project,
-                selection=GroupSelection(project, groups=[group], default=False),
+                selection=GroupSelection(project, groups=[group], dev=selection.dev, default=False),
```
frostming, line 177: Oops, that fails a test case, when the group is a dev group but `--dev` is not passed explicitly. I guess it's not the correct way to fix, neither for `pdm remove`.

I think you should fix the `_translated_groups()` function, so that when dev groups don't exist in the lock file they won't be added to the group_set

huxuan: Gotcha, I will take a look later.

frostming: You can use a set intersection to exclude the non-exist groups

Issue #3441, opened 2025-04-04
Support more trusted publishing providers


lioman:
### Feature Description

Currently, only GitHub is supported by `pdm`, but `pypi.org` supports others as well.

### Problem and Solution

If you use `pdm` as a tool and host your code on GitLab, you can't use trusted providers package publishing.
Solution would be to use `id` as `twine` does it, so that `pdm` don't need to implement every supported provider on its own.

### Are you willing to contribute to the development of this feature?

- [ ] Yes, I am willing to contribute to the development of this feature.

PR #3442, opened 2025-04-04
Use id to invoke oidc token generation

lioman:
`id` implements detection of different providers, like Github, Gitlab etc. With that pdm does not need to take care of all supported providers and can stand of the shoulder of that library that twine uses as well.
- removed pdm's implementation of generating a token
- add `id` dependency and use it
- add test dependency to mock httpx requests
- add test for oidc auth of Repository

## Pull Request Checklist

Closes: #3441

### DIFF
--- a/src/pdm/cli/commands/publish/repository.py
+++ b/src/pdm/cli/commands/publish/repository.py
@@ -5,7 +5,13 @@ from typing import TYPE_CHECKING, Any, Iterable, cast
 from urllib.parse import urlparse, urlunparse
 
 import httpx
-from rich.progress import BarColumn, DownloadColumn, TimeRemainingColumn, TransferSpeedColumn
+from id import AmbientCredentialError, detect_credential
+from rich.progress import (
+    BarColumn,
+    DownloadColumn,
+    TimeRemainingColumn,
+    TransferSpeedColumn,
+)
 
 from pdm import termui
 from pdm.cli.commands.publish.package import PackageFile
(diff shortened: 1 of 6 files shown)


frostming:
That's cool. Thanks

frostming (approved):

lioman:
@frostming please wait a bit. I will increase test  coverage and want to test it with gitlab, so that I can add some example to the configuration.

lioman:
I tested it with gitlab: https://gitlab.com/lioman/nextcloud_news_filter/-/jobs/9627357043 and it works.

src/pdm/cli/commands/publish/repository.py
```diff
@@ -71,33 +77,30 @@ def _ensure_credentials(self, username: str | None, password: str | None) -> tup
...
+                    "This platform is not supported for trusted publishing via OIDC",
+                    err=True,
+                )
+                return None
+            mint_token_url = urlunparse(parsed_url._replace(path="/_/oidc/mint-token"))
             resp = self.session.post(mint_token_url, json={"token": oidc_token})
             resp.raise_for_status()
             token = resp.json()["token"]
+        except AmbientCredentialError as e:
+            self.ui.echo(f"Unable to detect OICD token for CI platform: {e}", err=True)
```
frostming, line 100:
```suggestion
            self.ui.echo(f"Unable to detect OIDC token for CI platform: {e}", err=True)
```

tests/cli/test_publish.py
```diff
@@ -191,3 +215,74 @@ def test_repository_get_credentials_from_keyring(project, keyring, mocker):
...
+    test_pypi_url = "https://test.org/upload"
+
+    config = RepositoryConfig(config_prefix="repository", name="test", url=test_pypi_url)
+    # set env variable to get `id`` into detect_github function
+    monkeypatch.setenv("GITHUB_ACTIONS", "true")
+
+    with pytest.raises(PdmUsageError):
+        Repository(project, config=config)
+    captured = capsys.readouterr()
+    assert "Unable to detect OICD token for CI platform:" in captured.err
```
frostming, line 288:
```suggestion
    assert "Unable to detect OIDC token for CI platform:" in captured.err
```

lioman: Sorry - I screwed up. Fixed it now.

frostming (approved):

PR #3459, opened 2025-04-10
fix: pdm config -e opens in read-only

### LINKED ISSUE #3423: opening new config file with pdm config -e opens in read-only
### Discussed in https://github.com/pdm-project/pdm/discussions/3422

<div type='discussions-op-text'>

<sup>Originally posted by **Zyantist** March 12, 2025</sup>
Hello, 

I found this interesting library yesterday and am stuck at the configuration. I want to use it in a work environment where we have a PYPI-like server in our local network and with pip, I use `trusted-host`, `extra-index-url`s with personal access tokens.

That is why I was looking for a config file where I could add the (adapted) information for pdm somewhere in my HOME directory where different containers with access to my HOME can read it. 

After I read some bits in the docs and finally used `pdm config -h` to find the `pdm config -e` command, I hoped to finally be able to write my configuration items into a file in my HOME. I tried with nano and vim, but in both I enter a line, try to save and it tells me that it is opened in read-only mode. The file and folder (`~/.config/pdm/`) did not exist before-hand and still do not. 
I would expect the command to take care of creating folder and file and open it a writable mode since the help says

>  -e, --edit            Edit the configuration file in the default editor(defined by EDITOR env var)

I know I could of course create the folder and file manually and fill them but I was wondering if this is a bug or if I missed some additional options? 

Working inside a virtual environment (venv) with Python3.11 in a toolbox container with Fedora 41 workstation and gnome 47. The host system has the same OS/gnome.</div>

shaonianche:
fix #3423

## Pull Request Checklist


frostming (commented):
LGTM, please add a news fragment under `news/` folder

PR #3466, opened 2025-04-13
doc: autocorrect using autocorrect

yihong0618:
## Pull Request Checklist

using autocorrect to autocorrect 
FYI: https://github.com/huacnlee/autocorrect

### DIFF
--- a/README_zh.md
+++ b/README_zh.md
@@ -20,7 +20,7 @@
 
 </div>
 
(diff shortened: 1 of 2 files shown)


Kilerd:
probably adding the check on CI is a better solution, so that all incoming commits and PR can be check before merging.

frostming (approved):

PR #3472, opened 2025-04-15
fix: faster pytest local same as ci info to dev doc

yihong0618:
## Pull Request Checklist

as title

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,16 +16,20 @@ this section.
 
 1. First, fork this project to your own namespace using the fork button at the top right of the repository page.
 2. Clone the **upstream** repository to local:
+
    ```bash
    git clone https://github.com/pdm-project/pdm.git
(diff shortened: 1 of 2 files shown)


frostming (commented):
no, I would like to give more flexibility to local dev. They can run `pdm test -n auto` if they like and free to strip the option when one would like to debug a single case

yihong0618:
> pdm test -n auto

cool will that will better, will add it to dev doc

CONTRIBUTING.md
```diff
@@ -83,13 +87,19 @@ Now, all dependencies are installed into the Python environment you chose, which
 pdm run test
 ```
 
+faster test using pytest-xdist
+
+```bash
+pdm run test -n local
```
frostming, line 93: `local` is an invalid value, do you mean `auto`?

yihong0618: yes...

Issue #3487, opened 2025-04-19
Test failures in network-isolated environment (not marked as `network`)


mgorny:
### Describe the bug

When running the test suite in network-isolated environment, I'm seeing a few more tests fail, even though they are not marked as `network`:

```pytb
FAILED tests/cli/test_build.py::test_build_with_no_isolation - AssertionError: assert 0 == 1
 +  where 0 = RunResult(exit_code=0, stdout='Building sdist...\nBuilt sdist at /tmp/pytest-of-mgorny/pytest-41/popen-gw3/test_build_...test-41/popen-gw3/test_build_with_no_isolation0/dist/test_project-0.0.0-py3-none-any.whl\n', stderr='', exception=None).exit_code
FAILED tests/models/test_candidates.py::test_find_candidates_from_find_links - httpx.ConnectError: [Errno -2] Name or service not known
FAILED tests/test_project.py::test_find_interpreters_with_PDM_IGNORE_ACTIVE_VENV[True] - RuntimeError: Call command ['install'] failed(1): WARNING: Lockfile does not exist
Updating the lock file...
See /tmp/pytest-of-mgorny/pytest-41/popen-gw7/test_find_interpreters_with_PD0/logs/pdm-install-zu4iu2jr.log for detailed debug log.
[ConnectError]: [Errno -2] Name or service not known
```

### To reproduce

```console
$ git rev-parse HEAD
ba2af395c27cd378b326a7ce22244d755e11e580
$ pdm venv create
Virtualenv /tmp/pdm/.venv is created successfully
$ pdm install -dGtest
INFO: Virtualenv /tmp/pdm/.venv is reused.
Synchronizing working set with resolved packages: 50 to add, 0 to update, 0 to remove

  ✔ Install anyio 4.3.0 successful
  ✔ Install blinker 1.8.2 successful
  ✔ Install id 1.5.0 successful
  ✔ Install certifi 2024.8.30 successful
  ✔ Install filelock 3.16.1 successful
  ✔ Install colorama 0.4.6 successful
  ✔ Install idna 3.6 successful
  ✔ Install execnet 2.1.1 successful
  ✔ Install findpython 0.6.3 successful
  ✔ Install dep-logic 0.5.0 successful
  ✔ Install h11 0.14.0 successful
  ✔ Install httpx 0.27.2 successful
  ⠋ Installing charset-normalizer 3.3.2...                    
  ✔ Install distlib 0.3.8 successful
[...]

PR #3488, opened 2025-04-19
fix: Mark two more tests as requiring network (#3487)

mgorny:
## Pull Request Checklist

Mark two more fails with the `network` marker, as they fail without Internet access, as noted in #3487.

### DIFF
--- a/tests/models/test_candidates.py
+++ b/tests/models/test_candidates.py
@@ -283,10 +283,9 @@ def test_ignore_invalid_py_version(project):
     _ = next(iter(repo.find_candidates(req)))
 
 
-@pytest.mark.network
 def test_find_candidates_from_find_links(project):
     project.pyproject.settings["source"] = [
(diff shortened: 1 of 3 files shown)


tests/models/test_candidates.py
```diff
@@ -283,6 +283,7 @@ def test_ignore_invalid_py_version(project):
     _ = next(iter(repo.find_candidates(req)))
 
 
+@pytest.mark.network
 def test_find_candidates_from_find_links(project):
     project.pyproject.settings["source"] = [
         {"name": "test", "url": "http://fixtures.test/index/demo.html", "verify_ssl": False, "type": "find_links"}
```
frostming, line 289:
```suggestion
        {"name": "pypi", "url": "http://fixtures.test/index/demo.html", "verify_ssl": False, "type": "find_links"}
```

Can you verify this change fixes the network problem?

mgorny: Thanks, confirmed it does and updated the pull request.

frostming (approved):

mgorny:
Thanks!

Issue #3485, opened 2025-04-18
Install error on Windows 11 Chinese Version


JamCh01:
### Describe the bug

```
PS C:\Users\user\Desktop> powershell -ExecutionPolicy ByPass -c "irm https://pdm-project.org/install-pdm.py | py -"
Installing PDM (latest): Creating virtual environment
Installing PDM (latest): Installing PDM and dependencies
An error occurred when executing ['C:\\Users\\user\\AppData\\Roaming\\pdm\\venv\\Scripts\\python.exe', '-Im', 'pip', 'install', 'pdm[locked]']:
Traceback (most recent call last):
  File "<stdin>", line 41, in _call_subprocess
  File "D:\Program Files\Python313\Lib\subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['C:\\Users\\user\\AppData\\Roaming\\pdm\\venv\\Scripts\\python.exe', '-Im', 'pip', 'install', 'pdm[locked]']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 460, in <module>
  File "<stdin>", line 456, in main
  File "<stdin>", line 354, in install
  File "<stdin>", line 281, in _install
  File "<stdin>", line 44, in _call_subprocess
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 12462: invalid start byte
```

Python Version for venv
```
PS C:\Users\user\Desktop> C:\\Users\\user\\AppData\\Roaming\\pdm\\venv\\Scripts\\python.exe -V
Python 3.13.1
```

Python Version for installed
```
PS C:\Users\user\Desktop> python -V
Python 3.13.1
```

try to run subprocess
```
PS C:\Users\user\Desktop> C:\\Users\\user\\AppData\\Roaming\\pdm\\venv\\Scripts\\python.exe -Im pip install pdm[locked]
Collecting pdm[locked]
  Using cached pdm-2.24.0-py3-none-any.whl.metadata (15 kB)
Collecting blinker (from pdm[locked])
  Using cached blinker-1.9.0-py3-none-any.whl.metadata (1.6 kB)
Collecting packaging!=22.0,>=20.9 (from pdm[locked])
  Using cached packaging-24.2-py3-none-any.whl.metadata (3.2 kB)
[...]

PR #3511, opened 2025-05-12
Fix pyproject toml gbk

waketzheng:
## Pull Request Checklist

Fixes https://github.com/pdm-project/pdm/issues/3485#issuecomment-2835248012

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ ci:
   autoupdate_schedule: monthly
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.11.7'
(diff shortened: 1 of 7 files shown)


frostming:
This fix is not correct. [Any encoding other than UTF-8 should be considered invalid TOML](https://arc.net/l/quote/dnflblcs)

That's why tomllib accepts a binary data, instead of UTF-8 string. As it will validate behind the scene.

waketzheng:
> This fix is not correct. [Any encoding other than UTF-8 should be considered invalid TOML](https://arc.net/l/quote/dnflblcs)
> 
> That's why tomllib accepts a binary data, instead of UTF-8 string. As it will validate behind the scene.

OK, get it. Thanks for the quickly reply.

PR #3524, opened 2025-05-28
fix: Change the return type of `array_of_inline_tables` to list[dict]…

### LINKED ISSUE #3523: [bug]The type hint  of `array_of_inline_tables` seems to be wrong
### Describe the bug

Currently the definition of `array_of_inline_tables` is
```python
def make_inline_table(data: Mapping) -> dict:
    """Create an inline table from the given data."""
    table = cast(dict, tomlkit.inline_table())
    table.update(data)
    return table


def make_array(data: list, multiline: bool = False) -> list:
    if not data:
        return []
    array = cast(list, tomlkit.array().multiline(multiline))
    array.extend(data)
    return array


def array_of_inline_tables(value: list[Mapping], multiline: bool = True) -> list[str]:
    return make_array([make_inline_table(item) for item in value], multiline)
```
From the source code, it seems that the return value of `array_of_inline_tables` should be `list[dict]` as the `data` in `make_array` should be `list[dict]`

### To reproduce

No reproduce

### Expected Behavior

mypy or pyright should infer to `list[dict]` as the return value type

### Environment Information

PDM version:
  2.24.2
```json
{
  "implementation_version": "3.10.11",        
  "os_name": "nt",
  "platform_machine": "AMD64",
  "platform_release": "10",
  "platform_system": "Windows",
  "platform_version": "10.0.19045",
  "python_full_version": "3.10.11",
  "python_version": "3.10",
  "sys_platform": "win32"
}
```

### Verbose Command Output

_No response_

### Additional Context

_No response_

- [x] Yes, I would like to submit a PR.

monchin:
… from list[str](#3523)

## Pull Request Checklist

- [√] A news fragment is added in `news/` describing what is new.
- [√] Test cases added for changed code.

fix #3523

### DIFF
--- a/news/3523.bugfix.md
+++ b/news/3523.bugfix.md
@@ -1,0 +1,1 @@
+Change the return type of `array_of_inline_tables` to list[dict] from list[str]
--- a/src/pdm/formats/base.py
+++ b/src/pdm/formats/base.py
@@ -100,5 +100,5 @@ def make_array(data: list, multiline: bool = False) -> list:
     return array
 
 
-def array_of_inline_tables(value: list[Mapping], multiline: bool = True) -> list[str]:
+def array_of_inline_tables(value: list[Mapping], multiline: bool = True) -> list[dict]:
     return make_array([make_inline_table(item) for item in value], multiline)
--- a/src/pdm/formats/poetry.py
+++ b/src/pdm/formats/poetry.py
@@ -112,7 +112,7 @@ NAME_EMAIL_RE = re.compile(
 )
 
 
-def parse_name_email(name_email: list[str]) -> list[str]:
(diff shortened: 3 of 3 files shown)


monchin:
pre-commit shows that
```python
class PoetryMetaConverter(MetaConverter):
    @convert_from("authors")
    def authors(self, value: list[str]) -> list[str]:
        return parse_name_email(value)

    @convert_from("maintainers")
    def maintainers(self, value: list[str]) -> list[str]:
        return parse_name_email(value)
```
failed by mypy lint. 

According to [pep 621](https://peps.python.org/pep-0621/#authors-maintainers), both `authors` and `maintainers` should be `list[dict]`, so I'll change it.

frostming (approved):

Issue #3543, opened 2025-06-13
dependency-groups in pylock.toml output is unstable over multiple invocations


akaIDIOT:
### Describe the bug

While trying out the new lockfile format, I noticed the top level key for `dependency-groups` would change after every invocation. Guessing it's caused by (at least) `Project.iter_groups()` using a `set` for deduplication. 

### To reproduce

Using a project with at least one entry under `dependency-groups` in its `pyproject.toml`, run `pdm lock` (or `pdm update --update-all`) multiple times. The value in the resulting `pylock.toml` file under `dependency-groups` can / will change over the multiple runs.

### Expected Behavior

The contents of `pylock.toml` will not change (or at least: the `dependency-groups` key) over the multiple invocations.

### Environment Information

```plain
PDM version:
  2.25.0
Python Interpreter:
  /home/akaidiot/Documents/Projects/confidence/.venv/bin/python (3.12)
Project Root:
  /home/akaidiot/Documents/Projects/confidence
Local Packages:
  
{
  "implementation_name": "cpython",
  "implementation_version": "3.12.3",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.8.0-60-generic",
  "platform_system": "Linux",
  "platform_version": "#63-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 19:04:15 UTC 2025",
  "python_full_version": "3.12.3",
  "platform_python_implementation": "CPython",
  "python_version": "3.12",
  "sys_platform": "linux"
}
```

### Are you willing to submit a PR to fix this bug?

- [x] Yes, I would like to submit a PR.

PR #3544, opened 2025-06-13
Stabilize ordering of extras and dependency-groups in pylock output

akaIDIOT:
## Pull Request Checklist

As described in #3543, multiple locks of the same project would jitter around the output of `dependency-groups` in the resulting `pylock.toml` file. Looking at the code that produces this, it outputs the result of a number of possible sources, at least one of which (`Project.iter_group()`) using a `set` to collect possible groups. In order to minimize the impact of this change, I opted to sort the output just before it's being output to avoid changing behaviour anywhere else. This could of course be moved further up the stack. 

Currently didn't add any tests, would this change warrant its own test?

### DIFF
--- a/src/pdm/formats/pylock.py
+++ b/src/pdm/formats/pylock.py
@@ -127,8 +127,8 @@ class PyLockConverter:
                 "lock-version": self.lock_version,
                 "requires-python": str(project.python_requires),
                 "environments": make_array([str(marker) for marker in env_markers], multiline=True),
-                "extras": extras,
-                "dependency-groups": groups,
+                "extras": sorted(extras),
+                "dependency-groups": sorted(groups),
                 "default-groups": ["default"],
                 "created-by": "pdm",
             }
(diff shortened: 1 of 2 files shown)


frostming (approved):

src/pdm/formats/pylock.py
```diff
@@ -127,8 +127,8 @@ def convert(self, all_groups: Iterable[str] | None = None) -> dict[str, Any]:
                 "lock-version": self.lock_version,
                 "requires-python": str(project.python_requires),
                 "environments": make_array([str(marker) for marker in env_markers], multiline=True),
-                "extras": extras,
-                "dependency-groups": groups,
+                "extras": sorted(extras),
+                "dependency-groups": sorted(groups),
```
frostming, line 131:
```suggestion
                "dependency-groups": sorted(groups， key=_group_sort_key),
```

Issue #3766, opened 2026-04-15
build-system config in a template is overwritten during `pdm init template/`


akaIDIOT:
### Describe the bug

A `[build-system]` table in a `pyproject.toml` meant to be used with `pdm init` is ignored. A template I've been using was expanded to contain an additional build-system dependency to do with versioning. After this change, the template would create broken projects, as the additional dependency in `build-system.requires` is missing, causing builds to crash.

### To reproduce

Given the following minimal `pyproject.toml` in a folder like `template-folder/` on its own:

```toml
[build-system]
requires = ["pdm-backend", "example"]
build-backend = "pdm.backend"

[project]
name = "name"
```

After using it to create a new project with `pdm init template-folder/`, using the default values for all questions but the distribution (answering `y`), the resulting `pyproject.toml` contains a `[build-system]`, but not as the template supplied it:

```toml
[build-system]
requires = ["pdm-backend"]
build-backend = "pdm.backend"

[project]
name = "name"
...
```

When answering `n` at the distribution question, `[build-system]` is completely removed.

### Expected Behavior

If a template configures the build-system, in particular using the same build backend as `pdm init` would have used, I'd expect the additional configuration supplied with the template to remain intact. 

If the user of the template would request a different build backend to be used during the questionnaire, it might make sense to carry over any additions to the build systems, though that might make it fairly complicated.

### Environment Information

```console
$ pdm info && pdm info --env
PDM version:
  2.26.7
Python Interpreter:
  /tmp/clean/.venv/bin/python (3.12)
Project Root:
  /tmp/clean
Local Packages:
  /tmp/clean/.venv/lib/python3.12/site-packages
{
  "implementation_name": "cpython",
  "implementation_version": "3.15.0a8",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.17.0-20-generic",
  "platform_system": "Linux",
  "platform_version": "#20~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Mar 19 01:28:37 UTC 2",
  "python_full_version": "3.12.3",
  "platform_python_implementation": "CPython",
  "python_version": "3.12",
  "sys_platform": "linux"
}
```

### Verbose Command Output

```console
$ pdm init --verbose /tmp/template/
Creating a pyproject.toml for PDM...
Please enter the Python interpreter to use
 0. cpython@3.12 (/usr/bin/python)
 1. cpython@3.15t (/usr/bin/python3.15t)
[...]

PR #3769, opened 2026-04-17
Accept build system extensions from template on matching backend

akaIDIOT:
## Pull Request Checklist

This PR changes the behaviour of `pdm init <template>` when the template has additions to the `[build-system]` table. When the user's choice of build backend matches that of the template, the template's build system is merged with the user choice. When the user chooses a different build backend, the additions in the template are ignored and the chosen backend is used as it would have previously. This is the expected behaviour as described in #3766.

I'm not happy about having to use a deep copy of the template's `[build-system]` table, but `merge_dictionary` would reuse values from that dict and edit them later, while the changes in this PR would need the original value. This *could* be changed in `merge_dictionary` by using copies there, but I figured the single copy in the template handling would be simpler.

The added test was mostly copied from the test above it, edited to reflect the intentions of the change in this PR.

### DIFF
--- a/news/3766.bugfix.md
+++ b/news/3766.bugfix.md
@@ -1,0 +1,1 @@
+Fix `pdm init <template>` overwriting the template's additions to `build-system.requires` when `build-system.build-backend` matches the user's selection.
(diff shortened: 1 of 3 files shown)


frostming (commented):
LGTM

PR #3774, opened 2026-04-28
Remove dead code

duriantaco:
Remove unused dead code identified by static analysis:                                         
                                                                                                 
  - `ProjectInfo.from_metadata()`.. not called (removes `DYNAMIC` constant and `Any` import  too)                                                                                           
  - `ExtrasWarning` .. not raised                                                               
  - `SilentSpinner`.. not instantiated                    
  - `exec_opts` .. not referenced                                                         
  - `InstallManager.NO_CACHE_PACKAGES`.. never read
  - `PySpecSet.MAX_MAJOR_VERSION`.. never read                                                   
  - `Version.MIN` / `Version.MAX` .. declared but not used?                           
  - `Project.DEPENDENCIES_RE`.. never read      
  
No tests needed.. This is only removal of unreachable code and nth else.     

Dead code was identified by using [Skylos](https://github.com/duriantaco/skylos) and confirmed via manual verification.

### DIFF
--- a/src/pdm/models/versions.py
+++ b/src/pdm/models/versions.py
@@ -26,6 +26,8 @@ class Version:
     focused on supporting PEP 440 version identifiers, not specifiers.
     """
 
+    MIN: Version
+    MAX: Version
     # Pre-release may follow version with {a|b|rc}N
     # https://docs.python.org/3/faq/general.html#how-does-the-python-version-numbering-scheme-work
     pre: tuple[str, int] | None = None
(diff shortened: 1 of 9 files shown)


src/pdm/models/versions.py
```diff
@@ -26,8 +26,6 @@ class Version:
     focused on supporting PEP 440 version identifiers, not specifiers.
     """
 
-    MIN: Version
-    MAX: Version
```
frostming, line 30: keep this

duriantaco:
Thanks for the heads up.. restored!

PR #3782, opened 2026-05-08
Allow `exclude-newer` to be defined in `pyproject.toml`

### LINKED ISSUE #3776: Dependency Cooldowns
### Feature Description

A field in `pyproject.toml` to set a dependency cooldown, where versions in the lockfile are not updated until the release is at least as old as the cooldown window.

Mimicking pip's implementation: https://pip.pypa.io/en/stable/user_guide/#filtering-by-upload-time

That could look something like

```toml
[tool.pdm.resolution]
uploaded-prior-to = "P7D"
```

Or for package-specific overrides (consider this pseudocode for the purposes of illustration because I am not an expert in toml, and would work out the proper syntax on implementation)

```toml
[tool.pdm.resolution]
uploaded-prior-to = "P7D"

[tool.pdm.resolution.uploaded-prior-to]
# use the most recent version
mypackage = false

# use another cooldown
anotherpackage = "P14D"
```

(If the global setting is not set, then one could set cooldowns only for specific packages)

Pdm already has an `exclude-newer` flag: https://pdm-project.org/latest/usage/lockfile/#exclude-packages-newer-than-specific-date

so the simplest approach would be to allow that to be set in the pyproject.toml - I don't have a preference for naming, just using pip's names for illustration.

### Problem and Solution

E.g. see here: https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

Many package managers, including pip ( https://sethmlarson.dev/pip-relative-dependency-cooldowns ) are adding a "dependency cooldown" feature that sets some window of time where a dependency is not updated until a version is as old as the window.

This helps mitigate problems from supply chain attacks, where the impact of a compromised release can be decreased by giving a window of time for it to be detected before other projects start using it. 

### Additional Context

I have been looking for ways to contribute here because I love the tool and want to help out where I can, and this seems like a simple enough feature that I would be able to do it without causing you all too much review or coaching burden!

sneakers-the-rat:
## Pull Request Checklist

Fix: #3776 - allow setting dependency cooldowns on a projectwide basis from the `pyproject.toml` file. pretty straightforward, i'll comment inline with a few things i wasn't sure about.

There didn't seem to be a single place to add a field to `pyproject.toml`, and looking around at how other things in the `resolution` table made it seem like it was on an as-needed kind of basis since the `PyProject.resolution` prop is just an untyped dict, so i just walked backwards from how the cli flag was handled until i found the setting in the `State` object. I *think* i found the place in `core` to apply settings from pyproject -> state, but lmk if that is too much of a hack/what would be better.

### DIFF
--- a/docs/usage/lockfile.md
+++ b/docs/usage/lockfile.md
@@ -199,6 +199,16 @@ pdm lock --exclude-newer 2024-01-01
 pdm lock --exclude-newer 7d
 ```
 
+!!! tip
+    Added in 2.26.9.
+
+`exclude-newer` may also be set in the `pyproject.toml` file under the `[tool.pdm.resolution]` table:
+
+```toml
+[tool.pdm.resolution]
+exclude-newer = "7d"
+```
+
 !!! note
     The package index must support the `upload-time` field as specified in [PEP 700]. If the field is not present for a given distribution, the distribution will be treated as unavailable.
 
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -34,7 +34,7 @@ from pdm.installers import InstallManager
 from pdm.models.repositories import BaseRepository, PyPIRepository
 from pdm.project import Project
 from pdm.project.config import Config
-from pdm.utils import is_in_zipapp
+from pdm.utils import convert_to_datetime, is_in_zipapp
 
 if TYPE_CHECKING:
     from typing import Any, Iterable
@@ -135,6 +135,8 @@ class Core:
                 global_config=options.config or os.getenv("PDM_CONFIG_FILE"),
             )
         self.state.build_isolation = project.config["build_isolation"]
+        if exclude_newer := project.pyproject.resolution.get("exclude-newer"):
+            self.state.exclude_newer = convert_to_datetime(exclude_newer)
         return project
 
     def create_project(
--- a/src/pdm/core.py
+++ b/src/pdm/core.py
@@ -135,8 +135,6 @@ class Core:
                 global_config=options.config or os.getenv("PDM_CONFIG_FILE"),
             )
         self.state.build_isolation = project.config["build_isolation"]
-        if exclude_newer := project.pyproject.resolution.get("exclude-newer"):
-            self.state.exclude_newer = convert_to_datetime(exclude_newer)
         return project
 
     def create_project(
@@ -190,6 +188,9 @@ class Core:
 
         if overrides := getattr(options, "override", None):
             self.state.overrides = overrides
+
+        if exclude_newer := project.pyproject.resolution.get("exclude-newer"):
+            self.state.exclude_newer = convert_to_datetime(exclude_newer)
 
         if command is None:
             self.parser.print_help()
--- a/tests/cli/test_lock.py
+++ b/tests/cli/test_lock.py
@@ -343,18 +343,25 @@ def test_lock_inherit_metadata_strategy(project, pdm, mocker, working_set):
         assert key in working_set
 
 
-def test_lock_exclude_newer(project, pdm):
+@pytest.mark.parametrize("source", ["cli", "pyproject"])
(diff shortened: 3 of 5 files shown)


docs/usage/lockfile.md
```diff
@@ -199,6 +199,16 @@ pdm lock --exclude-newer 2024-01-01
 pdm lock --exclude-newer 7d
 ```
 
+!!! tip
+    Added in 2.26.9.
```
sneakers-the-rat, line 203: this is just a guess on the version and can change. i find these markers really helpful and thought it might prevent people from being like 'what the heck how come i can't set this' after seeing it in the docs without updating.

tests/cli/test_lock.py
```diff
@@ -343,18 +343,25 @@ def test_lock_inherit_metadata_strategy(project, pdm, mocker, working_set):
         assert key in working_set
 
 
-def test_lock_exclude_newer(project, pdm):
+@pytest.mark.parametrize("source", ["cli", "pyproject"])
+def test_lock_exclude_newer(project, pdm, source):
```
sneakers-the-rat, line 347: switched the order on this test, it seemed like the `3.7.0` check is just a sanity check to make sure that `3.6.0` is actually not the latest version so it didn't seem like order mattered. 

since the value gets stored in the `state` now, and it doesn't get unconditionally overridden as it did before, having it come second caused that part to fail, but it seems like that is just an artifact of the testing environment and in practice there wouldn't be stale state hanging around since it gets created on each invocation.

tests/cli/test_lock.py
```diff
@@ -343,18 +343,25 @@ def test_lock_inherit_metadata_strategy(project, pdm, mocker, working_set):
...
+        project.pyproject.settings["resolution"] = {"exclude-newer": "2024-01-01"}
+        cmd = ["lock"]
+
+    pdm(cmd, obj=project, strict=True, cleanup=False)
+    assert project.get_locked_repository().candidates["zipp"].version == "3.6.0"
 
-def test_lock_exclude_newer_accepts_relative_duration(project, pdm, monkeypatch):
+
+@pytest.mark.parametrize("source", ["cli", "pyproject"])
+def test_lock_exclude_newer_accepts_relative_duration(project, pdm, monkeypatch, source):
```
sneakers-the-rat, line 364: the parameterization is a bit hacky, but i thought it was better to hook into the existing tests than duplicate them except with pyproject rather than cli, lmk if you would prefer to do it that way rather than parameterized

src/pdm/core.py
```diff
@@ -135,6 +135,8 @@ def ensure_project(self, options: argparse.Namespace, obj: Project | None) -> Pr
                 global_config=options.config or os.getenv("PDM_CONFIG_FILE"),
             )
         self.state.build_isolation = project.config["build_isolation"]
+        if exclude_newer := project.pyproject.resolution.get("exclude-newer"):
+            self.state.exclude_newer = convert_to_datetime(exclude_newer)
```
frostming: LGTM, just a minor tweak, can we move L137-L139 to `handle` method, to clear the side effects of this method.

sneakers-the-rat: no prob

tests/cli/test_lock.py
frostming: That is okay

src/pdm/core.py
sneakers-the-rat: you mean `Core.handle` or the lock command's handle method?

sneakers-the-rat: i assumed you meant `Core.handle` so i went ahead and did that here because it's simple enough to change: https://github.com/pdm-project/pdm/pull/3782/commits/7dcf95fdcaf9bff233b7073ada8f97d41e13a4ae

but i can also move that to the lock command `handle` method if that's what you meant

PR #3784, opened 2026-05-09
Docs and TypedDicts for tool.pdm table and subtables

### LINKED ISSUE #3783: Document possible values of `[tool.pdm]` table in one place
### Feature Description

Searched and couldn't find an issue for this, but i was sort of surprised there wasn't one so feel free to close if it's a duplicate.

---

Create dataclasses/typed dicts for all the possible values that can be set in `[tool.pdm]` and its subtables, use them internally, and then also use them to autogenerate docs in a single place

### Problem and Solution

There are a lot of extra options that can be set in `[tool.pdm]` and the subtables, but their documentation is spread out over a few pages and done inline with the narrative docs (which is great!), so it's hard to know "what are all the possible config values."

The processing of the pyproject.toml values can also be a bit ad-hoc, where e.g. in #3782 I had to convert the string to the datetime at the point when the value would be used, rather than having it be immediately converted as a byproduct of parsing the pyproject.toml file.

So it would be nice to have dataclasses for these both for the sake of documentation, but also to standardize how fields are added to `pyproject.toml` and how they are handled internally.

### Additional Context

_No response_

sneakers-the-rat:
## Pull Request Checklist

- [x] Test cases added for changed code - none needed, as far as i'm aware

Fix: #3783 

(raising this as a draft but i have to run right now and can't do a full rundown just yet, wanted to open to see if you approved of the general approach)

So i figured out how to get mkdocs to do a summary-table style like what is done in the other refrence docs pages, but it is a bit of a hack. the big thing that is missing here is that functional typeddicts get rendered like garbage, so i think i'll need to write a small extension for that, if you would accept that ( opened a discussion: https://github.com/mkdocstrings/python/discussions/329 ) 

Screenshot 2026-05-08 at 7 07 29 PM

Getting that summary form with the correct TOC headings also required doing manual autodoc invocations for each class, which isn't so bad, but it does risk getting stale if there are a bunch of new tables, but not that bad.

I went through the docs and source to find all mentions of `[tool.pdm]` entries i could find, and added some docstrings as examples, but if we generally like this approach then i can go through and fill those in more completely (or you can do that, no preference).

Feel free to edit anything, if i've gotten something wrong, if you don't like my working, etc. i'm not precious - these are your docs, just trying to help out!

will be back later for more complete comment and to figure out an approach for the functional typed dicts.

## todo

- python 3.9 compat - use `Union`
- extension to document functional typed dicts
- ... will fill in later

### DIFF
--- a/docs/_ext/typeddict.py
+++ b/docs/_ext/typeddict.py
@@ -1,0 +1,111 @@
+"""
+griffe extension for documenting typed dicts:
+
+Converts functional-form TypedDicts to class-form.
+Functional TypedDicts may document their fields with comments in place of docstrings.
+
+e.g.
+
+```
+MyDict = TypedDict("MyDict", {
+    "a-field": str,
+    # This is interpreted as a docstring
+    # and it can span multiple lines
+})
+```
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Any, cast
+
+from griffe import (
+    Attribute,
+    Class,
+    Docstring,
+    ExprCall,
+    ExprDict,
+    ExprName,
+    Extension,
+    GriffeLoader,
+    Module,
+)
+
+
+class TypedDictExtension(Extension):
+    def __init__(self, modules: list[str] | None = None):
+        """
+        Args:
+            modules (list[str]): fully-qualified python module names
+                (e.g. "pdm.mproject.project_file").
+                If present, only transform TypedDicts in these modules.
+        """
+        self.modules = modules
+
+    def on_module(self, *, mod: Module, loader: GriffeLoader, **kwargs: Any) -> None:
+        # need to intercept at the module level because we want to modify the types of its members
+        # (otherwise we can only mutate the member's attributes, but not replace it)
+        if self.modules and mod.module.canonical_path not in self.modules:
+            return
+
+        self._transform_functional_typeddicts(mod)
+
+    def _transform_functional_typeddicts(self, mod: Module) -> Module:
+        """
+        Transform functional typed dicts into class-based typed dicts
+        We only do the minimal work needed here to create the representation in the docs -
+        just grab the field names and annotations, so other metadata will be incomplete.
+        """
+        tds = {
+            k: v
+            for k, v in mod.members.items()
+            if isinstance(v, Attribute)
+            and "module-attribute" in v.labels
+            and isinstance(v.value, ExprCall)
+            and v.value.function.name == "TypedDict"
+        }
+
+        classes = {}
+        for name, td in tds.items():
+            classes[name] = Class(name=name, bases=[ExprName(name="TypedDict", parent=mod, member=name)], parent=mod)
+            classes[name].docstring = td.docstring
+
+            # Convert the keys/values of the dict arg to classlike attributes
+            expression = cast(ExprCall, td.value)
+            fields = cast(ExprDict, expression.arguments[1])
+            attrs = {}
+            for key, val in zip(fields.keys, fields.values):
+                key: str
+                key = key.strip("'")
+                attrs[key] = Attribute(name=key, value=val, annotation=val)
+                attrs[key].parent = classes[name]
+                attrs[key].docstring = self._extract_comment_docstring(key, td)
+
+            classes[name].members = attrs
+
+        mod.members.update(classes)
+        return mod
+
+    def _extract_comment_docstring(self, key: str, td: Attribute) -> Docstring | None:
+        """
+        find comments after this expression to use as docstrings.
+        The ExprName we use to parse the key/annotation doesn't bear the line numbers,
+        so regex is unfortunately the best we can do
+        """
+        expr_line = [idx for idx, line in enumerate(td.lines) if re.match(rf'^\s*[\'"]{key}[\'"]', line)]
+        if not expr_line:
+            return
+        start_line = expr_line[0]
+
+        # find trailing comments, if any
+        docstring_lines = []
+        for line in td.lines[start_line + 1 :]:
+            if not re.match(r"^\s*#", line):
+                break
+            docstring_lines.append(line.strip().lstrip("#").lstrip())
+
+        # join like a naive markdown, not handling line breaks currently
+        docstring = " ".join(docstring_lines)
+        if docstring:
+            return Docstring(value=docstring)
--- a/docs/reference/tool.md
+++ b/docs/reference/tool.md
@@ -33,8 +33,6 @@
       summary:
         attributes: true
 
-(placeholder pending figuring out rendering docs for functional form typeddicts)
-
 ::: pdm.project.project_file.OptionsTable
     options:
       heading: "[tool.pdm.options]"
@@ -60,8 +58,6 @@
       show_if_no_docstring: true
       summary:
         attributes: true
-
-(placeholder pending figuring out rendering docs for functional form typeddicts)
 
 ::: pdm.project.project_file.UserScript
     options:
@@ -101,21 +97,3 @@
       show_if_no_docstring: true
       summary:
         attributes: true
-
-## Functional TypedDicts
-
-::: pdm.project.project_file
-    options:
-      show_root_heading: false
-      show_root_toc_entry: false
-      show_source: false
-      heading_level: 3
-      group_by_category: false
-      show_bases: false
-      show_if_no_docstring: true
-      show_labels: false
-      show_signature_annotations: true
-      signature_crossrefs: true
-      members:
-      - BuildTable
-      - ResolutionTable
--- a/docs/reference/tool.md
+++ b/docs/reference/tool.md
@@ -1,0 +1,121 @@
+# `[tool.pdm]`
+
+<style>
+/* Temporary hack pending guidance to show compact summary tables without double-displaying each attribute */
+.doc-class .doc-children {
+  display: none;
+}
+</style>
+
+::: pdm.project.project_file.ToolPDMTable
+    options:
+      heading: "[tool.pdm]"
+      toc_label: "[tool.pdm]"
+      heading_level: 2
+      show_bases: false
+      show_root_heading: true
+      show_root_toc_entry: true
+      show_source: false
+      show_if_no_docstring: true
+      summary:
+        attributes: true
+
+::: pdm.project.project_file.BuildTable
+    options:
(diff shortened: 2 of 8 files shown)


src/pdm/project/project_file.py
```diff
@@ -129,3 +132,170 @@ def content_hash(self, algo: str = "sha256") -> str:
...
+    """Control how pdm-backend builds the package"""
+    options: OptionsTable
+    """Add parameters to every pdm cli call"""
+    resolution: ResolutionTable
+    """Rules governing lockfile resolution"""
+    scripts: dict[str, str | UserScript]
+    """See [pdm-scripts][pdm-scripts]"""
+    source: list[SourceTable]
+    """Repositories where packages may be found"""
+    version: VersionTable
```
frostming: I do not recommend adding fields that are not used by this project to the schema, since they may be changed by the consumers.

sneakers-the-rat: are you referring to the version table? I am not sure what you mean 'not used by this project' - do you mean that `version` is interpreted by `pdm-backend` and not `pdm` and should go there? or what do you mean?

frostming: Yes, if this table is not directly used by PDM itself, it's better to no include it here.

sneakers-the-rat:
there we go! look at that, so nice: https://python-pdm--3784.org.readthedocs.build/en/3784/reference/tool/

got the mkdocs transformation of functional TypedDicts -> class TypedDicts working with the dict keys that can't be python names:

Screenshot 2026-05-08 at 11 49 59 PM

will annotate a few choices here but then i think this is ready for review - basically up to you on what fields need docstrings, what should go in them, etc. feel free to edit anything or ask further questions

docs/_ext/typeddict.py
```diff
@@ -0,0 +1,111 @@
...
+        We only do the minimal work needed here to create the representation in the docs -
+        just grab the field names and annotations, so other metadata will be incomplete.
+        """
+        tds = {
+            k: v
+            for k, v in mod.members.items()
+            if isinstance(v, Attribute)
+            and "module-attribute" in v.labels
+            and isinstance(v.value, ExprCall)
+            and v.value.function.name == "TypedDict"
```
sneakers-the-rat, line 66: intercepting at the module level means that the members are pretty polytyped, so these checks are both to filter to only functional typeddicts and also to avoid attribute errors/satisfy mypy/etc.

docs/_ext/typeddict.py
```diff
@@ -0,0 +1,111 @@
...
+                attrs[key] = Attribute(name=key, value=val, annotation=val)
+                attrs[key].parent = classes[name]
+                attrs[key].docstring = self._extract_comment_docstring(key, td)
+
+            classes[name].members = attrs
+
+        mod.members.update(classes)
+        return mod
+
+    def _extract_comment_docstring(self, key: str, td: Attribute) -> Docstring | None:
```
sneakers-the-rat, line 90: wrote this not to be super general, assuming we are defining things in a standard way that should be enforced by the linter (fields and comments on newlines, etc.)

the failure mode should be missing docstrings rather than failing building if any errors should occur, but it looks like it passes through all the markdown just fine

docs/reference/tool.md
```diff
@@ -0,0 +1,99 @@
+# `[tool.pdm]`
+
+<style>
+/* Temporary hack pending guidance to show compact summary tables without double-displaying each attribute */
+.doc-class .doc-children {
+  display: none;
+}
+</style>
```
sneakers-the-rat, line 8: so this is the remaining hack that i'm not sure about - the purpose is to make it so we just get the summary tables but don't get the larger attribute listings. i assumed we would like the former because it's more in the style of the other docs, but reverting this back to just being normal attribute listings wouldn't be hard. one bonus is that the summary tables link out to the types. 

both a) linking to types from attribute annotations and b) displaying summary tables without attribute listings are dropped features during the transition into zensical:

- https://github.com/mkdocstrings/python/issues/270
- https://github.com/mkdocstrings/python/issues/139 

I am not sure if it would be possible for me to hack this together with the extension, since the summary table is generated from the members, so while inline CSS is ugly it is scoped to just this one page and it does do the trick.

src/pdm/cli/commands/fix/fixers.py
```diff
@@ -74,7 +74,7 @@ class PackageTypeFixer(BaseFixer):  # pragma: no cover
     identifier = "package-type"
 
     def get_message(self) -> str:
-        package_type = self.project.pyproject.settings["package-type"]
+        package_type = self.project.pyproject.settings["package-type"]  # type: ignore[typeddict-item]
```
sneakers-the-rat, line 77: ignored these because this is a deprecated key and it probably doesn't go in the docs

src/pdm/project/core.py
```diff
@@ -499,8 +500,8 @@ def sources(self) -> list[RepositoryConfig]:
 
     def get_sources(self, expand_env: bool = True, include_stored: bool = False) -> list[RepositoryConfig]:
         result: dict[str, RepositoryConfig] = {}
-        for source in self.pyproject.settings.get("source", []):
-            result[source["name"]] = RepositoryConfig(**source, config_prefix="pypi")
+        for project_source in self.pyproject.settings.get("source", []):
```
sneakers-the-rat, line 503: mypy doesn't handle changing the type of a same-named variable, so changing the name is just for satisfying mypy

src/pdm/project/core.py
```diff
@@ -975,7 +976,7 @@ def get_setting(self, key: str) -> Any:
         Returns `None` if the key does not exists.
         """
         try:
-            return reduce(operator.getitem, key.split("."), self.pyproject.settings)
+            return reduce(operator.getitem, key.split("."), self.pyproject.settings)  # type: ignore[arg-type]
```
sneakers-the-rat, line 979: i am sort of a python type system head and i genuinely don't know how one would correct this mypy error:

> Argument 1 to "reduce" has incompatible type overloaded function; expected "Callable[[ToolPDMTable, str], ToolPDMTable]"

something with the generic types in `reduce` but also it's saying that the `list[str]` part in arg 1 is the problem? no idea. it would probably be cleanest to just remove this method and directly access the keys, but i didn't want to touch the rest of the code outside the scope of the PR.

src/pdm/project/project_file.py
```diff
@@ -129,3 +132,175 @@ def content_hash(self, algo: str = "sha256") -> str:
...
+)
+"""
+`[tool.pdm.build]`
+
+References:
+    <https://backend.pdm-project.org/build_config/>
+"""
+
+
+class OptionsTable(TypedDict, total=False):
```
sneakers-the-rat, line 162: this might not need a typeddict, or there might be some way to generate it dynamically, so this is just for completeness sake but could be removed/replaced with something that doesn't hardcode the command names like this.

mkdocs.yml
```diff
@@ -36,12 +36,16 @@ theme:
   custom_dir: docs/overrides
 
 plugins:
+  - autorefs
```
sneakers-the-rat, line 39: needed this to be able to refer to other parts of the docs from the docstrings like `[Display Text][reference-to-anchor]` - less fragile than direct URL references because they work locally and also the build process can warn if the link breaks.

sneakers-the-rat:
I'll pitch upstream mkdocstrings-python on adding this as a builtin extension so we can take it out of pdm's responsibility

docs/reference/tool.md
pawamoy: You could use custom mkdocstrings-python templates to avoid rendering the attributes (while still displaying the summary table). This way, no objects inventory pollution and no CSS needed.

sneakers-the-rat: Is there a way to control which template is used on a per-object/per-module level? i imagine we might want to control omitting the attrs/members in some cases and not others. it seems like that issue 139 might be the fix that's needed upstream, it does feel like something that should have an intuitive control - "show me summary, attrs, both, or neither" when currently only "show me summary+attrs, attrs, or niether" is possible.

pawamoy: > Is there a way to control which template is used on a per-object/per-module level?

You can override, say, the base children template, in which you check each attribute metadata:

```jinja
{% if not attr.extra.your_typeddict_extension_name.dont_render %}
    ...  {# proceed #}
{% endif %}
```

See https://mkdocstrings.github.io/griffe/guide/users/extending/?h=extra#extra-data.

I don't think the children template uses Jinja blocks though so it could mean a big copy-paste :thinking:

pawamoy: Heh, maybe just override the base attribute template and check the metadata to decide whether to render anything at all. This one has Jinja blocks, much easier.

pawamoy: Another idea: set `attr.extra["mkdocstrings"]["template"] = "empty.html.jinja"`. This needs either an entry-point (see https://mkdocstrings.github.io/usage/handlers/#handler-extensions), or you can just add this empty template to the folder pointed at by the `custom_templates` setting (see https://mkdocstrings.github.io/usage/theming/?h=custom_#templates).

sneakers-the-rat: thanks for the tips! i'll wait to see what the maintainers here say to see what they think is best. 

separately - since this seems like a pretty straightforward need that would probably come up pretty often (this combination of typed dict user-facing documentation and wanting to toggle between an abbreviated form for things where it makes sense vs. an expanded form when more documentation is needed. i mean if mkdocs did this more elegantly than sphinx i would switch over) would you be interested in a PR upstream to handle this? imo it seems like a three way switch: summary is "off/only/both" but i could also do the suggestion in your issue 139 et al of having it be a separate bool flag

pawamoy: Happy to help! 

mkdocstrings and handlers are in maintenance mode so we won't implement new features there.

sneakers-the-rat: ig idk where to contribute then! lmk where if anywhere. thanks again.

src/pdm/project/project_file.py
sneakers-the-rat: sorry missed the reply - so should I split this into two and pull the `pdm-backend`-specific changes over there?

frostming: You can just exclude them from the schema, unless they are directly referenced by PDM code

Issue #3673, opened 2025-11-13
Improve the startup performance


frostming:
### Feature Description

Now PDM startup is still very slow, most of the time is occupied by importing modules.

### Problem and Solution

If possible, make the imports lazy or wrap them in `if TYPE_CHECKING` when appropriate.

Before that, we should identify the hotspot by profiling PDM. Here are some recommendations: https://github.com/pdm-project/pdm/issues/3669#issuecomment-3524604799

### Are you willing to contribute to the development of this feature?

- [ ] Yes, I am willing to contribute to the development of this feature.

Mr-Sunglasses:
> ### Feature Description
> 
> Now PDM startup is still very slow, most of the time is occupied by importing modules.
> 
> ### Problem and Solution
> 
> If possible, make the imports lazy or wrap them in `if TYPE_CHECKING` when appropriate.
> 
> Before that, we should identify the hotspot by profiling PDM. Here are some recommendations: https://github.com/pdm-project/pdm/issues/3669#issuecomment-3524604799
> 
> ### Additional Context
> 
> _No response_
> 
> ### Are you willing to contribute to the development of this feature?
> 
> - [ ] Yes, I am willing to contribute to the development of this feature.

Hey @frostming I would love to work on this issue!! Could you please assign it to me!!

frostming:
Unassigned to release this issue to the community.

Future contributors don't need to ask for assignment, send the PR directly and I will review.

PR #3805, opened 2026-06-11
perf: defer command imports during startup

Yijian6:
## What

Defer imports used only by the `self`, `python`, and `use` command execution paths so they are not loaded while PDM is only starting up and registering CLI commands.

## Why

Refs #3673. PDM currently imports every command module during parser initialization, so top-level imports in rarely used commands contribute to startup cost even for simple invocations such as `pdm --version`.

## How

Move execution-only imports for package working sets, Python discovery/install helpers, environment helpers, and update checks into the functions or handlers that actually need them. The command registration and argument definitions are unchanged.

A local comparison using 7 runs of `python -m pdm --version` with the same virtualenv showed:

- `main`: min 573.50 ms, avg 828.56 ms, max 2294.72 ms
- this branch: min 550.61 ms, avg 572.28 ms, max 613.83 ms

## Testing

- `.\.venv\Scripts\python.exe -m pytest tests\cli\test_self_command.py -q` -> 8 passed
- `.\.venv\Scripts\ruff.exe check src\pdm\cli\commands\self_cmd.py src\pdm\cli\commands\python.py src\pdm\cli\commands\use.py tests\cli\test_self_command.py` -> passed
- `.\.venv\Scripts\python.exe -m compileall src\pdm\cli\commands\self_cmd.py src\pdm\cli\commands\python.py src\pdm\cli\commands\use.py tests\cli\test_self_command.py` -> passed
- `.\.venv\Scripts\pdm.exe --version`, `self --help`, `python --help`, and `use --help` -> passed

I also tried `tests\cli\test_python.py tests\cli\test_use.py`; on my Windows machine they fail before exercising this change because subprocess output from the local Python path under `C:\Users\觉\...` is decoded as UTF-8 while the child process emits the username in the system code page.

## Breaking changes

None.

### DIFF
--- a/news/3673.feature.md
+++ b/news/3673.feature.md
@@ -1,0 +1,1 @@
+Defer startup-time imports for Python and self-management commands.
--- a/src/pdm/cli/commands/python.py
+++ b/src/pdm/cli/commands/python.py
@@ -10,16 +10,14 @@ from typing import TYPE_CHECKING, cast
 
 from pdm.cli.commands.base import BaseCommand
 from pdm.cli.options import verbose_option
-from pdm.environments import BareEnvironment
 from pdm.exceptions import InstallationError, PdmArgumentError
-from pdm.models.python import PythonInfo
 from pdm.termui import Verbosity
-from pdm.utils import get_all_installable_python_versions
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace, _SubParsersAction
     from typing import Any
 
+    from pdm.models.python import PythonInfo
     from pdm.project.core import Project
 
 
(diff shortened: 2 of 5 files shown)


yihong0618:
I think we can wait for 3.15 use lazy import later?

frostming:
> I think we can wait for 3.15 use lazy import later?

that's a long way to go. and i guess we don't have a path to backport, do we?

yihong0618:
> > I think we can wait for 3.15 use lazy import later?
> 
> that's a long way to go. and i guess we don't have a path to backport, do we?

make sense

frostming (approved):

PR #3806, opened 2026-06-15
perf: defer virtualenv command imports

Yijian6:
## What

Defer virtualenv command runtime imports until the matching `pdm venv` command actually executes.

## Why

Refs #3673. Command registration was still importing virtualenv helpers, backend classes, and shell detection support during startup, even for commands like `pdm --version` that do not need virtualenv functionality.

## How

- Move venv helper/backend imports from module import time into command handlers.
- Keep the `pdm venv create --with` choices available without importing backend implementations during command registration.
- Add a regression test that initializes `Core()` in a clean subprocess and asserts the venv runtime modules are not loaded.

## Testing

- `ruff check src/pdm/cli/commands/venv/__init__.py src/pdm/cli/commands/venv/activate.py src/pdm/cli/commands/venv/create.py src/pdm/cli/commands/venv/list.py src/pdm/cli/commands/venv/purge.py src/pdm/cli/commands/venv/remove.py tests/cli/test_venv.py`
- `python -m compileall -q src/pdm/cli/commands/venv`
- `python -m pytest tests/cli/test_venv.py::test_core_registration_does_not_import_venv_runtime_helpers -q`
- `python -m pdm --version`
- `python -m pdm venv --help`
- `python -m pdm venv create --help`
- `python -m pdm venv activate --help`
- `python -X importtime -m pdm --version` shows no `pdm.cli.commands.venv.utils`, `pdm.cli.commands.venv.backends`, or `shellingham` import-time entries.

Attempted `python -m pytest tests/cli/test_venv.py -q` locally, but this Windows checkout is under a non-ASCII user path and the shared test fixture fails before test assertions with `UnicodeDecodeError` while decoding the interpreter path.

## Breaking changes

None.

### DIFF
--- a/news/3673.feature.md
+++ b/news/3673.feature.md
@@ -1,1 +1,1 @@
-Defer startup-time imports for Python and self-management commands.
+Defer startup-time imports for Python, virtualenv, and self-management commands.
--- a/src/pdm/cli/commands/venv/__init__.py
+++ b/src/pdm/cli/commands/venv/__init__.py
@@ -8,7 +8,6 @@ from pdm.cli.commands.venv.create import CreateCommand
 from pdm.cli.commands.venv.list import ListCommand
 from pdm.cli.commands.venv.purge import PurgeCommand
 from pdm.cli.commands.venv.remove import RemoveCommand
-from pdm.cli.commands.venv.utils import get_venv_with_name
 from pdm.cli.options import project_option
 
 if TYPE_CHECKING:
@@ -36,6 +35,8 @@ class Command(BaseCommand):
         self.parser = parser
 
     def handle(self, project: Project, options: Namespace) -> None:
+        from pdm.cli.commands.venv.utils import get_venv_with_name
+
         if options.path:
             venv = get_venv_with_name(project, options.path)
             project.core.ui.echo(str(venv.root))
--- a/src/pdm/cli/commands/venv/activate.py
+++ b/src/pdm/cli/commands/venv/activate.py
@@ -5,16 +5,13 @@ import shlex
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-import shellingham
-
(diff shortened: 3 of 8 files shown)


frostming:
Can you provide the benchmark on how much it improves the startup time?

Yijian6:
Thanks, I benchmarked it locally against the latest `origin/main`.

Environment:
- Windows-10-10.0.26200-SP0
- Python 3.11.7
- Same virtualenv for both runs, with `PYTHONPATH` pointed at each source tree
- `PDM_CHECK_UPDATE=0`, `PDM_IGNORE_ACTIVE_VENV=1`

Command measured: `python -m pdm --version`

Method: 5 warmup runs per target, then 30 interleaved runs per target.

Compared commits:
- main: `a8429941`
- this PR: `55ee9b24`

Results:

| target | mean | median | min | max | stdev |
| --- | ---: | ---: | ---: | ---: | ---: |
| main | 1144.12 ms | 1100.08 ms | 1027.33 ms | 1819.19 ms | 143.26 ms |
| this PR | 1111.99 ms | 1071.12 ms | 1028.26 ms | 1673.08 ms | 126.08 ms |

Median improvement: **28.96 ms** / **2.63%** on this machine.

I also checked `python -X importtime -m pdm --version` to verify the import-path impact:

```text
main:
import time:      2059 |       2059 |     shellingham._core
import time:      2455 |       4513 |   shellingham
import time:      1222 |      24616 |   pdm.cli.commands.venv.utils
import time:      1560 |       1560 |   pdm.cli.commands.venv.backends

this PR:
no venv utils/backends or shellingham importtime entries
```

So the wall-clock startup improvement is modest in my local Windows environment, but the PR does remove these venv runtime modules from the startup import path.

frostming:
That's not very exciting, as expected. merge it anyway

PR #3815, opened 2026-07-15
feat(scripts): add {run_cwd} placeholder for cross-platform scripts

### LINKED ISSUE #3734: {PDM_RUN_CWD} placeholder in scripts?
### Feature Description

For a project, we use

```
[tool.pdm.scripts]
create-symlinks = {shell = "python tools/create_symlinks.py $PDM_RUN_CWD"}
```

However, this solution is not cross-plateform since `$PDM_RUN_CWD` does not work on Windows.

Could we have a cross-platform solution, maybe something like

```
create-symlinks = {shell = "python tools/create_symlinks.py {PDM_RUN_CWD}"}
```

(inspired by `{pdm}` https://pdm-project.org/latest/usage/scripts/#pdm-placeholder)

Of course, for this particular case, we could also just make `create_symlinks.py` aware of `$PDM_RUN_CWD` but it would be less general for us and I would prefer to keep the things related to PDM in PDM scripts.

### Problem and Solution

`create-symlinks = {shell = "python tools/create_symlinks.py $PDM_RUN_CWD"}` is not cross-platform.

### Additional Context

_No response_

pctablet505:
Fixes #3734

Adds a `{PDM_RUN_CWD}` placeholder for PDM scripts so callers can keep PDM-specific paths out of tools that cannot read environment variables. Works in `cmd`, `shell`, and `composite` scripts; paths with spaces are handled via `shlex.quote`. Includes tests and docs.

### DIFF
--- a/docs/usage/scripts.md
+++ b/docs/usage/scripts.md
@@ -402,15 +402,15 @@ PDM, version 2.8.0 was called as <snip>/venvs/pdm2-8/bin/python -m pdm -V
 !!! note
     While the above example uses PDM 2.8, this functionality was introduced in the 2.10 series and only backported for the showcase.
 
-### `{PDM_RUN_CWD}` placeholder
+### `{pdm_run_cwd}` placeholder
 
-To make scripts referencing the current working directory cross-platform, you can use the `{PDM_RUN_CWD}` placeholder.
+To make scripts referencing the current working directory cross-platform, you can use the `{pdm_run_cwd}` placeholder.
 It expands to the absolute path of the directory from which `pdm run` was invoked, the same value that is exposed
 via the `PDM_RUN_CWD` environment variable.
 
 ```toml
 [tool.pdm.scripts]
-create-symlinks = { shell = "python tools/create_symlinks.py {PDM_RUN_CWD}" }
+create-symlinks = { shell = "python tools/create_symlinks.py {pdm_run_cwd}" }
 ```
 
 ## Show the List of Scripts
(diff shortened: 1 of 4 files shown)


yangfan-yf-yf (changes requested):
I found a Windows regression in the placeholder substitution.

`RE_CWD_PLACEHOLDER.subn(cwd, script)` treats `cwd` as a regular-expression replacement template. A normal Windows working directory such as `C:\Users\...` contains `\U`, which `re` interprets as an invalid replacement escape and raises `re.error: bad escape \U at position 2` before the command runs.

I reproduced this with the added coverage on Windows:

```text
python -m pytest tests/cli/test_run.py -k cwd_placeholder -q
6 failed, 74 deselected
```

This affects both the shell and command modes, including the new spaces/composite cases. Please pass a callable replacement (for example `lambda _: cwd`) to `subn`, or otherwise escape the replacement value, so the path is inserted literally.

yangfan-yf-yf (changes requested):
Thanks for changing the replacement to a callable; that fixes the original `\\U` replacement-template failure.

There is still a Windows shell-quoting issue in the new `for_shell` branch. `subprocess.list2cmdline()` follows the Microsoft C runtime argument-parsing rules, but the result is passed to `cmd.exe`. For a working directory without whitespace such as `C:\\repo&ver`, `subprocess.list2cmdline([path])` returns `C:\\repo&ver` without quotes. With `cmd.exe /d /c echo C:\\repo&ver`, `&` is then parsed as a command separator (the output includes both the path and the Windows version) rather than as part of the path.

Could the shell-specific interpolation escape or quote `cmd.exe` metacharacters even when the path has no spaces, and add a regression case for such a path? The existing spaced-path test does not cover this branch.

frostming:
@pctablet505 all other placeholders are lowercased, can you use lowercase, too?

frostming:
FYI @pctablet505 I've jumped in and changed to the name to `run_cwd`

PR #3819, opened 2026-07-21
fix(workspace): keep declared group for workspace members in lockfile

### LINKED ISSUE #3816: Workspace members are always treated as default dependencies
### Describe the bug

Any workspace dependency added to the project as an optional or development dependency is marked as residing in `default` group in `pdm.lock` file.

### To reproduce

Clone repo: https://github.com/DenisKuplyakov/pdm-issue-3816 and run `./reproduce.sh`.

### Expected Behavior

Workspace projects are treated according to the dependency groups they are reside in as dependencies.

### Environment Information

```
PDM version:
  2.28.0
Python Interpreter:
  <REDACTED>/__2026-07-16-pdm-bug/.venv/bin/python (3.13)
Project Root:
  <REDACTED>/__2026-07-16-pdm-bug
Local Packages:
  <REDACTED>/__2026-07-16-pdm-bug/.venv/lib/python3.13/site-packages
{
  "implementation_version": "3.13.3",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.15.0-1-default",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT_DYNAMIC Mon Jun  2 07:13:45 UTC 2025 (17ba886)",
  "python_full_version": "3.13.3",
  "python_version": "3.13",
  "sys_platform": "linux"
}
```

### Verbose Command Output

_No response_

### Additional Context

Removal of `[tool.pdm.workspace]` results in correct dependency groups handling. See https://github.com/DenisKuplyakov/pdm-issue-3816/commit/ee12697940729e25bee60fd7a1acb1569a4adeb7 in the reproducer.

- [ ] Yes, I would like to submit a PR.

Solaris-star:
## Description

Workspace members are always injected as implicit editable dependencies with `groups = ["default"]`. When a member is only declared in a non-default group (e.g. `pdm add -dG test ./subproject`), the lockfile still marked it as `default` (and sometimes both `default` and `test`).

Removing the workspace config made groups correct, so the bug is specifically the implicit default injection path.

## Fix

1. `with_workspace_dependencies(..., exclude=...)` — skip members already declared in other groups when injecting into `default`.
2. `_resolve_dependencies` — compute `declared_elsewhere` and pass it as `exclude`.
3. Resolver `find_matches` — `copy_with(original_req)` for workspace candidates so requirement groups follow the declared dependency.

## Verification

Minimal workspace repro (`pdm add -dG test ./subproject`):

```toml
# pdm.lock excerpt
[[package]]
name = "subproject"
groups = ["test"]   # no longer includes "default"
```

## Linked Issue

Closes #3816


frostming (approved):

frostming:
The tests are failed, please check @Solaris-star

yangfan-yf-yf (changes requested):
I reproduced the two CI failures on `f01a437` locally on Windows:

```powershell
py -m uv run pytest tests/cli/test_add.py::test_add_subdirectory_path_updates_workspace tests/test_project.py::test_workspace_member_declared_in_dev_group_not_forced_into_default -q
# 2 failed
```

The first failure is a regression from `can.copy_with(original_req)`: the workspace candidate used by `pdm add packages/foo` is no longer editable. The same existing test passes on the current `main` branch, so the workspace candidate needs to retain its editable requirement while still carrying the declared dependency's group information.

The new test also calls `project.get_dependencies(default)`. That resolves only the default group, so `declared_elsewhere` never sees the `test` dependency and the implicit workspace member is expected to be added. Please exercise the all-group lock-resolution path (or otherwise include the declared group in the resolution) when checking the lockfile group result.

Solaris-star:
Thanks for the precise repro. I pushed `ad919d74` with both fixes: the workspace candidate keeps its editable requirement while inheriting only the declared groups, and the regression test now resolves all groups and verifies the final lockfile records `foo` only in `test`. I also added the required news fragment. The two focused tests pass locally.

PR #3822, opened 2026-07-22
use_uv: Support adding local files

didib:
In a project configured `use_uv = true`, this fails:
```
$ pdm add /path/to/file.whl
...
[KeyError]: 'url'
```

Fix this:

src/pdm/resolver/uv.py:make_hash:
- Fallback to `filename` if `url` is missing
- This is the only part which I did by myself, using debug prints. The rest was done by Claude, with rather-shallow review by me.

src/pdm/formats/uv.py:_build_lock_source:
- Support `path`

src/pdm/formats/uv.py:_build_lock_entry:
- Write `filename` if entry is a local file

tests/resolver/test_uv_resolver.py:
- Add a suitable test

tests/test_formats.py:
- Add a suitable test

Assisted-By: Claude Code (Opus 4.6)

## Pull Request Checklist

### DIFF
--- a/news/3822.bugfix.md
+++ b/news/3822.bugfix.md
@@ -1,0 +1,1 @@
+Make `use_uv` support adding local files.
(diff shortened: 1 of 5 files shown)


didib:
Added a `news` item

PR #3852, opened 2026-08-18
fix: Poetry single-component tilde constraint converts to an invalid specifier

VXNCXNX:
`pdm import -f poetry` crashes on a single-component tilde constraint such as `foo = "~1"`. It is converted to `~=1`, which is not a valid PEP 440 specifier:

```
before   pdm.formats.base.MetaConvertError:
         dependencies: Invalid specifier for foo: ~=1: Invalid specifier: '~=1'

after    dependencies = [
             "foo<2,>=1",
             "bar~=2.5",
             "baz~=1.2.3",
         ]
```

Poetry defines `~1` as `>=1.0.0,<2.0.0`, so the single-component case reuses `_caret_upper_bound`, which gives the same bound for this shape. Multi-component tilde keeps mapping to `~=`, so `~2.5` and `~1.2.3` are untouched.

Follow-up to #3848, which fixed the caret branch and left this one. Test added in `tests/test_formats.py`; removing the branch brings back `Invalid specifier: '~=1'`. `tests/test_formats.py` passes apart from `test_export_from_pylock_not_empty`, which fails identically on an unmodified checkout. ruff check and format clean.

AI disclosure: written with Claude Code. I ran `pdm import` before and after and checked the mutation myself.


src/pdm/formats/poetry.py
```diff
@@ -62,6 +62,11 @@ def _convert_specifier(version: str) -> str:
     parts = []
     for op, ver in VERSION_RE.findall(str(version)):
         if op == "~":
+            if "." not in ver:
+                # ``~=1`` is not a valid PEP 440 specifier, and Poetry's ``~1``
+                # allows the minor version to change, so it means ``>=1,<2``.
+                parts.append(f">={ver},<{_caret_upper_bound(ver)}")
```
frostming, line 68: How about changing to `~=1.0`, i think they are equivalent

PR #3860, opened 2026-08-28
fix(formats): parse multi-constraint and v-prefixed Poetry versions

deepspace28:
The operator group of `VERSION_RE` matched any non-digit character. In a multi-constraint string that swallowed the separating comma into the *next* operator, which was then re-emitted verbatim:

```python
>>> from pdm.formats.poetry import _convert_specifier
>>> _convert_specifier(">=1.0,<2.0")
">=1.0,,<2.0"
>>> _convert_specifier(">=1.2,!=1.5,<2.0")
">=1.2,,!=1.5,,<2.0"
```

`SpecifierSet` happens to tolerate the empty item, so this is not fatal, but every imported multi-constraint dependency lands in `pyproject.toml` with a doubled comma.

The same group also absorbed a `v` prefix, which Poetry accepts, leaving an operator of `^v` that matched no branch and passed straight through:

```python
>>> _convert_specifier("^v1.2.3")
"^v1.2.3"   # InvalidSpecifier
```

Restricting the group to the operator characters Poetry actually uses and skipping an optional `v` fixes both. Behaviour is unchanged for every other form I could find - bare versions, wildcards, epochs, spaced operators, `~`/`^` bounds.

Covered by a parametrized test alongside the existing caret/tilde ones.

### DIFF
--- a/news/3860.bugfix.md
+++ b/news/3860.bugfix.md
@@ -1,0 +1,1 @@
+Keep the separating comma out of the operator and accept a `v` prefix when converting Poetry version constraints.
--- a/src/pdm/formats/poetry.py
+++ b/src/pdm/formats/poetry.py
@@ -41,7 +41,11 @@ def check_fingerprint(project: Project | None, filename: Path | str) -> bool:
     return "tool" in data and "poetry" in data["tool"]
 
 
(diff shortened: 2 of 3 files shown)


frostming:
@deepspace28 please resolve the conflicts

deepspace28:
Rebased on main, conflicts resolved. Kept both new tests in `tests/test_formats.py`.

PR #3863, opened 2026-08-28
fix(formats): keep Windows paths intact in requirements option lines

deepspace28:
Option lines go through `shlex.split`, which runs in POSIX mode where a backslash escapes the next character. A native Windows path in a `-e` or `-r` line is destroyed before argparse ever sees it:

```python
>>> shlex.split(r"-e C:\src\pkg")
["-e", "C:srcpkg"]
>>> shlex.split(r"-r C:\src\reqs.txt")
["-r", "C:srcreqs.txt"]
```

So `pdm import` of a requirements file that references a local package or a nested requirements file by absolute Windows path silently resolves the wrong location. There is no error - the path just loses its separators.

Escape backslashes on Windows before splitting, which keeps quoting behaving as it does everywhere else:

```python
>>> RequirementParser._split_args(r"-e \"C:\src\my pkg\"")
["-e", "C:\src\my pkg"]
```

POSIX platforms are untouched, since a backslash there is a legitimate escape. Covered by a parametrized test that pins `os.name` so it runs on every platform.

### DIFF
--- a/news/3863.bugfix.md
+++ b/news/3863.bugfix.md
@@ -1,0 +1,1 @@
+Keep native Windows paths intact when parsing `-e` and `-r` lines in a requirements file.
--- a/src/pdm/formats/requirements.py
+++ b/src/pdm/formats/requirements.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import argparse
 import hashlib
+import os
 import re
 import shlex
 import urllib.parse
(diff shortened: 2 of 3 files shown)


deepspace28:
Withdrawing this - I should have checked the reference implementation first.

pip does exactly the same thing, for the same reason: `break_args_options` hands the whole option string to `shlex.split` in POSIX mode.

```python
>>> from pip._internal.req.req_file import break_args_options, build_parser
>>> # pip 26.2.1
>>> opts, _ = parser.parse_args(shlex.split(break_args_options(r"-e C:\src\pkg")[1]), defaults)
>>> opts.editables
["C:srcpkg"]
```

`-r` and `--find-links` behave the same. Since `RequirementParser` documents itself as following pip's requirements-file format, matching pip here is the correct behaviour and this patch would be a silent divergence rather than a fix. Sorry for the noise.

PR #3875, opened 2026-09-03
feat: support exclude-newer overrides

### LINKED ISSUE #3809: Allow overrides to `exclude-newer`
### Feature Description

Add additional config options for resolution for specific packages:

```toml
[tool.pdm.resolution]
exclude-newer = "7d"

[tool.pdm.resolution.exclude-newer-override]
mypackage = 0
anotherpackage = "3d"
```

```
pdm lock --exclude-newer 7d --exclude-newer-override mypackage=0 anotherpackage=3d
```

### Problem and Solution

Many packages are developed in parallel with a family of packages that are updated closely together. the purpose of dep cooldowns is to limit supply chain risk from external packages and the dep tree, but it doesn't really make sense to enforce within the boundary of a dev group. to make exclude-newer work as a projectwide option, it needs to be possible to override. 

this could be done with a simple exclude statement of package names, probably easier to implement, but i figure an override dict is the most general so i thought i'd pitch that.

### Additional Context

e.g. locking fails here: https://github.com/miniscope/minian/actions/runs/27744808625/job/82080473340?pr=335

because the package `minisim` which is developed concurrently was uploaded within the 7d dep cooldown window

lmk what's good. sorry for stalling on the typing PRs, but i figure those are low priority, will get back to those soon. would be willing to fix this asap once you get a second to check it out and say what path you'd like to take on this

frostming:
## Pull Request Checklist

Add package-specific `exclude-newer` overrides in `pyproject.toml` and the lock CLI. Overrides accept a date or duration, while `false` disables the global cutoff for a package.

Package names are normalized before matching. The overrides are applied by both the PDM and uv resolvers, with command-line values taking precedence over project configuration.

Closes #3809

### DIFF
--- a/docs/usage/lockfile.md
+++ b/docs/usage/lockfile.md
@@ -229,10 +229,33 @@ pdm config strategy.exclude-newer 7d
 exclude-newer = "7d"
 ```
 
+Package-specific values can override the global cutoff:
+
+```toml
+[tool.pdm.resolution]
+exclude-newer = "7d"
+
+[tool.pdm.resolution.exclude-newer-override]
+mypackage = false
+anotherpackage = "3d"
+```
+
+An override accepts the same date, timestamp, and duration formats as `exclude-newer`. Set it to `false` to disable the
+cutoff for that package. Package names are normalized before matching, and package-specific values take precedence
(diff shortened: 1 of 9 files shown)


sneakers-the-rat:
thank you!

PR #3883, opened 2026-09-08
feat: allow disabling managed Python selection

wolfgang-aura:
Closes #3753.

PDM currently has no way to require a separately installed Python. In
particular, `PDM_NO_MANAGED_PYTHON` is not recognized, and Python selection can
continue to use or install an interpreter under PDM's managed directory. This
prevents organizations from enforcing centrally approved Python installations.

The change adds the `python.use_managed` setting, enabled by default, together
with the `PDM_NO_MANAGED_PYTHON=true` environment override. When managed Python
is disabled, saved selections, automatic installation, PATH lookups, and pyenv
shims are checked against the managed install root. Explicit external
interpreters, including interpreters linked into that root, remain selectable.

### How this was tested

The issue reproduces at the base commit: `pdm python install
--no-managed-python cpython@3.12` exits 2 with `unrecognized arguments`.

With the new regression tests retained and the source change reverted, 8
managed-Python policy tests failed and 11 tests were deselected. With the
change restored, the focused suite passes 17 tests with 2 deselected:

```text
python -m pytest tests/cli/test_python.py -q -k "not link_python"
17 passed, 2 deselected
```

The passing command was run by Mailman in a clean candidate workspace on
Windows 11 with Python 3.14.3. The unfiltered file has one unrelated Windows
symlink-permission test, so `link_python` is excluded from this focused gate.
Optional ruff and mypy checks were unavailable in the run environment.

### An alternative I did not take

I considered adding a one-off flag only to `pdm use`. A persisted setting plus
an environment override applies the same policy to saved selections,
discovery, and automatic installation, which is the organization-wide control
requested here. The trade-off is two configuration surfaces instead of a
single command-line flag.

### AI disclosure

This change was drafted with AI assistance (codex (`gpt-5.6-luna`) wrote the
patch, and codex (`gpt-5.6-luna`) reviewed it), running under Mailman. The test
results above come from the harness executing the commands itself, not from
either agent's account of its own work. The submitting human must read, test,
and take responsibility for every line before filing.

### DIFF
--- a/src/pdm/project/core.py
+++ b/src/pdm/project/core.py
@@ -6,6 +6,7 @@ import itertools
 import operator
 import os
 import shutil
+import subprocess
 import sys
 from collections.abc import Callable, Collection, Iterable, Mapping, Sequence
 from copy import deepcopy
@@ -245,6 +246,13 @@ class Project:
         return collections.ChainMap(self.project_config, self.global_config)
 
     @property
+    def use_managed_python(self) -> bool:
+        """Whether PDM may select or install a Python interpreter it manages."""
+        if ensure_boolean(os.getenv("PDM_NO_MANAGED_PYTHON")):
+            return False
+        return self.config["python.use_managed"]
+
+    @property
     def scripts(self) -> dict[str, str | dict[str, str]]:
         return self.pyproject.settings.get("scripts", {})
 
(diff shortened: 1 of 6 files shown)


src/pdm/project/core.py
```diff
@@ -244,6 +245,13 @@ def config(self) -> Mapping[str, Any]:
 
         return collections.ChainMap(self.project_config, self.global_config)
 
+    @property
+    def use_managed_python(self) -> bool:
+        """Whether PDM may select or install a Python interpreter it manages."""
+        if ensure_boolean(os.getenv("PDM_NO_MANAGED_PYTHON")):
+            return False
+        return self.config["python.use_managed"]
```
frostming: Such a property can be omitted, because environment variables have already been handled at the configuration layer.

PR #3885, opened 2026-09-09
fix(formats): keep quoted local paths containing space-dash

r3wretrhy:
`RequirementParser._parse_line` used `line.split(" -")` on requirement lines. A quoted local path containing `` - `` such as `"./foo - bar"` was therefore truncated to `"./foo"` before `parse_requirement` ran.

Split with `shlex` instead, stop at the first option token, and `shlex.join` so the quoted path is restored. Lines that start with `-` still go through `parse_known_args`.

The added test fails on the unpatched tree and passes with the change.

### DIFF
--- a/src/pdm/formats/requirements.py
+++ b/src/pdm/formats/requirements.py
@@ -30,6 +30,19 @@ if TYPE_CHECKING:
 _COMMENT_RE = re.compile(r"(^|\s+)#.*$")
 
 
+def _requirement_before_options(line: str) -> str:
+    """Drop unquoted `` -`` options; keep quoted paths that contain `` - ``."""
+    in_single = in_double = False
+    for i, ch in enumerate(line):
+        if ch == "'" and not in_double:
+            in_single = not in_single
+        elif ch == '"' and not in_single:
+            in_double = not in_double
+        elif not in_single and not in_double and line.startswith(" -", i):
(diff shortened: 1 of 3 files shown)


src/pdm/formats/requirements.py
```diff
@@ -61,8 +61,16 @@ def _clean_line(self, line: str) -> str:
 
     def _parse_line(self, filename: str, line: str) -> None:
         if not line.startswith("-"):
-            # Starts with a requirement, just ignore all per-requirement options
-            req_string = line.split(" -", 1)[0].strip()
+            # Starts with a requirement, just ignore all per-requirement options.
+            # Split with shlex so a quoted local path containing " - " is kept intact;
+            # str.split(" -") used to cut those paths in half.
```
frostming:
```suggestion
```

r3wretrhy: Removed in b323994.

r3wretrhy: CI fix in eac1ac3: quote-aware split so markers and --hash stay intact.

r3wretrhy: 7ce2f4f

r3wretrhy: 202ff38

PR #3888, opened 2026-09-14
fix(formats): keep the marker of a quoted local path

### LINKED ISSUE #3887: pdm import -f requirements drops the marker of a quoted local path
### Describe the bug

`pdm import -f requirements` drops the environment marker of a quoted local path and corrupts the path itself. #3885 added support for quoted local paths, but it decides that a line is a quoted path by checking whether the line ends with the same quote character it starts with. When the requirement carries a marker, the line ends with the marker's closing quote, so the whole `"./local pkg" ; python_version < "3.9"` is passed to `FileRequirement.create()` as one path.

The marker is silently lost and the resulting dependency is unusable, rather than raising an error.

### To Reproduce

`requirements.txt`:

```
"./local pkg" ; python_version < "3.9"
```

where `./local pkg/pyproject.toml` declares `name = "mylocal"`.

```bash
pdm import -f requirements requirements.txt
```

Expected:

```toml
dependencies = [
    "mylocal @ file:///${PROJECT_ROOT}/local%20pkg ; python_version < \"3.9\"",
]
```

Actual:

```toml
dependencies = [
    "file:///${PROJECT_ROOT}/local%20pkg%22%20%3B%20python_version%20%3C%20%223.9",
]
```

At the parser level:

```python
from pdm.formats.requirements import RequirementParser

parser = RequirementParser(None)
parser._parse_line("reqs.txt", '"./my pkg" ; python_version < "3.9"')
req = parser.requirements[0]
print(repr(req.str_path), req.marker)
# './my pkg" ; python_version < "3.9' None
```

The same line without the marker parses correctly (`str_path == './my pkg'`), and `whoosh==2.7.4; sys_platform == "win32"` keeps its marker, so the mis-detection only affects the quoted-path branch.

### Expected behavior

The quoted path and the trailing `; marker` clause are parsed as separate parts, so both survive the import.

### Environment

- PDM 2.29.1
- Python 3.13

tunglambk:
## Pull Request Checklist

Fixes #3887.

`_parse_line` decided that a line was a quoted local path by comparing the first character of the line with the last one. A quoted path followed by an environment marker ends with the marker's closing quote, so `"./local pkg" ; python_version < "3.9"` was passed whole to `FileRequirement.create()`: the marker was dropped and the path was percent-encoded into a dependency that cannot be installed.

`_split_quoted_local_path()` now finds the closing quote of the path and returns the path and the trailing `; marker` clause as two parts, so `FileRequirement` receives them as separate fields. A line that is not a closed quoted section, or that has something other than a marker after the quotes, still goes through `parse_requirement()` unchanged.

Importing `"./local pkg" ; python_version < "3.9"` from a directory named `mylocal`:

```toml
# before
dependencies = [
    "file:///${PROJECT_ROOT}/local%20pkg%22%20%3B%20python_version%20%3C%20%223.9",
]

# after
dependencies = [
    "mylocal @ file:///${PROJECT_ROOT}/local%20pkg ; python_version < \"3.9\"",
]
```

The tests cover double quotes, single quotes and a missing space before `;` at the parser level, and a `convert()` case asserting the imported dependency keeps the name, the path and the marker. `tests/test_formats.py` passes locally except `test_convert_setup_py_project`, which fails here while creating a virtualenv on an unmodified checkout as well.


frostming:
> "./local pkg" ; python_version < "3.9"

I don't think this is an allowed syntax in pip's requirements file. I tried consuming the requirements.txt with that line with pip 26.2.1 and it throws an InstallationError.

PR #3898, opened 2026-09-22
fix: make pdm install interruptible with Ctrl+C

### LINKED ISSUE #2180: `pdm install` cannot be interrupted with ^C on Linux

There are a few issues dealing with `^C` but none seem to be related to this problem specifically.

## Steps to reproduce

Run a `pdm install` command somewhere with a large lockfile, then press Ctrl+C

## Actual behavior

Something like this:

```
$ pdm install
The saved Python interpreter doesn't match the project's requirement. Trying to find another one.
python.use_venv is on, creating a virtualenv for this project...
Virtualenv is created successfully at /home/whitequark/Projects/zig-pypi/.venv
Synchronizing working set with lock file: 30 to add, 0 to update, 0 to remove

  ✔ Install wheel 0.38.4 successful
  ⠦ Installing bleach 6.0.0...
  ⠦ Installing certifi 2023.7.22...
  ⠦ Installing cffi 1.15.1...
  ⠦ Installing charset-normalizer 3.2.0...
  ⠦ Installing cryptography 41.0.3...
  ⠦ Installing docutils 0.20.1...
  ⠦ Installing idna 3.4...
  ⠦ Installing importlib-metadata 6.7.0...
  File "/home/whitequark/.local/bin/pdm", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/whitequark/.local/pipx/venvs/pdm/lib/python3.11/site-packages/pdm/core.py", line 290, in main
    return Core().main(args or sys.argv[1:])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/whitequark/.local/pipx/venvs/pdm/lib/python3.11/site-packages/pdm/core.py", line 203, in main
    self.handle(project, options)
  File "/home/whitequark/.local/pipx/venvs/pdm/lib/python3.11/site-packages/pdm/core.py", line 157, in handle
  File "/home/whitequark/.local/pipx/venvs/pdm/lib/python3.11/site-packages/pdm/cli/commands/install.py", line 95, in handle
    actions.do_sync(
  File "/home/whitequark/.local/pipx/venvs/pdm/lib/python3.11/site-packages/pdm/cli/actions.py", line 213, in do_sync
    synchronizer.synchronize()
  File "/home/whitequark/.local/pipx/venvs/pdm/lib/python3.11/site-packages/pdm/installers/synchronizers.py", line 430, in synchronize
    with self.create_executor() as executor:
  File "/usr/lib/python3.11/concurrent/futures/_base.py", line 647, in __exit__
    self.shutdown(wait=True)
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 235, in shutdown
    t.join()
  File "/usr/lib/python3.11/threading.py", line 1112, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.11/threading.py", line 1132, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
  ✔ Install certifi 2023.7.22 successful
r  ✔ Install bleach 6.0.0 successful
  ✔ Install idna 3.4 successful
m   ✔ Install jaraco-classes 3.2.3 successful
  ✔ Install jeepney 0.8.0 successful
  ✔ Install docutils 0.20.1 successful
  ✔ Install importlib-metadata 6.7.0 successful
^CException ignored in: <module 'threading' from '/usr/lib/python3.11/threading.py'>
  File "/usr/lib/python3.11/threading.py", line 1553, in _shutdown
    atexit_call()
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 31, in _python_exit
    t.join()
  File "/usr/lib/python3.11/threading.py", line 1112, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.11/threading.py", line 1132, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt:
  ✔ Install libarchive-c 5.0 successful
```

## Expected behavior

The command terminating immediately on the first Ctrl+C press.

## Environment Information

```bash
PDM version:
  2.8.2
Python Interpreter:
  /home/whitequark/Projects/zig-pypi/.venv/bin/python (3.11)
Project Root:
  /home/whitequark/Projects/zig-pypi
Local Packages:

{
  "implementation_version": "3.11.2",
  "os_name": "posix",
  "platform_machine": "x86_64",
  "platform_release": "6.1.0-9-amd64",
  "platform_system": "Linux",
  "platform_version": "#1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1 (2023-05-08)",
  "python_full_version": "3.11.2",
  "python_version": "3.11",
  "sys_platform": "linux"
}
```

Vinayak19112003:
Pressing Ctrl+C during `pdm install` didn't actually stop anything. The interrupt lands while the main thread is blocked waiting on the install thread pool, so the queued installs keep running in the background anyway, and the user gets a raw `KeyboardInterrupt` traceback instead of a clean exit.

What this changes:

- `src/pdm/installers/synchronizers.py`: instead of one blocking `executor.shutdown(wait=True)`, wait on the install futures in short slices so ^C is acted on promptly (on newer Pythons a blocking thread join doesn't react to signals at all — the interrupt is deferred until every install finishes). On interrupt, installs that haven't started are cancelled, while the ones already running are allowed to finish so no package is left half-installed. A second ^C aborts the wait immediately. The progress callback now also tolerates cancelled futures.
- `src/pdm/core.py`: `Core.main()` catches `KeyboardInterrupt` and prints a one-line `Interrupted` message, exiting with code 130 instead of dumping a traceback. This applies to every command, not just `install`.
- `tests/test_installer.py`: regression test driving the real `synchronize()` with a stubbed slow installer and a real SIGINT delivered to the main thread mid-install. It fails on current main (the queued install still runs after ^C) and passes with the fix.
- `news/2180.bugfix.md`: news fragment.

How I tested:

- Reproduction script against current main: 3 queued installs of 2s each, SIGINT after 1s — before the fix all 3 installs ran to completion (6s); after the fix the in-flight install finishes and the queued ones are cancelled (2s), `KeyboardInterrupt` propagates for the CLI to report.
- New regression test passes; verified it fails on the unpatched code.
- `tests/test_installer.py` and `tests/test_synchronizer_base.py`: no new failures (the 30 failures in `test_installer.py` are pre-existing sandbox proxy issues, identical with and without this change).
- `ruff check`, `ruff format --check`, and `mypy` are clean on the changed files.

Closes #2180.


src/pdm/installers/synchronizers.py
```diff
@@ -202,11 +207,36 @@ def update_progress(future: Future, kind: str, key: str) -> None:
...
                             future = executor.submit(handlers[kind], key, status.progress)
                             future.add_done_callback(functools.partial(update_progress, kind=kind, key=key))
                             state.jobs.append(future)
+                        # Wait in short slices instead of one blocking shutdown, so that
+                        # ^C is acted on promptly even on Python versions where a blocking
+                        # thread join doesn't react to signals.
+                        while True:
+                            _, pending = wait(state.jobs, timeout=0.2, return_when=ALL_COMPLETED)
+                            if not pending:
+                                break
```
frostming: It is worth a helper function.

Vinayak19112003: Done — extracted it into a `_wait_for_jobs` helper and both call sites use it now. Let me know if the shape looks right to you.

frostming (approved):

